PR #9, opened 2017-02-01
Added shared scheme for Carthage support

natestedman:
Using Lottie with [Carthage](https://github.com/carthage/carthage) requires a shared scheme.

After adding it, works perfectly!

```
$ carthage build --platform ios --no-skip-current
*** xcodebuild output can be found in /var/folders/pp/6pmv88_520d30c39vlrr7w4w0000gn/T/carthage-xcodebuild.LWNBYI.log
*** Building scheme "Lottie" in Lottie.xcodeproj
```


buba447:
Thank you for grabbing this!

PR #10, opened 2017-02-02
Implemented macOS support

pawlowskialex:
It's not perfect, the compatibility layer could be prettier, but it works.
I'm using 10.12 improvements in AutoLayout here, so it probably won't work on lesser systems.

Also added macOS example project.


buba447:
Woah! Amazing work! Im going to give this a review today. Thank you for the hard work!

pawlowskialex:
Thanks!
Let me know if you have any suggestions for improvement

buba447:
@pawlowskialex Man this is incredible! I can't wait to merge it and build out a comparable animation explorer with it! Seriously excellent work!

I checked the code and it seems legit! My only comment is the OS X version. I bumped mine down to 10.11 and it worked. Well at least it ran without crashing. Im not so up to date with the layout changes added in 10.12 but is there any way you can drop the version down to 10.11 and fix any issues you might know of?

Thanks!!!

pawlowskialex:
Yeah, 10.11 is surely doable, maybe with a couple of caveats, but doable.
I'm just not sure that the current implementation of [NSView layout] will work properly in all cases.

I'll look into it in a couple of days when I get back to civilization.

pawlowskialex:
Thanks!

pawlowskialex:
I've checked with the docs, and according to them [NSView layout] should get called regardless of Auto Layout being on/off if [NSView wantsUpdateLayer] == YES.
So it seems my doubts about <10.12 systems were unfounded.

I've tested the example on 10.12, 10.11 and 10.10 machines, and everything seems to be working great.

buba447:
Awesome! Thank you!

buba447:
@pawlowskialex hey! Can you fix the merge conflict? I'm away from my computer today. If you can't I will grab it tomorrow. Thank you again for this incredible work.

pawlowskialex:
Done! Thank you for kind words!

PR #11, opened 2017-02-02
Fix all project warnings, specify nullability

hansemannn:
Hey guys! Thanks for accepting the first PR. Taking a deeper look into the source, I found some semantic warnings that are all resolved using this patch. Here is what's included:
- [x] Resolve docs warnings by normalizing them to the Apple docs recommendation
- [x] Resolve the `TODO` regarding throwing an exception when no layer is specified
- [x] Add missing nullability-specifiers to the `LAAnimationState` class

Some additional notes:
- We should try to document all public headers, including the class interfaces, properties and methods
- There should be a `Lottie.framework` in the releases tab for easier integration
- I've made [Ti.Lottie](https://github.com/hansemannn/ti.lottie), a native module for the cross-platform Titanium framework. The same was done for the keyframes lib, so I'd be happy to have a "Frameworks" section in the README to list those integrations, together with react and others. Are you guys open for that?

### DIFF
--- a/lottie-ios/Classes/LAAnimationView.m
+++ b/lottie-ios/Classes/LAAnimationView.m
@@ -386,9 +386,10 @@ const NSTimeInterval singleFrameTimeValue = 1.0 / 60.0;
   newChild.childView = view;
   
   if (!layer) {
-    // TODO Throw Error
-    [self.layer addSublayer:view.layer];
-    newChild.layer = self.layer;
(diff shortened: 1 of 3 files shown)


buba447:
This is great! Thank you!

m1ga:
To make it cross platform I've created a Android version for Titanium at [Ti.Animation](https://github.com/m1ga/ti.animation) 
❤️ Titanium

PR #24, opened 2017-02-03
Initialize control point to CGPointZero

ianegordon:
Fix warnings about uninitialized variable.


buba447:
Thanks for cleaning this up!

PR #27, opened 2017-02-04
fix crash in AnimatedTextField

flappyFeline:
When i continuous typing and deleting on AnimatedTextField,it will crash because of beyond array.(sorry I forgot delete my code when I screenshot)
6471a414-7a52-45c8-8773-e401e7431729

Also, I found when i do pod install in terminal it will warning:
91345afe-5757-4ae1-b471-4678f0cda083
So I just edit the podfile by Xcode and fix this.


buba447:
Lgtm! Thanks for grabbing this!

PR #28, opened 2017-02-04
Fixed Carthage support for Swift projects

Jauzee:
Actual master branch builds with Carthage but project (on Swift) can't access `LAAnimationView` and `LAAnimationTransitionController` because of umbrella header doesn't import them.

### DIFF
--- a/Lottie/Lottie.h
+++ b/Lottie/Lottie.h
@@ -16,4 +16,4 @@ FOUNDATION_EXPORT const unsigned char LottieVersionString[];
 
 // In this header, you should import all the public headers of your framework using statements like #import <Lottie/PublicHeader.h>
 
-
+#import <Lottie/Lottie.h>


buba447:
Renamed Lottie.h? Wouldn't this break the example project? 
Couldn't you create public header and import Lottie there?

Jauzee:
@buba447 I renamed that file back to `Lottie.h` [here](https://github.com/airbnb/lottie-ios/pull/28/commits/7bd98b6b4250b3f75f500c2144ff24a3b4664915)
 and imported it to framework header as `<Lottie/Lottie.h>`, otherwise Swift project won't see any `LAAnimationView` or `LATransitionController` even with bridging header.

Actually we have only one change in this pull request — import `Lottie.h` from `PublicHeaders` to framework header file.

I can build Example project with this change.

Jauzee:
I'll try to remove that ugly commit.

Jauzee:
@buba447 it's done. I can confirm that Example project builds without any errors.

mbuchetics:
@Jauzee Tried your PR/fork, the example project builds fine but our Swift project still does not find ` LAAnimationView` and ` LAAnimationTransitionController` (using `import Lottie` without any additional bridging header). Can you confirm it is working for you?

buba447:
Thank you!

PR #38, opened 2017-02-06
resolve #16 changed LA prefixes to LOT

### LINKED ISSUE #16: Classes should use a three letter prefix.
"Two-letter prefixes... are reserved by Apple for use in framework classes."
https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/Conventions/Conventions.html

Consider changing LA -> LOT to avoid namespace collisions.

amandahogan:


buba447:
Thanks for grabbing this but I ended up doing the same! Im going to merge mine since it doesn't have any conflicts!!! Sorry for the run-around!

PR #51, opened 2017-02-07
Btw/reorg

buba447:
Reorganize the project and change namespace to LOT


hansemannn:
Following semantic versioning, this is a breaking change for developers and should result in a major version bump (2.0.0). Sorry for being fussy, just for the future.

PR #63, opened 2017-02-08
LLAnimationView rename in README.

ryanstout:
Looks like the LA prefix got renamed to LOT at some point.  The README example shows LLAnimationView, but the class name is LOTAnimationView.

Also, thanks for releasing this library.  Made my life way easier.


buba447:
Thank you for grabbing this!

PR #64, opened 2017-02-09
Add Static Library Target

lelandrichardson:
to: @buba447 

This adds a static library target to Lottie. I presume this might be useful for people in the iOS community who aren't using frameworks, but this ends up being really important for the React Native ecosystem where linking with static libraries is expected.

I've tested this out on an example RN project and it worked wonderfully.


baires:
Nice!

buba447:
Thanks for grabbing this @lelandrichardson !

PR #69, opened 2017-02-09
LOTAnimationView intrinsicContentSize

AlexQuinlivan:
The sceneModel has an intrinsic content size already given by the composition, so it makes sense that the animation view should use that size as it's intrinsic content size to help with laying out the view.


buba447:
Awesome! Thanks for grabbing this!

PR #77, opened 2017-02-11
Add Community Contributions

m1ga:
The Community Contributions plugins support both platforms so they can be listed here too


buba447:
Can you add https://github.com/airbnb/lottie-ios/pull/10

PR #79, opened 2017-02-11
Resolve Xcode Analyser warnings, throw error when invalid JSON resource specified

hansemannn:
When there is no JSON resource is specified or the specified resource is invalid, Lottie is a bad girl and continues without a valid scene model, causing the Xcode Analyser to complain. I recently resolved the same when an invalid layer is specified, so the Developers will know what's wrong and not receive an empty / stateless view.

To reproduce, just run the Xcode Analyser (`Shift+Command+B`) and follow the nil-path.


buba447:
Looks good!Thanks for setting the standard here!

Issue #80, opened 2017-02-11
Deployment target is too high (builds fail)


dcow:
I cloned the repo and attempted to build the project for iPhone but Xcode reported the project could not be built for my 10.1 iPhone because the deployment target is 10.2. This feels unnecessary.

I did this because building with Carthage was giving me: 

```
Ineligible destinations for the "Lottie" scheme:
		{ platform:iOS, id:dvtdevice-DVTiPhonePlaceholder-iphoneos:placeholder, name:Generic iOS Device }
		{ platform:iOS Simulator, id:dvtdevice-DVTiOSDeviceSimulatorPlaceholder-iphonesimulator:placeholder, name:Generic iOS Simulator Device }
```

After lowering the deployment target, things worked fine.

dcow:
I'm not sure if a workaround would be creating an iPhone simulator with the latest iOS version so that _something_ exists with the a valid deployment target.

dcow:
I suspect updating Xcode might fix the issue, but I'm only one version behind and it's probably not that reasonable to expect everyone to be on the latest Xcode all the time.

buba447:
@dcow Sorry you experienced problems! The version limit wasn't intentional, but a default by Xcode! Im not sure how this happened as I am also on 10.1 (I also take a long time to update myself.) and it runs ok over here. Hmmm.
Ill look into it!

TommyLeong:
So I kicked this "stone" (deployment target is too high) now at year of 2020. What was the followed up action done? The master is still showing 12.0 as deployment target. 

Can lower the deployment version down to 9.0? @buba447

PR #81, opened 2017-02-11
Set a low deployment target

dcow:
This fixes builds for people who do not have the latest and greatest.

cf. #80


buba447:
Looks good! I missed this one! Turns out i was only working from the pods project which has everything set!
My bad!

PR #87, opened 2017-02-13
Update lottie to correct version for cocoapods and npm

baires:


buba447:
Thanks for grabbing this.

PR #91, opened 2017-02-13
Fixed bug setting initial progress

buba447:
This should fix the issue others were seeing with setting the initial progress of a lottie view.

### DIFF
--- a/Example/lottie-ios/AnimatedTextField.m
+++ b/Example/lottie-ios/AnimatedTextField.m
@@ -91,10 +91,10 @@
 - (void)displayCharacter:(BOOL)animated {
   if (animated) {
     [animationView_ play];
-  } else if (animationView_.animationProgress != 1) {
-    [[NSOperationQueue mainQueue] addOperationWithBlock:^{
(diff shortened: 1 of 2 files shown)


simonmitchell:
@buba447 I'm still seeing this issue when using lottie-react-native, I've tried to have a play around with the code and make sure that the `setNeedsAnimationUpdate` function is only called once the animation is actually shown in the window, but it's still exhibiting the same behaviour. Do you have any idea at what point the animation progress needs to be set during the render, it doesn't seem to be once the animation view has a window, so it must be something else 😞 see issue here, that is also complaining of the same thing https://github.com/airbnb/lottie-react-native/issues/122

simonmitchell:
Okay, after more playing around I think I've fixed it, it was a mix of things including:

- didMoveToSuperview being called before the view was in a window when being used in react-native (meaning once it did move to a window the `setNeedsAnimationUpdate` was being blocked by the previous one)
- other animation properties being set on the CALayer before the view was in a window

PR here: https://github.com/airbnb/lottie-ios/pull/245

PR #101, opened 2017-02-16
Fixes an issue with contentsOfURL

rustyshelf:
There's an issue with using initWithContentsOfURL where the view is initialised with a zero sized rect, and doesn’t draw as a result. You can see this in the sample project (by loading from a URL).

### DIFF
--- a/lottie-ios/Classes/Private/LOTAnimationView.m
+++ b/lottie-ios/Classes/Private/LOTAnimationView.m
@@ -228,6 +228,7 @@ const NSTimeInterval singleFrameTimeValue = 1.0 / 60.0;
   if (self) {
     LOTComposition *laScene = [[LOTAnimationCache sharedCache] animationForKey:url.absoluteString];
     if (laScene) {
+      self.frame = laScene.compBounds;
       [self _initializeAnimationContainer];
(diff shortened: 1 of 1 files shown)


lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -303,6 +303,7 @@ - (void)_setupWithSceneModel:(LOTComposition *)model restoreAnimationState:(BOOL
   }
   
   if (_sceneModel) {
+    self.frame = _sceneModel.compBounds;
```
buba447: Can you change this to `self.bounds  = _sceneModel.compBounds;`

rustyshelf:
@buba447 I tried that, but then the view is drawn in the wrong spot. Basically there are issues with using CGRectZero and Auto Layout which just setting the bounds later won't fix, so I've moved the code that sets the frame so that it only affects the one method that needs it `initWithContentsOfURL`. Let me know if that's acceptable or not.

rustyshelf:
Ok, I'm an idiot, I hadn't set up the auto layout constraints properly on the custom view I was making. I'm just going to back away slowly and pretend I didn't just make such an obvious mistake because these changes aren't required...

Issue #14, opened 2017-02-02
Support for image layers


erichoracek:
Thanks for making this awesome tool!

We'd love to use it, but we'd need support for image-backed layers to use our existing after effects animations. From taking a quick glance through the code, it would appear that our image layers are being ignored as they map to the `LALayerTypeUnknown` type. Below is the entirety of one of our simple exports with an image. Here's a [link](https://github.com/bodymovin/bodymovin/blob/master/docs/json/layers/image.json) to a bodymovin image layer schema as well.

I'd be happy to take a stab at this if this is a feature you'd be interested in adding. Thanks!

```
{
    "assets": [{
        "id": "image_0",
        "w": 811,
        "h": 976,
        "u": "images/",
        "p": "img_0.png"
    }, {
        "id": "image_1",
        "w": 1039,
        "h": 1283,
        "u": "images/",
        "p": "img_1.png"
    }],
    "layers": [{
        "ddd": 0,
        "ind": 0,
        "ty": 2,
        "nm": "Link.png",
        "cl": "png",
        "refId": "image_0",
        "ks": {
            "o": {
                "k": 100
            },
            "r": {
                "k": [{
                    "i": {
                        "x": [0.667],
                        "y": [1]
                    },
                    "o": {
                        "x": [0.333],
                        "y": [0]
                    },
                    "n": ["0p667_1_0p333_0"],
                    "t": 60,
                    "s": [0],
                    "e": [-5]
                }, {
                    "i": {
                        "x": [0.667],
                        "y": [1]
                    },
                    "o": {
                        "x": [0.333],
                        "y": [0]
                    },
                    "n": ["0p667_1_0p333_0"],
                    "t": 65,
                    "s": [-5],
                    "e": [5]
                }, {
                    "i": {
                        "x": [0.667],
                        "y": [1]
                    },
                    "o": {
                        "x": [0.333],
                        "y": [0]
                    },
                    "n": ["0p667_1_0p333_0"],
                    "t": 70,
                    "s": [5],
                    "e": [-5]
                }, {
                    "i": {
                        "x": [0.667],
                        "y": [1]
                    },
                    "o": {
                        "x": [0.333],
                        "y": [0]
                    },
                    "n": ["0p667_1_0p333_0"],
                    "t": 75,
                    "s": [-5],
                    "e": [5]
                }, {
                    "i": {
                        "x": [0.667],
                        "y": [1]
                    },
                    "o": {
                        "x": [0.333],
                        "y": [0]
[...]

PR #130, opened 2017-02-26
Fix animated rotation of views

erichoracek:
First of all, thanks for adding image support! This brings us much closer to being able to use Lottie in our app!

However, as far as I can tell, when a layer has its [rotation](https://github.com/bodymovin/bodymovin/blob/master/docs/json/helpers/transform.json#L44-L56) animated with keyframes, this unfortunately has no effect currently. I have seen this behavior after attempting to view the animation posted in #14.

As far as I can tell, this is due to the `sublayerTransform` is being used to perform layer rotation, rather than `transform`. If I’m interpreting the code correctly, this decision was made due to the idea that the `transform` keypath could not have both its `scale` and `rotation` animated concurrently. However, I don’t believe that’s an issue.

I’ve updated both the `rotation` and `scale` properties to animate using the `transform` property, which seems to have fixed the issue with our animation.

Thanks!

### DIFF
--- a/lottie-ios/Classes/AnimatableProperties/LOTAnimatableBoundsValue.m
+++ b/lottie-ios/Classes/AnimatableProperties/LOTAnimatableBoundsValue.m
@@ -171,9 +171,9 @@
   return (self.boundsKeyframes.count > 0);
 }
 
-- (nullable CAKeyframeAnimation *)animationForKeyPath:(nonnull NSString *)keypath {
+- (NSArray<CAKeyframeAnimation *> *)animationsForKeyPath:(NSString *)keypath {
   if (self.hasAnimation == NO) {
-    return nil;
+    return @[];
   }
   CAKeyframeAnimation *keyframeAnimation = [CAKeyframeAnimation animationWithKeyPath:keypath];
   keyframeAnimation.keyTimes = self.keyTimes;
@@ -182,7 +182,7 @@
   keyframeAnimation.duration = self.duration;
   keyframeAnimation.beginTime = self.delay;
   keyframeAnimation.fillMode = kCAFillModeForwards;
-  return keyframeAnimation;
+  return @[keyframeAnimation];
 }
 
 - (NSString *)description {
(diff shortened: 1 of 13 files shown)


lottie-ios/Classes/AnimatableProperties/LOTAnimatableBoundsValue.m
```diff
@@ -182,7 +182,7 @@ - (nullable CAKeyframeAnimation *)animationForKeyPath:(nonnull NSString *)keypat
   keyframeAnimation.duration = self.duration;
   keyframeAnimation.beginTime = self.delay;
   keyframeAnimation.fillMode = kCAFillModeForwards;
-  return keyframeAnimation;
+  return @[keyframeAnimation];
```
buba447: Why are you wrapping in an array here?

erichoracek: @buba447 Thanks for taking a look at this!

I've updated the signature of the `LOTAnimatableValue` method for querying animations to be able to return more than one animation (with no animation is now represented by an empty array).

This behavior is required to ensure that `LOTAnimatableScaleValue` returns more than one `CAKeyframeAnimation`: one to animate `transform.scale.x` and another to animate `transform.scale.y`. Whereas previously a single animation with a keypath of `transform` was returned, we now only animate a single property of the `transform` (the scale). This allows for the rotation to be animated separately via the `transform.rotation.z` keypath without clobbering the transform.scale`.

Hope that makes sense—let me know if anything needs additional clarification!

buba447:
Hmmm. I can't figure out why, but it seems to break with my image form text (attached below)
I also see that with master both rotation and anchor animation is broken.

[ImageXform.zip](https://github.com/airbnb/lottie-ios/files/805189/ImageXform.zip)

buba447:
I figured it out on to add support for anchor and rotation. 
I went ahead and did them both on my branch over here
https://github.com/airbnb/lottie-ios/pull/137

Thanks for bringing this to light!
We should definitely move away from using that sublayer transform, specially if we want to add support for 3d layers. Im going to keep it for now and mark it for a structural refactor thats coming.

erichoracek:
#137 appears to have fixed the issue on my branch. Thanks for taking a look into this.

PR #144, opened 2017-03-01
Fixes attempt assign NSString to NSNumber.

valeriyvan:
Fixes attempt assign NSString to NSNumber.
Adds missing nullability annotations.


buba447:
Thanks for grabbing this!

PR #147, opened 2017-03-01
Add more warnings

xinsight:
Added a few warnings so developers have some info why the Lottie animation might differ from what is in bodymovin or After Effects.

- image not found (sidebar: would be so nice if bodymovin would use the actual image or layer name instead of `img_0.png`! If you have more than one animation that uses images in your app, you have to rename the images in the json manually. Yuck.)
- using expressions to animate values (supported in bodymovin, but not in Lottie)
- effects (e.g. motion blur, particles) seem to be easy to add in After Effects, but are not supported by Lottie.

Also, I didn't see any information about "Effects" being unsupported in the readme - should that also be added?


xinsight:
Added another warning for a "merge shape" as well as a warning for any unknown/unsupported shape.

buba447:
This is great! Thank you so much for doing this bit of housecleaning! Should really help out!

PR #148, opened 2017-03-02
add factory method with bundle param

jqsilver:
I'd like to specify a different bundle when using the `animationNamed` factory method. I parameterized the `bundle` and passed in the main bundle from the old method. 

cc @buba447


buba447:
Thanks for grabbing this!

PR #154, opened 2017-03-02
Annotates unused functions return value as __unused

valeriyvan:
Annotates unused functions return value as __unused making static analyzer happy.


buba447:
Thanks for grabbing this!

PR #155, opened 2017-03-02
Removes unused ivar which was hidden by params in some methods.

valeriyvan:
This makes static analyzer happier.


buba447:
Thanks!

Issue #134, opened 2017-02-27
Crash in Demo: typing other characters


sanatory:
try to type any other letter, not present in english - any russian character, any letter with umlaut( ö, ê, ...) you'll get ResourceNotFoundException

PR #158, opened 2017-03-03
Fixes bug leading to crash of example app

valeriyvan:
Fixes bug leading to crash of example app on attempt to load url from gibberish string, e.g. containing emoji or cyrillic.

Solves issue #134


buba447:
Thanks for grabbing this!

PR #164, opened 2017-03-06
not noly support json get from bundle ,but also support image get from bundle

upworldcjw:


upworldcjw:
support image get from bundle

buba447:
Could you remove the version bump and the file renames?
We would like to control the version bumping so we can include other changes.

Issue #170, opened 2017-03-13
Adding animation to UICollectionViewCell/UITableViewCell


petoye:
I need to know a way to add a LOTAnimationView to a collectionView prototype and then link it to the collectionViewCell. The animation will play when you press a button on the cell. Any ideas?

xinsight:
You can add an `LOTAnimationView` to a `UICollectionViewCell` like any other custom view.

petoye:
I did try doing that but then I don't know how to exactly play the animation

xinsight:
Ah, I guess you're using a storyboard. Perhaps someone else can confirm, but It looks like you can't add create `LOTAnimationView`s from interface builder.

Easiest workaround is to create a `UICollectionViewCell` subclass, and then add the animation in code. Something like:

```
import UIKit
import Lottie

class AnimCollectionViewCell: UICollectionViewCell {

    var anim: LOTAnimationView!
    
    override func awakeFromNib() {
        super.awakeFromNib()
        
        anim = LOTAnimationView(name: "LottieLogo1")
        anim.contentMode = .scaleAspectFit
        self.contentView.addSubview(anim)
        
        anim.centerXAnchor.constraint(equalTo: self.contentView.centerXAnchor)
        anim.centerYAnchor.constraint(equalTo: self.contentView.centerYAnchor)
        
    }
    
    public func play() {
        anim.play()
    }

}
```

AmanToshniwal:
Thank you so much, I was searching all over for this and your solution did wonders.

Issue #161, opened 2017-03-04
LOTAnimationView should catch exceptions from parsing JSON


benpious:
`LOTAnimationView.animationFromJSON` performs no validation of the contents of `animationJSON` which can result in an Objective-C Exception being thrown if the objects in the keys it reads aren't the types that it expects. 

One doesn't typically expect an exception to be thrown while instantiating a UIView subclass, and this is particularly inconvenient behavior in Swift, given that Swift actually can't catch Objective-C exceptions. Also, the potential exists for invalid JSON to *not* cause a crash, but instead to silently not produce any animation at all; as far as I'm aware there is no way to detect this at runtime. 

An easy solution to this problem would be to catch the exceptions inside of `_mapFromJSON`, and simply return `nil` from `initWithJSON`. If combined with #160, this would allow the caller to easily respond to invalid JSON from either Swift or Objective-C by providing a placeholder or taking other appropriate actions, since by convention in both Swift and Objective-C, constructors that take arguments that can be invalid are usually failable.

buba447:
Lottie V2 should fix this.

Issue #160, opened 2017-03-04
Changes to LOTComposition and LOTAnimationView interfaces


benpious:
I'd like to propose two changes to Lottie; the first is allowing the user to set the animation on an `LOTAnimationView` after instantiating it, and the second is making a subset of `LOTComposition`'s API public. 

There's several elements of `LOTAnimationView` which make it difficult to work with. Firstly, you can't create an `LOTAnimationView` without first creating whatever animation data you want to display. 

For example, consider this example of the normal flow for getting an image off the network and displaying it:

```
urlSession.getData { [weak self] data in 
   if data {
       self?.imageView.image = UIImage(data)
   }
}
```

Lottie, however, provides no way to instantiate an `LOTAnimationView` and subsequently change the animation (or even to instantiate a functional `LOTAnimationView` instance without having the animation data on-hand). Therefore, you'd have to instantiate it when the animation data comes in. 

This isn't ergonomic for several reasons; it requires you to use an optional for your view property in Swift (unless you wrap `LOTAnimationView` in a wrapper), it may require you to break up autolayout code to only apply constraints after the view has come in, and it may require you to create a new `LOTAnimationView` (resetting all of the associated state) if you want to change animations; for example if you were using an `LOTAnimationView` in a collection view or tableview with reusable cells.

In my local branch, I found that only minor changes were required to support this; the setter was as simple as

```
- (void)setSceneModel:(LOTComposition *)sceneModel {
    _sceneModel = sceneModel;
    [self _setupWithSceneModel:sceneModel restoreAnimationState:NO];
    [self invalidateIntrinsicContentSize];
}
```

and properly overriding `initWithFrame` to call`_initializeAnimationContainer`, though there may be cases I'm missing in my approach. 

Since the `LOTAnimationView` actually works just fine without animation data until it's provided. I suppose there's issues with what happens if `LOTAnimationView.play` is called before animation data is set, but this could be dealt with with a `shouldPreserveAnimationState` property allowing the user to configure behavior when the scene model changes. 

Another issue is the decision to make `LOTComposition` private to the module. This prevents the caller from inflating compositions on a thread of their choice, which, given the size of the JSON files produced by animations, seems to be a potential performance issue. My profiling on the iPhone 7 simulator shows that it takes 4.0 ms to convert a fairly simple animation into an `LOTComposition`, which is 1/4 of a frame. I think it would be better to publicly expose `LOTComposition` and its initializer, so the user can construct one however they like. 

[...]

PR #172, opened 2017-03-14
Set animation after initialization

xinsight:
Prompted by issue #170, I attempted to layout a lottie animation using a storyboard and it wasn't possible since the animation could only be set in an initializer. This PR extracts some code out of the initializer so that the animation could be set after initialization. (This refactoring is also outlined in #160.)

For example, the animation can now be set with:

`self.lottieLogo.animationName = @"LottieLogo1";`

I've also a warning if the animation name contains a period, and creates an error if the animation is not found. (I've followed the existing convention of throwing an exception, but as mentioned in #161, it might be better to just log something and return a nil.)

I changed the example code to use the storyboard for layout as the programmatic calling style can still be seen in other places in the example app.

Let me know if you have any questions. Thanks!


Example/lottie-ios/LottieRootViewController.m
```diff
@@ -74,7 +61,7 @@ - (void)_playLottieAnimation {
   [self.lottieLogo play];
 }
 
-#pragma mark -- UITableViewDataSource and Delegate
+#pragma mark - UITableViewDataSource and Delegate
```
xinsight: btw, i fixed these since the `--` shows up as "-" in the Xcode code navigator (pulldown), not a line.

buba447:
I think a better interface would be to expose the LOTComposition as a settable property and then have it set manually. You'd have to expose the LOTComposition header file publicly as well.
This would be more in keeping with how UIImageView works. 
Maybe move the loading from the bundle over to [LOTComposition compositionNamed: inBundle:]
Having a string property for animationNamed seems a little weird.

Also if you are allowing LOTAnimationView to be allocated without an animation you have to setup a blank animationState in order to record any state that is set on the view before the animation become available.

xinsight:
Thanks for the feedback.

> Having a string property for animationNamed seems a little weird.

Yeah, I'm not fond of string APIs in general, but if you add an `@IBDesignable` to the animationName property, then the animation could be set in the storyboard. (However the animation would still have to be started in code - which means you still need the hassle of wiring up the storyboard to the code - so maybe that isn't so amazeballs after all.)

I'll try to find some time today to expose `LOTComposition` and load the animation there.

xinsight:
@buba447: I've exposed `LOTComposition` but I'm not sure about your comment to make a composition a property.

The calling style would be:
```
LOTComposition *comp = [LOTComposition compositionNamed:@"Foo"];
animationView.composition = comp;
```

Since the callers will rarely be using `LOTComposition`, it seems like we're exposing needless complexity.

The current API is:
```
[LOTAnimation animationNamed:@"Foo"];
```

So, setting the animation name after instantiation follows that convention:

```
animationView.animationName = @"Foo";
```

I'm also not sure how the animation state works (or why it is in the AnimationView file) or what it means to create an anim state before we have an animation loaded. Can you elaborate on that?

buba447:
Exposing the composition object separates out the concerns of loading the data for the animation and setting the animation. This will help a lot of people in situations othe than your own. Animations can be loaded from bundle, or from web or cached independently of the Lottie framework. Just exposing as settable string doesnt offer anymore flexibility, it only allows you to delay the existing initializaion calls.

The animaiton state is the class that records the animation views state, such as wether it is playing, or loops, or if its set to an existing progress. This records state and then applies it to the layer to ensure proper state. If an animation hasnt been created yet we create a state without a layer and it records any state changes performed externally until the animation is set. EG you are asynchronously downloading an animaiton and set .play() on the animationview the animation will start playing once the download is finished.
An example of this can be found inside of the initWithURL method of animation view.

Issue #142, opened 2017-02-28
Completion block not called after changing loopAnimation flag


christoph-eero:
Use case:  We have animations that have an intro, loop and outro that are supposed to be strung together seamlessly.  Since we cannot loop on parts of a Lottie animation, a workaround was to swap views as animations end.  The middle loop is configured with loopAnimation = true and a completion block. 

The block is not called while the animation is looping which is logically consistent and expected.  However, when changed to loopAnimation = false it should still call the block because it was played with a block.

1.  Create and lay out an animation view
2.  Set view.loopAnimation = true
3.  Call view.playWithCompletion() with a valid block
4.  Set a breakpoint in the block
5.  Programmatically set view.loopAnimation = false

Expected result:  View finishes current animation, then calls block
Actual result:  View finished current animation, does not call block

buba447:
Ohhh when you set loop animation LOT removes the reference to the completion block. This was to avoid keeping completion blocks forever. 

Hmmm i can see that this is a pretty valid use case though.

christoph-eero:
Yup, I just found the line of code that does it and commented it out.  To avoid keep unused completion blocks, how about only create the _completionDisplayLink if playWithCompletion() was called?

PR #174, opened 2017-03-16
Allow animation to stop looping via flag [#142]

jmyers0022:
I'm running into the same use case as @christoph-eero in https://github.com/airbnb/lottie-ios/issues/142.

These changes have allowed me to set the `loopAnimation` flag, which when set to false on a looping animation will stop at the end of the current loop and call the completion callback.

Let me know if there are any ramifications to doing this. It looks like `[self stopDisplayLink]` is being called later on down the line.

@buba447 - We're also using this on an Android project if we can get this figured out. Is this something that can be easily implemented on the Android side?


buba447:
I think this looks good!

PR #181, opened 2017-03-21
Fix @import compatibility with .mm files

singro:
.mm does not support `@import` module syntax which could cause a compile error when import `Lottie.h` to a .mm file or projects that modules disabled.


buba447:
Thanks for grabbing this!@

Issue #180, opened 2017-03-20
README.md broken


chaseacton:
Some of the markdown in the readme is broken. This is because GitHub formalized their markdown standards and headings now require a space after "#".

https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown

![screen shot 2017-03-20 at 1 40 01 pm](https://cloud.githubusercontent.com/assets/3307194/24120692/be8ea3f6-0d72-11e7-8811-8d8b0e4f4c81.png)

madsonic:
I believe there is a typo too
`LOTAnimatedView.animationNamed("here_goes_your_json_file_name_without_.json")`
but should be 
`LOTAnimationView(name "here_goes_your_json_file_name_without_.json")`

PR #185, opened 2017-03-22
Fixed README.md typo error

tapanprakasht:
Fix the typo error in README.md mentioned in the #180 by @madsonic
Method signature gave the README.md was wrong


letsgogeeky:
It doesn't require the `name:` while passing the argument since it's a single parameter function.

tapanprakasht:
@letsgogeeky without name: parameter it is giving an error. It looks like parameter is needed. I tried with Swift 3

letsgogeeky:
Yeah got it! with swift 3 it's required!

buba447:
Can you make a note about swift 3? @tapanprakasht

Issue #198, opened 2017-03-30
after addSubview toLayerNamed, layer position will shift down


focuspirit:
like doc:
```
let snapshot = view.snapshotView(afterScreenUpdates: true)!
snapshot.backgroundColor = UIColor.blue
animationView.addSubview(snapshot, toLayerNamed: "layerName Outlines")
```

PR #110, opened 2017-02-21
Add support for Precomps

buba447:

PR #205, opened 2017-04-04
Fix bounds calculation for masked sublayers

Pretz:
This fixes an issue where views added via `addSublayer:toLayerNamed:` would receive incorrect frames.

It looks like when the bounds calculation for `LOTCustomChild` was moved from `LOTAnimationView` to `LOTCompositionLayer` in #110, the reference layer for the bounds of the child layer didn't get updated correctly. This is visible in the sample transition animations included with `lottie-ios-Example`.

Additionally, this removes the mask layer from its superlayer, as CALayer.mask's docs say `The layer you assign to this property must not have a superlayer. If it does, the behavior is undefined.`

### Before:
apr-04-2017 14-37-17

### After:
apr-04-2017 14-35-50


Pretz:
Looks likely this will fix #198, although it's hard to be certain without a specific example.

PR #209, opened 2017-04-08
let Lottie support custom filePath resource

liuzhiyi1992:


buba447:
I think we are going to go with the solution found here https://github.com/airbnb/lottie-ios/pull/223
Thanks for the work!

PR #216, opened 2017-04-13
Added custom images support

gralexey:
Now you can supply an animation in runtime with your own images.


lottie-ios/Classes/Models/LOTAssetGroup.h
```diff
@@ -8,12 +8,14 @@
 
 #import <Foundation/Foundation.h>
 #import <CoreGraphics/CoreGraphics.h>
+#import <UIKit/UIKit.h>
```
buba447: This will break MacOS Support.

PR #242, opened 2017-05-21
Allow consumer to control clipping behavior

swift-ortal:
Currently when you want to allow items to be drawn outside of the LOTAnimationView's frame, you must set two properties:

1. `LOTAnimationView.clipsToBounds = false`
1. `LOTCompositionLayer.masksToBounds = false`

Doing the former is easy, as that is how UIKit is built: 

```
lotAnimationView.clipsToBounds = false
```

Doing the latter is more difficult, since the LOTCompositionLayer is not exposed, and we're forced to mess around with the internals of LOTAnimationView:

```
if let lotCompositionLayer = lotAnimationView.layer.sublayers?.first?.sublayers?.first {
  assert(String(describing: type(of: lotCompositionLayer)) == "LOTCompositionLayer", "Make sure we're modifying the correct layer")
  lotCompositionLayer.masksToBounds = false
}
```

Instead, I propose we remove the line of code that sets `masksToBounds = YES` in `LOTCompositionLayer`, so that it uses the default `NO` instead.

This will allow the consumer to control the clipping behavior by a single, exposed property `LOTAnimationView.clipsToBounds` rather than having to muck with an internal setting, and/or set two different properties.

This shouldn't introduce problems because it seems like setting `masksToBounds` on the layer wasn't adding any new functionality, since `clipsToBounds` was already being controlled by the view.

# Screenshots

Notes:

- In both cases, I manually set `lotAnimationView.clipsToBounds = false`
- The red background represents the LOTAnimationView's frame

**Before**: (Notice how the bottom left image is incorrectly clipped)

screen shot 2017-05-21 at 10 50 12 am

**After**: (Notice how the bottom left image is correctly no longer clipped)

screen shot 2017-05-21 at 10 50 54 am

### DIFF
--- a/lottie-ios/Classes/AnimatableLayers/LOTCompositionLayer.m
+++ b/lottie-ios/Classes/AnimatableLayers/LOTCompositionLayer.m
@@ -36,7 +36,6 @@
                         withBounds:(CGRect)bounds {
   self = [super init];
   if (self) {
-    self.masksToBounds = YES;
     [self _setupWithLayerGroup:layerGroup withAssetGroup:assetGroup withBounds:bounds];
   }
   return self;


swift-ortal:
By using a wider asset and a content mode of aspectFill, this PR is rendered unneeded. It would have also caused sub-composition layers to be unclipped incorrectly.

PR #256, opened 2017-06-12
Fix for layer out frame

buba447:
The out time for layers was a frame late. This was a simple mathematical error. Fixed now!


gnarlo:
For layers that only exist for 1 frame, the layer length and duration are now set to 0 with this change, and never show up at all in the animation.

Issue #235, opened 2017-05-09
How to clear LOTAnimationCache


LaughingJack:
What is the suggested mechanism for clearing the animation cache? I currently have a pair of somewhat large animations that I only need to run once and don't want to keep around from memory management purposes. How it it suggested to clear out the cache and free up our run time memory?

buba447:
Ohh! this is a good suggestion. Will added it to future enhancement.
The cache definitely needs an upgrade to support different caching strategies. Sorry!

bellchen:
any schedule about this? Very eager to get this function

Issue #271, opened 2017-06-22
Pointer is missing a nullability type specifier in 1.5.2 version


andreyreddit:
After updating to 1.5.2 I see many errors like this: 
Pointer is missing a nullability type specifier (_Nonnull, _Nullable, or _Null_unspecified)
Project stopped compiling.

buba447:
Should be fixed in Lottie 2.0! If you still have a problem please create a new issue!

PR #273, opened 2017-06-22
user can remove animationView from cache now!

bellchen:
user can remove animationView from cache by calling:  animationView.cacheEnable = NO;
or add into cache again by set cacheEnable YES
fix the issue #235


bellchen:
fix issue #271

buba447:
Great! Thank you for grabbing this!

PR #275, opened 2017-06-26
support load image from bundle

gejw0623:


bellchen:
You've changed a lot, but they don't have anything to do with "support load image from bundle"

buba447:
Should be fixed in Lottie 2.0! If you still have a problem please create a new issue!

PR #286, opened 2017-06-30
Add nullability specifiers to LOTAnimationTransitionController

knt:
Adds nullability specifiers in LOTAnimationTransitionController which are needed for better compatibility with Swift.

Also fixed indention.

### DIFF
--- a/lottie-ios/Classes/AnimationCache/LOTAnimationCache.h
+++ b/lottie-ios/Classes/AnimationCache/LOTAnimationCache.h
@@ -15,6 +15,7 @@
 + (instancetype)sharedCache;
 
 - (void)addAnimation:(LOTComposition *)animation forKey:(NSString *)key;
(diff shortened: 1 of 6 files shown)


tinder-ntreadway:
Is NSBundle meant to be nullable? 

I noticed this comment:
```
 @param bundle custom bundle to load animation and images, if no bundle is specified will load
 from mainBundle
```

I can change it if so.

knt:
Is NSBundle meant to be nullable?

I noticed this comment:

```
 @param bundle custom bundle to load animation and images, if no bundle is specified will load
 from mainBundle
```

I can change it if so.

robnadin:
I think it is OK to keep it annotated as `nonnull`, as it is an overloaded function call for `initWithAnimationNamed:fromLayerNamed:toLayerNamed:` which calls through with the main bundle.

Issue #184, opened 2017-03-22
The animation is delayed


rickytan:
Hi all,
  have you ever faced this problem, I started the animation immediately after use clicked a button, but the animation was delayed. See the attached file:
![lottie-bug](https://cloud.githubusercontent.com/assets/1250207/24180668/4290248e-0ef1-11e7-8bc9-7cd4e51b5224.gif)

maurovc:
+1

upworldcjw:
+1, after i playing short movie, the lottie animation is delay.  the following  log has some problem????

- (BOOL)animationIsPlaying {
  if (_animationIsPlaying && !_loopAnimation && _layer) {
    CFTimeInterval localTime = [_layer convertTime:CACurrentMediaTime() fromLayer:nil];
    NSLog(@"CFTimeInterval = %f",localTime);

2017-04-28 15:08:27.838266 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.854894 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.871650 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.888364 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.906218 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.922026 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.947379 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.955228 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.976928 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:27.991014 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:28.005313 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:28.021863 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:28.038652 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:28.055016 inke[18132:5888874] CFTimeInterval = 0.000000
2017-04-28 15:08:28.071675 inke[18132:5888874] CFTimeInterval = 0.000000
[...]

PR #297, opened 2017-07-07
fix https://github.com/airbnb/lottie-ios/issues/184

kevinsan:
adjust the containerLayer beginTime


rickytan:
This code works! 👍

PR #305, opened 2017-07-14
Allow custom image cache for 'animationWithFilePath'

panmingyang2009:
When creating lottie view by using 'animationWithFilePath:', the images will not be cached. Thus, when creating a couple of lottie view from same resource, it cost many times of memory than it should be.
So, I wrote a cache provider to allow a third party cache. Other developers can use their own image cache in lottie.


panmingyang2009:
The using of `#if TARGET_OS_IPHONE || TARGET_IPHONE_SIMULATOR` lead to a compile error.  
I can't figure out **why** it happened.🤔
But I have **solved** the problem in my last commit (`03beae5`).

I will be very grateful if some one can tell me why.
😊

buba447:
This is really great! Thank you! Ill look into that compiler warning. I suspect it has something to do with out MacOS compatibility layer.

panmingyang2009:
Thanks for your reply.

 I‘ve solved the merge conflict and reorder the commit messages. 
Looking forward for merging.

buba447:
Hey! Sorry for this but we just merged Lottie V2 which was a pretty big change to everything. How hard is it to work this into the new version?

panmingyang2009:
Hello! I'll try it as soon as possible. 
Can't wait to try the new version! 👍

panmingyang2009:
I found that I used wrong account to push my commits. So I restored the branch and the PR closed automatically.😵  I'll open a new PR.

PR #307, opened 2017-07-14
Feature/custom images support

gralexey:
Now you can supply your animation with your custom images in runtime! (+macOS support)


gralexey:
@buba447 Can I ask why it's closed?

buba447:
Sorry to leave you hanging! We just completely blew away the old rendering system and a few things got left behind when we merged it.
Another pull request added custom images through the form of image caches, see it here https://github.com/airbnb/lottie-ios/pull/336

I see that https://github.com/airbnb/lottie-ios/pull/336 doesnt offer mac support though. Im not in favor of adding a blank UIImage class though, maybe a class alias for UIIimage/NSImage called LOTImage like what we did here https://github.com/airbnb/lottie-ios/blob/master/lottie-ios/Classes/PublicHeaders/LOTAnimationView_Compat.h#L17

If you want to make a new PR that does that I think that should cover the same thing.

PR #330, opened 2017-08-02
Expose completionBlock

rob-keepsafe:
Replaces #316 

### Background

PR #174 was a good first step towards enabling the chaining of animations, but it became apparent that the `completionBlock` should be decoupled from setting only via `play()` or else you'd end up with a very large nesting of animations.

Consider this a quick fix (only applicable for looping scenarios where you can explicitly control timing), but might be interesting to allow more explicit chaining or stitching together of animations in v2.0.

### Example

Let's say you were creating a network spinner and needed a continuously rotating icon to eventually turn into a different icon after the network response returned success/failure.

Currently, you can create the looping animation and then when the network response returns, you set looping to `false` and it will call the completion block. This is fine, but if you also have many other animations, it soon becomes very messy and very nested since you are defining async code (e.g. the completion block) far away from its area of concerns.

If you could set the completion block on a given animation any time and then stop looping, it allows you to isolate the code better and keep your nesting to maybe one or two levels deep maximum.

### DIFF
--- a/lottie-ios/Classes/Private/LOTAnimationView_Internal.h
+++ b/lottie-ios/Classes/Private/LOTAnimationView_Internal.h
@@ -17,7 +17,6 @@ typedef enum : NSUInteger {
 @interface LOTAnimationView () <CAAnimationDelegate>
 
 @property (nonatomic, readonly) LOTComposition * _Nonnull sceneModel;
-@property (nonatomic, copy, nullable) LOTAnimationCompletionBlock completionBlock;
 @property (nonatomic, copy, nullable) NSString *cacheKey;
 
 @end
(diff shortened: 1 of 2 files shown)


buba447:
This exposes the completion block but will not do what you intend. Currently I dont think that setting loopAnimation to `false` will do anything at all if the animation is already in progress.
Ill need to fix that!

rob-keepsafe:
Got it. In Lottie v1, setting `loopAnimation` to `false` will allow the animation to finish from its current progress and then call the `completionBlock` so I'll check if that behavior was maintained

buba447:
@rob-keepsafe I looked into it and fixed the issue with this https://github.com/airbnb/lottie-ios/pull/332

rob-keepsafe:
@buba447 Thanks again!

PR #374, opened 2017-08-16
fix the Memory leak in LOTBezierPath.m file

Calvix-Xu:
There is a while loop in the method trimPathFromT: toT: offset:  in lottie-ios/Classes/Extensions/LOTBezierPath.m. Because the break keyword make the loop exit，and the  less LOT_Subpath memory couldn't free. so to fix this problem, add the code to free the less LOT_Subpath memory before break.


buba447:
Great catch! Shew how did i miss that?

PR #377, opened 2017-08-17
Add better support for negative animation speeds

welshm:
- The progress can be calculated basd on negative speed
  + The end frame would be a progress of 0 and the start frame a progress of 1
  + Recalculate progress when the speed changes, otherwise a "jump" will be seen when restarting animation

- The frame for the progress can be calculated using the newly inverted progress
  + Determine the absolute progress based on current progress and determine which frame to use based on animation speed

- This fixes issues of animations playing with negative speeds "bouncing" back to their final position
  which was caused by the final frame being calculated based on a completed progress of 1 which then did not take into
  account the negative animation speed.

### DIFF
--- a/lottie-ios/Classes/Private/LOTAnimationView.m
+++ b/lottie-ios/Classes/Private/LOTAnimationView.m
@@ -210,27 +210,14 @@ static NSString * const kCompContainerAnimationKey = @"play";
   if (!_sceneModel) {
     return 0;
   }
-  CGFloat absoluteProgress = ((frame.floatValue - _sceneModel.startFrame.floatValue) / (_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue));
-  if ([self _isPlayingForwards]) {
-    return absoluteProgress;
-  } else {
-    // If the animation is playing backwards, the progress is inverted.
-    return 1 - absoluteProgress;
-  }
+  return ((frame.floatValue - _sceneModel.startFrame.floatValue) / (_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue));
 }
 
 - (NSNumber *)_frameForProgress:(CGFloat)progress {
   if (!_sceneModel) {
     return @0;
   }
-  CGFloat absoluteProgress = ((_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue) * progress);
-  if ([self _isPlayingForwards]) {
-    // If we're moving forward, then add the absolute progress to the start.
-    return @(absoluteProgress + _sceneModel.startFrame.floatValue);
-  } else {
-    // It the animation is playing backwards, subtract the progress from the end.
-    return @(_sceneModel.endFrame.floatValue - absoluteProgress);
-  }
+  return @(((_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue) * progress) + _sceneModel.startFrame.floatValue);
 }
 
 - (BOOL)_isPlayingForwards {
@@ -397,18 +384,11 @@ static NSString * const kCompContainerAnimationKey = @"play";
 }
 
 -(void)setAnimationSpeed:(CGFloat)animationSpeed {
-  BOOL directionChange = NO;
-  if ((animationSpeed >= 0 && _animationSpeed < 0) || (animationSpeed < 0 && _animationSpeed >= 0)) {
-    directionChange = YES;
-  }
   _animationSpeed = animationSpeed;
-  NSNumber *frame = [_compContainer.presentationLayer.currentFrame copy];
   if (_isAnimationPlaying && _sceneModel) {
+    NSNumber *frame = [_compContainer.presentationLayer.currentFrame copy];
     [self setProgressWithFrame:frame callCompletionIfNecessary:NO];
     [self playFromFrame:_playRangeStartFrame toFrame:_playRangeEndFrame withCompletion:self.completionBlock];
-  } else if (directionChange) {
-    // The progress needs to be re-calculated if the speed direction changes if the animation is not playing.
-    _animationProgress = [self _progressForFrame:frame];
   }
 }
 
--- a/lottie-ios/Classes/Private/LOTAnimationView.m
+++ b/lottie-ios/Classes/Private/LOTAnimationView.m
@@ -220,7 +220,7 @@ static NSString * const kCompContainerAnimationKey = @"play";
   return @(((_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue) * progress) + _sceneModel.startFrame.floatValue);
 }
 
-- (BOOL)_isPlayingForwards {
+- (BOOL)_isSpeedNegative {
   // If the animation speed is negative, then we're moving backwards.
   return _animationSpeed >= 0;
 }
@@ -302,7 +302,7 @@ static NSString * const kCompContainerAnimationKey = @"play";
   NSNumber *currentFrame = [self _frameForProgress:_animationProgress];
 
   currentFrame = @(MAX(MIN(currentFrame.floatValue, toEndFrame.floatValue), fromStartFrame.floatValue));
-  BOOL playingForward = [self _isPlayingForwards];
+  BOOL playingForward = [self _isSpeedNegative];
   if (currentFrame.floatValue == toEndFrame.floatValue && playingForward) {
     currentFrame = fromStartFrame;
   } else if (currentFrame.floatValue == fromStartFrame.floatValue && !playingForward) {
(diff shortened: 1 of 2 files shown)


lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -208,14 +210,32 @@ - (CGFloat)_progressForFrame:(NSNumber *)frame {
   if (!_sceneModel) {
     return 0;
   }
-  return ((frame.floatValue - _sceneModel.startFrame.floatValue) / (_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue));
+  CGFloat absoluteProgress = ((frame.floatValue - _sceneModel.startFrame.floatValue) / (_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue));
+  if ([self _isPlayingForwards]) {
+    return absoluteProgress;
+  } else {
+    // If the animation is playing backwards, the progress is inverted.
+    return 1 - absoluteProgress;
```
buba447: _animationProgress should always be in absolute time, not relative time. So if an animation is playing in reverse the progress will decrease, if it plays in autoReverse it will go up then down.

lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -14,6 +14,8 @@
 #import "LOTAnimationCache.h"
 #import "LOTCompositionContainer.h"
 
+static NSString * const kCompContainerAnimationKey = @"play";
```
buba447: Thanks for adding this! Things I always forget to do haha.

lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -208,14 +210,32 @@ - (CGFloat)_progressForFrame:(NSNumber *)frame {
...
   }
-  return @(((_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue) * progress) + _sceneModel.startFrame.floatValue);
+  CGFloat absoluteProgress = ((_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue) * progress);
+  if ([self _isPlayingForwards]) {
+    // If we're moving forward, then add the absolute progress to the start.
+    return @(absoluteProgress + _sceneModel.startFrame.floatValue);
+  } else {
+    // It the animation is playing backwards, subtract the progress from the end.
+    return @(_sceneModel.endFrame.floatValue - absoluteProgress);
+  }
```
buba447: If _animationProgress stays absolute then this will also be fine as it was.

lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -596,12 +626,14 @@ - (void)_layout {
...
-  if ([_compContainer animationForKey:@"play"] == anim &&
+  if ([_compContainer animationForKey:kCompContainerAnimationKey] == anim &&
       [anim isKindOfClass:[CABasicAnimation class]]) {
     CABasicAnimation *playAnimation = (CABasicAnimation *)anim;
     NSNumber *frame = _compContainer.presentationLayer.currentFrame;
     if (complete) {
-      frame = (NSNumber *)playAnimation.toValue;
+      // Set the final frame based on the animation to/from values. If playing forward, use the
+      // toValue otherwise we want to end on the fromValue.
+      frame = [self _isPlayingForwards] ? (NSNumber *)playAnimation.toValue : (NSNumber *)playAnimation.fromValue;
```
buba447: Why not just check the anim.speed here and to the binary switch?
This way the rest of the logic can be removed but it will still have the effect you are looking for?

welshm: Explained my reasoning for the change above. If you disagree, will keep this part of the change and revert the rest. Not sure which binary switch you're referring to, though.

lottie-ios/Classes/Private/LOTAnimationView.m
welshm: My reasoning for inverting this was as follows:

If I have a 10 second animation that has run for 3 seconds and then I set the speed to -1, it would begin reversing the animation and the progress should jump from 30% to 70%.

Same for the frames. If there are 12 frames and I'm at frame 3 going forward, that is 25% completion but going in reverse it's 75% completion.

Otherwise, callers of the SDK who set the speed to negative one need to do all these inversions themselves. If that is the case, I'll keep the constant change above (kCompContainerAnimationKey) and add documentation to the header file about using negative speeds?

lottie-ios/Classes/Private/LOTAnimationView.m
buba447: Im referring to `frame = [self _isPlayingForwards] ? (NSNumber *)playAnimation.toValue : (NSNumber *)playAnimation.fromValue;`

Instead of `[self _isPlayingForwards] ?`
You can do `anim.speed < 0 ?`

lottie-ios/Classes/Private/LOTAnimationView.m
buba447: Yes, but speed implicitly changes the animation. 
If you have auto-reverse animation going and you are tracking the animation in realtime via animationProgress it will be unclear where the animation is unless animationProgress is absolute.

Animation progress is intended to be a 1:1 mapping to the entire animation timeline, not to the currently playing animation.

This comes up in a lot of different states.
When `playFromProgress:0.5 toProgress:1` is called then animationProgress is still absolute and reports a from 0.5 to 1 even though the internal animation 0-1 would map 0.5 to 1.

Another way to reverse playing an animation is to call `playFromProgress:1 toProgress:0` with a positive speed. Then when you reverse the speed the animation would actually play forward.

lottie-ios/Classes/Private/LOTAnimationView.m
welshm: I left it as `[self _isPlayingForwards]` because I think that still makes sense to use in

```
- (void)playFromFrame:(nonnull NSNumber *)fromStartFrame
               toFrame:(nonnull NSNumber *)toEndFrame
        withCompletion:(nullable LOTAnimationCompletionBlock)completion
```

buba447: That seems reasonable. In that case, on last nitpick can you change `_isPlayingForwards` to `_isSpeedNegative` or something similar? Since there are cases where speed can be negative and the animation is still technically playing forward. 

Just to keep confusion down in the future.

lottie-ios/Classes/PublicHeaders/LOTAnimationView.h
```diff
@@ -46,9 +46,12 @@ typedef void (^LOTAnimationCompletionBlock)(BOOL animationFinished);
 
 // TODO
 /// Sets a progress from 0 - 1 of the animation. If the animation is playing it will stop and the compeltion block will be called.
+/// The animation progress is in terms of absolute progress of the defined animation and does not
+/// take into account negative speeds.
```
buba447: Mind changing the wordage here a bit? Currently it seems as if animationProgress does not work if the speed is negative, thats not true.

Something like 'the current progress of the animation in absolute time. EG a value of 0.75 always represents the same point in animaiton regardless of positive or negative speed.'

lottie-ios/Classes/PublicHeaders/LOTAnimationView.h
```diff
@@ -46,9 +46,12 @@ typedef void (^LOTAnimationCompletionBlock)(BOOL animationFinished);
...
 @property (nonatomic, assign) CGFloat animationProgress;
 
-/// Sets the speed of the animation. Accepts a negative value for reversing animation
+/// Sets the speed of the animation. Accepts a negative value for reversing animation.
+/// Negative speeds do not affect animationProgress
```
buba447: Also, not necessarily true. Negative speeds do affect animation progress in that as the animation plays the  progress is updated to its absolute value.

lottie-ios/Classes/Private/LOTAnimationView.m
welshm: Sounds good to me! Thanks for the speedy comments and review

lottie-ios/Classes/PublicHeaders/LOTAnimationView.h
welshm: Sure.

Just a thought... if I call play, the progress goes from 0 to 1 (regardless of play direction) where as setting the value directly will cause a different behaviour so the property does different things depending on if the animation is playing or not.

lottie-ios/Classes/PublicHeaders/LOTAnimationView.h
welshm: Removed comment as I think the comment on animationProgress better captures the intent

lottie-ios/Classes/PublicHeaders/LOTAnimationView.h
welshm: Actually that's not necessarily true - never mind!

buba447:
Great! Thank you for grabbing this and for working with me on it!

buba447:
@welshm Do you mind grabbing those merge conflicts?

buba447:
@welshm Nevermind! I got it!

PR #385, opened 2017-08-23
Reverse support

welshm:
The offset time for an animation should be relative to the direction in which it's playing


lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -310,7 +310,9 @@ - (void)playFromFrame:(nonnull NSNumber *)fromStartFrame
   }
   _animationProgress = [self _progressForFrame:currentFrame];
   
-  NSTimeInterval offset = MAX(0, (_animationProgress * (_sceneModel.endFrame.floatValue - _sceneModel.startFrame.floatValue)) - fromStartFrame.floatValue) / _sceneModel.framerate.floatValue;
+  // If the animation is playing backrwars, to get the offset we need to invert the progress.
+  CGFloat offsetProgress = playingForward ? _animationProgress : 1 - _animationProgress;
```
buba447: This should only be the case if fromFrame is less than toFrame.
If speed is negative then the offset shouldnt make a difference?

welshm: I was finding the offset was causing the animation to jump to the finish before playing. Changing the offset here seems to have fixed it, but maybe it's masking another problem?

welshm:
I tried to squash this to not have so many commits but git didn't seem to like this.

buba447:
This one here got it! https://github.com/airbnb/lottie-ios/pull/424

lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -597,7 +599,7 @@ - (void)animationDidStop:(CAAnimation *)anim finished:(BOOL)complete {
     if (complete) {
       // Set the final frame based on the animation to/from values. If playing forward, use the
       // toValue otherwise we want to end on the fromValue.
-      frame = [self _isSpeedNegative] ? (NSNumber *)playAnimation.toValue : (NSNumber *)playAnimation.fromValue;
+      frame = [self _isSpeedNegative] ? (NSNumber *)playAnimation.fromValue : (NSNumber *)playAnimation.toValue;
```
welshm: Is this change still relevant or worth doing?

PR #396, opened 2017-08-29
Merge Lottie master into welshm

welshm:


welshm:
Created in the wrong direction again

PR #424, opened 2017-09-19
Fix backwards playback

blackm00n:
Right now backwards playback is broken. Code that is calculating how much progress to skip (by shifting `animation.beginTime` to past) when the current position is later in time than a starting one is not taking into account:
1) speed could be negative
2) `toEndFrame` could be less than `fromStartFrame`, e.g., when we call `animationView.play(fromProgress: 1, toProgress: 0, withCompletion: nil)`

Direct playback:
```
animationView.play()
```
Backwards playback:
```
animationView.animationProgress = 1
animationView.play(fromProgress: 1, toProgress: 0, withCompletion: nil)
```

|playback|before|after|
|-|-|-|
|direct|before_direct|after_direct|
|backwards|before_backwards|after_backwards|

### DIFF
--- a/lottie-ios/Classes/Private/LOTAnimationView.m
+++ b/lottie-ios/Classes/Private/LOTAnimationView.m
@@ -267,19 +267,34 @@ static NSString * const kCompContainerAnimationKey = @"play";
     _isAnimationPlaying = YES;
     return;
   }
+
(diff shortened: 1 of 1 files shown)


buba447:
This is great! Thank you for grabbing this. Its been on my todo for ages now!

Issue #459, opened 2017-10-22
Lottie build settings aren't marked as app extension safe


eskerber:
## Check these before submitting:
- [x] Updated to the latest (2.1.3) version of Lottie
- [x] The issue doesnt involve and [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#currently-unsupported-after-effects-features)
- [x] This issue isnt related to another open issue

## This issue is a
- [ ] Non-Crashing (Visual or otherwise) Bug
- [ ] Crashing Bug
- [x] Feature Request
- [ ] Regression (Something that once worked doesnt work anymore)

## What Platform are you on? 
- [ ] MacOS (Specify Platform Version)
- [x] iOS (Specify Platform Version)

## Expected Behavior 
Lottie, which does not use any extension-unsafe APIs (such as using `UIApplication` or notification APIs) does not trigger an unsafe dylib warning in targets that avoid unsafe APIs.

## Actual Behavior
Warnings

![2017-10-21 at 9 31 pm](https://user-images.githubusercontent.com/75236/31857576-75d93322-b6a7-11e7-8681-7eaba5a58125.png)

## Code Example

N/A
## Animation JSON

N/A

PR #460, opened 2017-10-22
Mark lottie as extension safe

eskerber:
This marks the Lottie framework as extension safe (it already is).

This prevents warnings in targets that require extension-safe APIs.

https://github.com/airbnb/lottie-ios/issues/459


buba447:
LGTM!

PR #479, opened 2017-11-09
add GLKit

rickytan:


buba447:
I removed the methods that use GLKKit and removed GLKKit all together from the project in a recent PR. They arent necessary. Thanks for grabbing this though!

PR #492, opened 2017-11-20
Stop animation before leaving screen to prevent freezing

mattlisiv:
I believe that when the view is being detached, if the current animation is not stopped, it will not properly re-enter the window. In order for the animation to continue when the a new window enters, it should be stopped.

I tested this with a few examples, and it functioned correctly when entering and re-entering the window from a detached view.

### DIFF
--- a/lottie-ios/Classes/Private/LOTAnimationView.m
+++ b/lottie-ios/Classes/Private/LOTAnimationView.m
@@ -562,7 +562,7 @@ static NSString * const kCompContainerAnimationKey = @"play";
   } else {
     // The view is being detached, capture information that need to be restored later.
     if (_isAnimationPlaying) {
-        [self stop];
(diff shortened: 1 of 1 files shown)


lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -562,6 +562,7 @@ - (void)willMoveToWindow:(UIWindow *)newWindow {
   } else {
     // The view is being detached, capture information that need to be restored later.
     if (_isAnimationPlaying) {
+        [self stop];
```
buba447: Should be pause and not stop. Stop rewinds the animation.

mattlisiv:
You are correct. I have updated it to pause.

buba447:
Great! Thanks!

PR #497, opened 2017-11-22
Check type to avoid app crash.

Brsoyan:


buba447:
Im not sure how this avoids a crash as the type should be enforced by the method declaration.

Brsoyan:
NSURL and NSString  subclass of NSObject and both confirm to protocols <NSCopying, NSMutableCopying, NSSecureCoding>.  When you set NSString by mistake, compile time we received only warning and we can build and run our app. After that app is crashed.

buba447:
I see. I would actually prefer an assert instead of just letting the function continue incorrectly. Either way crashing is the expected behavior here.

Brsoyan:
Ok I Agree.

PR #499, opened 2017-11-22
Implement initWithCoder: in LOTAnimatedControl

ansman:
Without it you cannot use LOTAnimatedControl in interface builder.


buba447:
Looks good to me! Thanks for grabbing this!

Issue #425, opened 2017-09-20
LOT_applyTransform crash


erema:
## Lottie-ios Issue
Hello! Sorry you are having an Issue!

## Check these before submitting:
- [ ] Updated to the latest (2.1.3) version of Lottie
- [ ] The issue doesnt involve and [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#currently-unsupported-after-effects-features)
- [ ] This issue isnt related to another open issue

## This issue is a
- [ ] Crashing Bug

Thread 6 Crashed:
0   libobjc.A.dylib                      0x0000000180d057f4 objc_object::release() + 8
1   Someapp                               0x000000010054563c -[LOTBezierPath LOT_applyTransform:] (LOTBezierPath.m:203)
2   Someapp                               0x0000000100553588 -[LOTRenderGroup performLocalUpdate] (LOTRenderGroup.m:151)
3   Someapp                               0x0000000100543174 -[LOTAnimatorNode updateWithFrame:withModifierBlock:forceLocalUpdate:] (LOTAnimatorNode.m:55)
4   Someapp                               0x0000000100553388 -[LOTRenderGroup updateWithFrame:withModifierBlock:forceLocalUpdate:] (LOTRenderGroup.m:137)
5   Someapp                               0x000000010055af10 -[LOTTrimPathNode updateWithFrame:withModifierBlock:forceLocalUpdate:] (LOTTrimPathNode.m:62)
6   Someapp                               0x0000000100553348 -[LOTRenderGroup updateWithFrame:withModifierBlock:forceLocalUpdate:] (LOTRenderGroup.m:135)
7   Someapp                               0x000000010054e5f8 -[LOTLayerContainer displayWithFrame:forceUpdate:] (LOTLayerContainer.m:238)
8   Someapp                               0x0000000100547a28 -[LOTCompositionContainer displayWithFrame:forceUpdate:] (LOTCompositionContainer.m:89)
9   Someapp                               0x000000010054e3e0 -[LOTLayerContainer display] (LOTLayerContainer.m:214)
10  QuartzCore                           0x000000018559ac20 CA::Layer::display_if_needed(CA::Transaction*) + 224
11  QuartzCore                           0x000000018559acb4 CA::Layer::layout_and_display_if_needed(CA::Transaction*) + 40
12  QuartzCore                           0x000000018551634c CA::Context::commit_transaction(CA::Transaction*) + 248
[...]

PR #506, opened 2017-11-29
fix crash by addding @synthesize to [LOTLayerContainer display]

wlei-fx:
In our app, we use Lottie to do tab switch animations. We found a few crash logs happen in [LOTLayerContainer display] but in a webthread. This “display callback in webthread” can be  reproduced after UIWebView used in app, and iOS will call display on webthread and main thread alternately. So we add a @synthesize protection at each LOTLayerContainer object level in [LOTLayerContainer display]. And the crash never happen again at large scale.

[LOT_applyTransform crash #425](https://github.com/airbnb/lottie-ios/issues/425)

### DIFF
--- a/lottie-ios/Classes/AnimatableLayers/LOTLayerContainer.m
+++ b/lottie-ios/Classes/AnimatableLayers/LOTLayerContainer.m
@@ -206,12 +206,14 @@
 }
 
 - (void)display {
-  LOTLayerContainer *presentation = self;
-  if (self.animationKeys.count &&
+  @synchronized(self) {
+    LOTLayerContainer *presentation = self;
+    if (self.animationKeys.count &&
       self.presentationLayer) {
-    presentation = (LOTLayerContainer *)self.presentationLayer;
+        presentation = (LOTLayerContainer *)self.presentationLayer;
+    }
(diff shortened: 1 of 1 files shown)


buba447:
thanks for grabbing this!

drkibitz:
@wlei-fx @buba447 
I can't help but think this fix seems a bit like a hammer, wouldn't using `[CATransaction lock];` and `[CATransaction unlock];` not be more appropriate for CALayer display?

drkibitz:
e.g. from this:
```objc
- (void)display {
  @synchronized(self) {
    LOTLayerContainer *presentation = self;
    if (self.animationKeys.count && self.presentationLayer) {
      presentation = (LOTLayerContainer *)self.presentationLayer;
    }
    [self displayWithFrame:presentation.currentFrame];
  }
}
``` 
to this:
```objc
- (void)display {
  [CATransaction lock];
  LOTLayerContainer *presentation = self;
  if (self.animationKeys.count &&
      self.presentationLayer) {
    presentation = (LOTLayerContainer *)self.presentationLayer;
  }
  [self displayWithFrame:presentation.currentFrame];
  [CATransaction unlock];
}

Issue #295, opened 2017-07-04
Can't play two animations in one UIView


machinefixer:
I have two animations load from different image bundle (with JSON file inside, of course), I want to put them in one UIView as a container view to play these animations one by one. But the weird thing is that one of the animations is broken, I finally find out that the animation initialized later would always broken while the first one plays well.  It seems the later animation initialized with the images of prior one, cause their image file name is same, but how could that happen? They are in totally different bundles. And if I initialize only one animation in the UIViewController, each animation works fine separately. Does lottie not support two animation in one UIView?  Here's the snapshot of how I load animations from custom image bundle. (The lottie-ios pod version is github repo branch master latest commit)

![2017-07-05 1 14 06](https://user-images.githubusercontent.com/5358642/27838803-727edb1a-611f-11e7-8e73-95384b8bf9e3.png)
![2017-07-05 1 07 59](https://user-images.githubusercontent.com/5358642/27838807-7aec5aa2-611f-11e7-8372-c3dd44ab5575.png)

machinefixer:
Finally! I find out workaround! Sounds stupid but it really works. I changed AE BodyMovin plugin's export settings to "use the original name of the image file when export images assets", before I export animation I just rename the image files' names with certain format like this (I have to admit that it is a time-consuming job :)), :
![2017-07-05 2 13 26](https://user-images.githubusercontent.com/5358642/27839912-8a8c9f32-6127-11e7-8694-601d5fa686f8.png)
 now each image in this animation have a unique name, and the two lottie animation view can be in same UIView eventually. But I still hope an official solution from Lottie team :)

therealsalih:
@machinefixer That was the reason why Hernan added the "use original name" feature back into bodymovin. To allow for multiple animations. You're doing the right thing  👍

zhangqifan:
It's solved my problems!  Using default names (like img_0 etc.) may cause unpredictable issues. Thanks @machinefixer👍

PR #511, opened 2017-12-06
[bugfix] fix collision of images with same name in different bundles.…

axoic:
images shared the same name from different bundles under the main bundle may not be always returned correctly when using `-[UIImage imageNamed:inBundle:compatibleWithTraitCollection:]`, especially when using lottie assets from more than one bundle in one viewController (which may due to Apple's cache mechanism).

Refers to issue: https://github.com/airbnb/lottie-ios/issues/295


buba447:
Thanks for grabbing this!

Issue #512, opened 2017-12-06
Using LOTAnimationView in Interface Builder


Xeaza:
I've been trying to figure out how to use a `LOTAnimationView` in Interface Builder so I can be consistent with how the rest of my app is set up. I'd also like to take advantage of auto layout in IB instead of having to hard code a `frame` for any `LOTAnimationView` I use. 

I'd like to propose implementing `initWithCoder:` on `LOTAnimationView` and a helper method to assign an animation to said view. 

I could be thinking about this incorrectly if so I'd love to hear how I'd use a `LOTAnimationView` in IB. Otherwise, would you be open to me submitting a PR? 
Thanks for reading! 

## Code Example
```objective-c
- (instancetype)initWithCoder:(NSCoder *)aDecoder {
  self = [super initWithCoder:aDecoder];
  if (self) {
    [self _commonInit];
  }
  return self;
}
```

```objective-c
- (void)setAnimationNamed:(nonnull NSString *)animationName {
  LOTComposition *comp = [LOTComposition animationNamed:animationName inBundle:[NSBundle mainBundle]];

  [self _initializeAnimationContainer];
  [self _setupWithSceneModel:comp];
}
```

## Use
* Add a view to a nib/storyboard set its class to `LOTAnimationView`
* Add an outlet to that view
* Call the helper method on your outlet. 
```objective-c
[self.myAnimationView setAnimationNamed:@"loading_animation"];
[self.myAnimationView play];
``` 

## Check these before submitting:
- [x] Updated to the latest version of Lottie (2.1.5)
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#currently-unsupported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Feature Request

## What Platform are you on? 

- [x] iOS

buba447:
This is great! Please do!

fadylateef:
Any solution !
Used latest pod , still found nil while using IB

luayzaemanuel:
Hi guys, any update with this? Still found nil while using IB.

Xeaza:
Hey @luayzaemanuel and @fadylateef, I'm not sure what the issues you're experiencing are. I just tried again and can confirm that if you follow the steps for use that I indicated above, as of the latest version (2.5.0) this still works. Best of luck!

prmacias:
[...]

PR #526, opened 2018-01-10
Add helper method to load animation on view loaded in Interface Builder

Xeaza:
Refers to issue #512


Xeaza:
I didn't update the example app to use this method because I couldn't think of a good place for it to go. If you think it's worth doing, let me know and I'll find a place.

buba447:
Great! Thanks for grabbing this! This will be in the pending 2.5 release.

rockylive:
Can you please release this fix to cocoa pods

fadylateef:
any new updates !

PR #527, opened 2018-01-10
Fix Analyser warning in LOTRadialGradientLayer

WDUK:
The static analyser (using Xcode 9.2) was complaining about `gradientComponents` being a variable length array with the possibility of having 0 length. This could occur with `numberOfComponents` keeping the value of 0 if no colours exist. Having a VLA with size 0 causes undefined behaviour.

The change is to bypass rendering the LOTRadialGradientLayer if there are no colours provided, as it doesn't make sense to render anything without a colour, and to presume a default colour may not be the right solution.


buba447:
Good catch! Thank you!

PR #528, opened 2018-01-10
set Public Headers

50percentgrey:
set LOTBlockCallback.h, LOTInterpolatorCallback.h and LOTValueCallback.h as public to avoid error on carthage.


buba447:
Ahh! Good catch! Somehow I missed that one. 
Ill loop back around and add them to the static library.

PR #533, opened 2018-01-14
Fix convenience class methods of public classes

floschliep:
Lots of the classes the framework exposes publicly offer convenience class methods for their instantiation. Unfortunately those class methods don’t create instances of their respective instancetype as declared, but create instances of hardcoded classes.

This means subclasses cannot take advantage of those convenience methods without introducing bugs. When using an instance that is expected to inherit from the subclass but does in fact inherit from the superclass, calling methods of the subclass will throw an exception. Furthermore, it’s not possible to add variables to those instances without causing a buffer overflow.

While it is possible to work around this using the Objective-C runtime, it would cause unnecessary overhead. As the return type of those class methods is declared as instancetype, it is expected that they return an instance of the current class, self.


buba447:
Nice catch! Never thought of this! Thanks!

PR #554, opened 2018-02-05
Fix spelling errors in lottie-ios

sujaypat:


buba447:
Thanks for grabbing this!

PR #560, opened 2018-02-18
Xcode warning fix

tikoyesayan:
Fix Xcode warning `block implicitly retains 'self'; explicitly mention 'self' to indicate this is intended behavior`


thecivillain:
Thanks for sorting this out, really annoying!

PR #589, opened 2018-03-19
Fix initializer chain

sanekgusev:
This is somewhat minor and easy to work around at the call site, but I stumbled upon this and thought I'd submit a PR.

Currently, `LOTAnimationView` does not override `UIView`/`NSView`'s  `initWithFrame:` designated initializer at all, but _does_ override `init`. This results in a correctly initialized instance when using `init`, and in a somewhat weird state (mainly due to `animationSpeed` remaining at `0`) when using `initWithFrame:`. Since `init` will call through to `initWithFrame:` anyway, it makes more sense to override `initWithFrame:` instead, which is exactly what I did in this PR.

`initWithContentsOfURL:` and `initWithModel:inBundle` are now also funneled through the overridden `initWithFrame:` to reduce code duplication to a small extent.


buba447:
Thanks for grabbing this!

PR #590, opened 2018-03-19
Rasterize _compContainer when not animating

sanekgusev:
This might be specific to my particular use case, but I've found that when displaying a lot of `LOTAnimationView` instances at the same time on a screen in a scroll view the scrolling frame rate is significantly reduced even though the animations are all stopped. This is because Lottie recreates the entire layer hierarchy of the original After Effects animation which can be arbitrarily complex. This results in a lot of strain to the device's GPU when compositing a lot of animation layers, which is especially noticeable during scrolling.

This PR introduces a new `shouldRasterizeWhenIdle` property defaulting to `YES` which sets `shouldRasterize` to `YES` on the underlying `_compContainer` layer whenever it is not animating. This results in the entire sublayer hierarchy being drawn once into a bitmap and then reused, effectively flattening the animation contents and reducing the GPU load to be comparable with rendering a regular image. There are also two new `didMoveToWindow`/`viewDidMoveToWindow` overrides for setting the `rasterizationScale` to match that of the current screen.

Rasterization is undesirable when actively resizing the `LOTAnimationView` instance, hence a separate property as a way to opt out of it, but I assume that the default value of `YES` should be more useful to most users of this library.

Since this behavior is pretty trivial to implement in the application code by either subclassing or wrapping a `LOTAnimationView` instance, I will be totally happy if this PR won't make it to the codebase.

### DIFF
--- a/lottie-ios/Classes/Private/LOTAnimationView.m
+++ b/lottie-ios/Classes/Private/LOTAnimationView.m
@@ -190,6 +190,9 @@ static NSString * const kCompContainerAnimationKey = @"play";
 - (void)_removeCurrentAnimationIfNecessary {
   _isAnimationPlaying = NO;
   [_compContainer removeAllAnimations];
+  if (_shouldRasterizeWhenIdle) {
+    _compContainer.shouldRasterize = YES;
+  }
 }
 
 - (CGFloat)_progressForFrame:(NSNumber *)frame {
@@ -340,6 +343,9 @@ static NSString * const kCompContainerAnimationKey = @"play";
       animation.beginTime = CACurrentMediaTime() - (offset * 1 / _animationSpeed);
     }
     [_compContainer addAnimation:animation forKey:kCompContainerAnimationKey];
+    if (_shouldRasterizeWhenIdle) {
+      _compContainer.shouldRasterize = NO;
+    }
   }
   _isAnimationPlaying = YES;
 }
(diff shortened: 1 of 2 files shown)


lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -190,6 +190,9 @@ - (void)_restoreState {
 - (void)_removeCurrentAnimationIfNecessary {
   _isAnimationPlaying = NO;
   [_compContainer removeAllAnimations];
+  if (_shouldRasterizeWhenIdle) {
+    _compContainer.shouldRasterize = YES;
```
buba447: If Im not mistaken this will override the ShouldRasterizeWhenIdlebehavior whenever the animation stops. Shouldnt this be:
`_compContainer.shouldRasterize = _shouldRasterizeWhenIdle`

sanekgusev: @buba447  Not sure I fully understand what you mean by override the ShouldRasterizeWhenIdle behavior…

The idea was to rasterize the layer at the time when animation stops if and only if `shouldRasterizeWhenIdle == YES`. If `shouldRasterizeWhenIdle == NO` then `LOTAnimationView` will not try to mess with `_compContainer`'s `shouldRasterize` at all (and will potentially let the client control mess with it freely).

sanekgusev: That said, I guess it might indeed be more consistent and easy to follow if we just did `_compContainer.shouldRasterize = _shouldRasterizeWhenIdle` here and removed the `if (_shouldRasterizeWhenIdle) {` check in `playFromFrame:` and just set `shouldRasterize = NO` there unconditionally.

buba447: Thats all i meant! Thanks!

sanekgusev:
@buba447 Here we go.

larssondaniel:
This did not improve performance in my case. I have `LOTAnimationViews` inside `UICollectionViewCells` and the only thing that worked for me was to rasterize entire cells while their lottie views weren't animating.

PR #622, opened 2018-05-04
Add basic support for base64 image asset URLs.

stristr:
This allows us to bake raster images directly into JSON without relying on a separate `images/` directory or similar—basically, instead of JSON that says:

```"u": "image/", "p": "filename.png"```

we can instead say:

```"u": "", "p": "data:image/png;base64,…"```

Note that (due to the simple way images are handled in the browser) this works already in `lottie-web`, and I'm planning to send a similar PR to `lottie-android`.

Regarding the approach, I noticed was a significant amount of refactoring of `LOTAsset` etc. required to avail ourselves of the image cache in a sane way when image "names" come through as giant strings, so I opted for something quick and easy. I'm definitely open to alternative approaches that might be more robust if this approach seems problematic in terms of performance.


buba447:
LGTM

PR #623, opened 2018-05-04
Calling boolValue instead of converting NSNumber

Garolla:
Problem found with static analysis code of Xcode 9.3.
"Converting a pointer value of type 'NSNumber *' to a primitive boolean value; instead, either compare the pointer to nil or call -boolValue"


buba447:
A nil check should be performed here, not boolValue

PR #630, opened 2018-05-22
Add ability to set a new JSON on an existing animation

dimitryz-luxuryretreats:
I've added a new interface method to the LOTAnimationView that allows changing the animation through a JSON string, not just a name (as the current implementation allows).

```
- (void)setAnimationFromJSON:(nonnull NSDictionary *)animationJSON NS_SWIFT_NAME(setAnimation(json:));
```


lauraskelton:
cc @buba447

PR #633, opened 2018-05-24
LOTImageCache: Rename methods in order to avoid Non-public API calls.

Shachlan:
According to Apple's policies, apps cannot use the non-public
selectors imageForKey: and setImage:forKey:. Thus, the methods
signature must change. +


FrankSwift:
I had met some problem, how to solve?

Shachlan:
@FrankSwift, it seems to be some statistical issue - we managed to solve the issue without changing Lottie. Apparently you may have some illegal method names, but not too many.
Try using my change to see if it solves the issue, and if not, go through your code and dependencies and find all illegally named methods.

CrazyFanFan:
> @FrankSwift, it seems to be some statistical issue - we managed to solve the issue without changing Lottie. Apparently you may have some illegal method names, but not too many.
> Try using my change to see if it solves the issue, and if not, go through your code and dependencies and find all illegally named methods.

I have encountered the same problem and solved it in the same way.

PR #636, opened 2018-05-25
Fixing issue: macOS doesn't handle window changes as iOS does. 

daniele-pizziconi:
If the LOTAnimationView has been added to a container view just after the play method was called, then the animation won't start.
A recovery mechanism is in place for iOS, but not for macOS, hence my fix to apply the same iOS logic of handling window changes to macOS.


buba447:
LGTM

PR #670, opened 2018-07-11
Fix Xcode warning when using an NSNumber as a boolean condition.

Coeur:


Coeur:
@buba447, any update regarding reviewing pending pull requests? :)

Coeur:
@buba447, please, iOS 12 is being released today, and we're still waiting for feedback on pending pull requests. Would you like some help in maintaining the project?

PR #681, opened 2018-07-26
Added fix for loading images from asset catalogue

ttanatb:
This fix is built from https://github.com/airbnb/lottie-ios/pull/596 but the check is moved down to catch more cases. In the past it will only look for the image from the asset catalogue only if the root directory was never specified (if users initialized the animation with `[LOTAnimationView animationWithFilePath]`, the root directory will always be filled out). 

This allows us to use the asset catalogue for multiple cases (which is great because the asset catalogue resizes the images for us !!)


ttanatb:
@buba447 @revolter

PR #757, opened 2019-01-09
Fix hidden not Precomp or not Solid frames if _assetGroup is missing

nekrich:
Minor fix of #729  
Static analyzer is still happy


nekrich:
Closing, cause you just reverted the breaking commit.

PR #761, opened 2019-01-17
Support image in the disk for MacOS

loongliu:
Currently, lottie for MacOS only support image in the Application Assets.
This commit add local image support


loongliu:
@buba447 Would you mind speed some time reviewing this commit?

buba447:
Looks good!

PR #772, opened 2019-02-12
Early return nil if animationName is nil to prevent a crashing issue

chuganzy:
Since `animation` [here](https://github.com/airbnb/lottie-ios/blob/master/lottie-ios/Classes/PublicHeaders/LOTAnimationView.h#L41) is marked as nullable, it can be technically nil, which cases a crash in LRU cache.


buba447:
Looks Good!

PR #782, opened 2019-03-13
Add LOTValueCallback.m to LottieLibraryIOS target

dflems:
The `LottieLibraryIOS` static library target was missing `LOTValueCallback.m` from its Compile Sources phase. It's in every other framework and static library target, so this just seems like an oversight.


Coeur:
Click "Edit" and change the base to be "airbnb:lottie/objectiveC", thank you.

Coeur:
I think this project and its targets may be used by Carthage only.
Thanks for the LottieLibraryIOS target fix.

You could also fix the **Lottie_tvOS** target, as it's missing:
- LOTBlockCallback.m
- LOTInterpolatorCallback.m
- LOTValueCallback.m

buba447:
Nice catch! That was definitely an oversight. Please change the merge branch to Objective-C and Ill merge it. 
Thanks!

dflems:
@buba447 updated the base branch

Coeur:
@dflems so, no love for tvOS? ;)

PR #784, opened 2019-03-14
Fix warnings in Xcode 10.2

thedrick:
This fixes warnings in Xcode 10.2 related to access modifiers

@buba447 @Coeur


thedrick:
@Coeur the build was failing with an error that the cache variable wasn’t accessible due to an internal declaration. I had to make it public for the project to compile. Maybe I’m missing something obvious that would have fixed that?

Coeur:
Oh, my bad, you're right, I had not test this latest commit: so it got broken by buba a few hours earlier. Thank you for fixing it.

Issue #789, opened 2019-03-18
Crash on LOTCircleAnimator.m 


kaannesinn:
Hi, 

First of all, thanks for a great library that you create for using to show animative loading effects with some json files.

I am using Lottie-iOS library 2.5.3 in my ios app (it is our application's second version). We dont have a crash on the 1st version but now there is a minor crash comes from "LOTCircleAnimator.m" with the line 37 and you can see the details below. 

Our application details are like that;

- Using Xcode 10.1
- Min deployment target 10.0
- Using Swift 4.2

## This issue is a:
- [x] Crashing Bug
But we cant reproduce it with debug versions of the application. 

## Which Version of Lottie are you using?

Lottie 2.5.3

## What Platform are you on? 
- [X] iOS

## What Language are you in?
- [X] Swift

## Expected Behavior 
Cant reproduce
Crash I think comes from the lottie animation we use on splash screen. 

## Actual Behavior
Cant reproduce

## Code Example
```
import Lottie
    @IBOutlet weak var animationView: LOTAnimationView!

in awakeFromNib method on our custom LoadingView view-class;

animationView.isHidden = false
        animationView.animation = "loading"
        animationView.loopAnimation = true
        animationView.play()
```

## Animation JSON
[twoAnimationFilesWeUse.zip](https://github.com/airbnb/lottie-ios/files/2977154/twoAnimationFilesWeUse.zip)

Is there a bug on this Lottie's version or is there something we miss while using the SDK/json?

Thanks, regards
Kaan Esin

kaannesinn:
Crash logs are like that;
`Crashed: com.apple.main-thread
0  libobjc.A.dylib                0x18100c910 objc_msgSend + 16
1  Lottie                         0x10208e784 -[LOTCircleAnimator needsUpdateForFrame:] (LOTCircleAnimator.m:37)
2  Lottie                         0x10208b0f0 -[LOTAnimatorNode updateWithFrame:withModifierBlock:forceLocalUpdate:] (LOTAnimatorNode.m:44)
3  Lottie                         0x10208b118 -[LOTAnimatorNode updateWithFrame:withModifierBlock:forceLocalUpdate:] (LOTAnimatorNode.m:48)
4  Lottie                         0x10209e094 -[LOTRenderGroup updateWithFrame:withModifierBlock:forceLocalUpdate:] (LOTRenderGroup.m:142)
[...]

PR #796, opened 2019-03-18
Fix for storyboard crash

buba447:
Fixes a storyboard agent crash when initializing an AnimationView from storyboard.

### DIFF
--- a/Example/lottie-swift/Base.lproj/Main.storyboard
+++ b/Example/lottie-swift/Base.lproj/Main.storyboard
@@ -12,7 +12,7 @@
         <!--View Controller-->
         <scene sceneID="ufC-wZ-h7g">
             <objects>
-                <viewController id="vXZ-lx-hvc" customClass="ViewController" customModule="lottie_swift_Example" customModuleProvider="target" sceneMemberID="viewController">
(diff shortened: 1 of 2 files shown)


kaannesinn:
@buba447 Is this change on the Lottie's 3.0.1 pod version? Because when I update to 3.0.1 and want to use AnimationView class on our storyboard as an outlet, app crashes. (saying AnimationView class from Interface builder isnt known)

buba447:
@kaannesinn This is in 3.0.1. Did you set the module for the AnimationView to Lottie?

buba447:
@kaannesinn I found another crash in the Interface Builder and put a fix up here https://github.com/airbnb/lottie-ios/pull/808
Hopefully that fixes the issues. These interface builder crashes are hard to debug since there is no crash logs. Test it out and please open a new issue if you continue experiencing a crash so i can look into it!

kaannesinn:
@buba447 ok I have updated to the master branch with the issue #789 and now set the module to Lottie. Now crash disappeared while try to show the animation view. Thanks :)

I am using the last commit with your new fix #808 . I will track it again and tell you if I see a new crash with that, ok.

PR #800, opened 2019-03-19
Fixes issue in AnimatedSwitch when setting the same state

thedrick:
There is an issue in AnimatedSwitch right now where if you set isOn to the same value multiple times it doesn't seem to update properly. This fixes that issue.

@buba447


buba447:
Nice catch!

PR #802, opened 2019-03-20
Added base64 image support

buba447:
Adds support for base 64 encoded images.


gdoublev:
I think you accidentally killed the changes from #801 with this PR?

buba447:
@gdoublev I dont think so. I intentionally commented out the switch from the `Example` App. Otherwise nothing from the actual Lottie library should have been changed.

gdoublev:
Feeling pretty sheepish right now, didn't even pay attention that it was just an example of its usage. Sorry! 😨

buba447:
@gdoublev No worries! Im just happy someone out there is double checking my work! ❤️

Coeur:
That reminds me that I used `typedef` in objective-c (`typealias` in swift) to factor most iOS and macOS code together, as seen [here](https://github.com/laptobbe/TSMarkdownParser/blob/3.x/TSMarkdownParser/TSFoundation.h):

````
#if TARGET_OS_IPHONE
// iOS, watchOS, tvOS
@import UIKit;
#else
// Compatibility with macOS
@import AppKit;
typedef NSColor UIColor;
typedef NSImage UIImage;
typedef NSFont UIFont;
typedef NSFontDescriptor UIFontDescriptor;
typedef NSFontSymbolicTraits UIFontDescriptorSymbolicTraits;
#endif
````

Coeur:
Oh, too bad it wasn't included in 3.0.1 release. :(
I guess we'll have to wait for next release then.

PR #806, opened 2019-03-21
Support removing animation by setting animationName to nil

chuganzy:
Title says it all


buba447:
Looks good. Thanks for grabbing this.

PR #831, opened 2019-03-27
Fix macOS & tvOS builds

nekrich:
Hi.

Found and fixed some build issues:

1. Removed codesigning
2. `TextCompositionLayer.swift` was duplicated in compile sources twice in tvOS and macOS targets


buba447:
Thanks for grabbing this!

Issue #853, opened 2019-04-09
Impact generator generates impact on view creation for AnimatedSwitch


thedrick:
## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [] Feature Request
- [] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.0.3

## What Platform are you on? 

- [] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [] Objective-C

## Expected Behavior 
- the impact generator should not generate an impact when an AnimatedSwitch view is initialized

## Actual Behavior
- the impact generator does generate an impact when an AnimatedSwitch view is initialized

## Code Example
```swift
let animatedSwitch = AnimatedSwitch(animation: anyAnimation)
// impact occurs here
```
Stack trace:
![Screen Shot 2019-04-08 at 10 27 33 AM](https://user-images.githubusercontent.com/796488/55764992-88155c80-5a22-11e9-97a2-4ebec907e476.png)

Issue #837, opened 2019-03-30
Cannot programmatically set AnimatedButton on/off state


Speedrockracer:
Hi,
I just upgraded our app from Lottie 2.5.2 to 3.0.3 and I ran into an Issue with the new AnimatedSwitch.
Previously I would use `button.setOn(true, animated: true)` to programmatically set the on/off state and it would animate nicely to that state.
However the new AnimatedSwitch does not have this method I can only do `button.isOn = true` and it wil jump to that state.
I checked the code and internally it uses the function `updateOnState(isOn: Bool, animated: Bool)`
But this function is internal and swift wont allow me to call it from my app.
Can we make this function public so we can animate the on/off state? I can make a pr for this if you guys are busy.

## This issue is a:
- [] Non-Crashing Bug (Visual or otherwise)
- [] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?
Lottie 3.0

## What Platform are you on? 
- [] iOS

## What Language are you in?
- [] Swift

## Expected Behavior 
Being able to programmatically set the button on/off state animated.

## Actual Behavior
Can only set the button on/off state without animation.

buba447:
Added here https://github.com/airbnb/lottie-ios/pull/856

PR #856, opened 2019-04-09
Allow for programmatic animations in AnimatedSwitch and fix haptics issue

thedrick:
This fixes https://github.com/airbnb/lottie-ios/issues/853 and https://github.com/airbnb/lottie-ios/issues/837

I made 2 commits - the first is the changes I made and the second is just running `pod install` in the `Example` directory. Let me know if I should revert that commit. 

@buba447


buba447:
Looks great! Thanks for grabbing this.

PR #857, opened 2019-04-10
Allow AnimatedControl to set a value provider

thephamiliar:
In order to [change the colors at runtime](https://github.com/airbnb/lottie-ios#now-lets-change-their-colors) for `AnimatedControl`, we need to expose a method so that the private `animatedView` can set its value provider.

Main changes are in: https://github.com/airbnb/lottie-ios/commit/d5f1ae03b0d3f800d2b0aa1029e9cafd12aa8044

cc: @thedrick

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -6,11 +6,11 @@ DEPENDENCIES:
 
 EXTERNAL SOURCES:
   lottie-ios:
-    :path: ../
+    :path: "../"
 
 SPEC CHECKSUMS:
-  lottie-ios: 3e7a6340cfb224366afe213a50fd1f4056df3026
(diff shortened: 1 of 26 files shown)


thedrick:
Lets merge this through and we can make updates in the future if we feel this isn't sufficient. thanks for adding this @thephamiliar

DennisHirschgaenger:
Why is the `animationView` not exposed from `AnimatedControl`? I need access to the `textProvider` from an `AnimatedSwitch` (with a label indicating the current state "on"/"off").

As there are numerous other cases where properties of the `AnimationView` need to be accessed from `AnimatedControl`s, the obvious solution would be to directly expose it. Is there a reason not to do so?

PR #860, opened 2019-04-11
bump version to 3.0.5

thedrick:
Bumps version to 3.0.5. I'm having trouble publishing a new tag, so I may need someone else to handle that for me.

@buba447 @thephamiliar

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - lottie-ios (3.0.4)
+  - lottie-ios (3.0.5)
 
 DEPENDENCIES:
   - lottie-ios (from `../`)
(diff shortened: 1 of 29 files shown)


Example/Podfile.lock
```diff
@@ -9,8 +9,8 @@ EXTERNAL SOURCES:
     :path: "../"
 
 SPEC CHECKSUMS:
-  lottie-ios: 55a2de40418f8de2b155efaba4d4d1a025a5c570
+  lottie-ios: 364ab2a5168e95ca4f0775b967fded860428ea04
 
 PODFILE CHECKSUM: d7d1d3df3df6c9862ab67dbbfb2f695895dc9d7b
 
-COCOAPODS: 1.6.1
+COCOAPODS: 1.5.3
```
thedrick: we should probably have a Gemfile in here to prevent these versions from changing so much

buba447: It's really no big deal since its just the example project. @Coeur

Issue #850, opened 2019-04-05
Animation is broken after removing and adding the View


mariohahn:
## Check these before submitting:
- [X] This issue isn't related to another open issue

## This issue is a:
- [X] Non-Crashing Bug (Visual or otherwise)
- [X] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.0.3

## What Platform are you on? 
- [X] iOS

## What Language are you in?
- [X] Swift

## Expected Behavior 
play normally 

## Actual Behaviour
If you remove the superview and add it again the animation is played automatically. if you try to play it after adding it again it will not play anymore. 

I guess the reason is because the animation context is not nil after playing the animation. After adding the view again animationMovedToWindow() is called and is playing the animation automatically.

## Code Example

```swift
class Test: UIView {
    
    class FastSeekOverlay: UIView {
        
        let fastForward: AnimationView = {
            $0.translatesAutoresizingMaskIntoConstraints = false
            
            return $0
        }(AnimationView(name: "fastForward"))
        
        
        override init(frame: CGRect) {
            super.init(frame: frame)
            
            addSubview(fastForward)
            
            NSLayoutConstraint.activate([
                fastForward.centerYAnchor.constraint(equalTo: centerYAnchor),
                fastForward.trailingAnchor.constraint(equalTo: readableContentGuide.trailingAnchor, constant: -20),
                fastForward.widthAnchor.constraint(equalToConstant: 73),
                fastForward.heightAnchor.constraint(equalToConstant: 20)
            ])
        }
        
        let testView = FastSeekOverlay()
        
        override init(frame: CGRect) {
            super.init(frame: frame)

            testView. fastForward.play()
            addSubview(testView)
        }
        
        func removeAndAddAgain() {
            testView.removeFromSuperview()
            addSubview(testView)
        }
    }
}
```

## Animation JSON
```
[...]

PR #864, opened 2019-04-15
Fixed Bug

mariohahn:


mariohahn:
https://github.com/airbnb/lottie-ios/issues/850

Coeur:
I agree with buba447, as it is, this PR can't be approved: a solution compatible with background-behavior should be found. Also, the title of the PR should be made explicit.

buba447:
This should be fixed properly over here https://github.com/airbnb/lottie-ios/pull/926

PR #867, opened 2019-04-16
Fix duplicate reference to TextCompositionLayer.swift in project

Coeur:
TextCompositionLayer.swift was included twice in the project, for overlapping targets.


Coeur:
@buba447 this is an easy PR. You can eventually see and fix the issue yourself:
Screen Shot 2019-04-23 at 13 45 57

PR #870, opened 2019-04-16
Enhancement/swift5migration

amrangry:
Enable static analyzer check for missing localizability
+
Swift 5 support

### DIFF
--- a/Lottie.xcodeproj/project.pbxproj
+++ b/Lottie.xcodeproj/project.pbxproj
@@ -1450,6 +1450,7 @@
 				TargetAttributes = {
 					486E83B1220A317C007CD915 = {
 						CreatedOnToolsVersion = 10.0;
+						LastSwiftMigration = 1020;
 					};
 					486E84E1220A357D007CD915 = {
 						CreatedOnToolsVersion = 10.0;
@@ -2302,7 +2303,7 @@
 				PRODUCT_NAME = Lottie;
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				SKIP_INSTALL = YES;
-				SWIFT_VERSION = 4.2;
+				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Debug;
@@ -2329,7 +2330,7 @@
 				PRODUCT_NAME = Lottie;
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				SKIP_INSTALL = YES;
-				SWIFT_VERSION = 4.2;
+				SWIFT_VERSION = 5.0;
 				TARGETED_DEVICE_FAMILY = "1,2";
 			};
 			name = Release;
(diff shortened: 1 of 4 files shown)


jonas-sleepcycle:
Love to see this merged!

appfrilans:
Love to see this merged!

salmaanahmed:
This should be merged soon, waiting for it

iDevid:
Waitin' for this :)

dfed:
We should also update [lottie-ios.podspec](https://github.com/airbnb/lottie-ios/blob/master/lottie-ios.podspec#L25) to use 5.0 as the `swift_version`

thedrick:
Thanks for this PR @amrangry - I'm working on this now. Could you make the change @dfed suggested before we merge this in? I'm going to do a bump of Lottie that only includes this change

Coeur:
@thedrick @dfed changing the minimum swift version is out of scope here: this PR is to fix the localizability on the EXAMPLE project (by adding `Base,`) and could be merged as is.

Changing the minimum swift version in the podspec is not strictly needed and should be done in a separate PR.

dfed:
@Coeur maybe I'm missing something. I see that the `SWIFT_VERSION` [is changed on Lottie.xcodeproj](https://github.com/airbnb/lottie-ios/pull/870/files#diff-3145c4e4052f34ea1c3a594af502d418R2303) in this PR. It's good practice to keep the Xcode project's Swift version and the Cocoapods Swift version in sync. I don't actually see any changes to the [Example](https://github.com/airbnb/lottie-ios/tree/master/Example) directory in this PR.

> Changing the minimum swift version in the podspec is not strictly needed and should be done in a separate PR. It would also be a breaking change for users on Xcode 10.1 that have other dependencies on Swift 3 for instance.

You're right that adding `Base,` could be done in a different PR from updating to Swift 5 on the Xcode project. But that isn't what was done as far as I can tell.

Regarding breaking changes: I'd been thinking that most consumers of Swift Cocoapods are manually setting their `swift_version` when integrating pods to avoid breakages these days. But you're right that even setting the `SWIFT_VERSION` in `Lottie.xcodeproj/project.pbxproj` could be a breaking change for someone integrating Lottie via submodules. We should probably revert that part of the change then.

> Also, if we just wait a bit for the release of CocoaPods 1.7.0 (coming soon, see https://github.com/cocoapods/cocoapods/releases), then we'll be able to specify multiple supported swift versions like [4.2, 5.0] instead of just the minimum one.

I'm excited for this! But in the meantime, we should keep our Xcode project `SWIFT_VERSION` and Cocoapods `swift_version` in sync

Coeur:
Oh, you're right, the change is affecting Carthage builds and/or submodules integrations.

Well, maybe let's postpone this change until the release of CocoaPods 1.7.0 (which, I believe, will be during the WWDC week, so that they can validate it with Xcode 11 beta)

vg-identance:
@Coeur Cocoapods 1.7.0 just released https://github.com/CocoaPods/CocoaPods/releases/tag/1.7.0

Coeur:
@vg-identance yeah, with a regression, I know: #8837

buba447:
@Coeur We will add the 'multiple swift versions' in a different pull request when cocoapods gets sorted out. Im going to go ahead and merge this PR and ship Swift 5 under 3.1.0

PR #880, opened 2019-04-22
macOS. Fix `'LottieContentMode' has no member 'rawValue'` compile error

nekrich:
Hi.

#868 brought compile error to macOS target, cause `LottieContentMode` is not `RawRepresentable`.
Error: `Value of type 'LottieContentMode' has no member 'rawValue'`. https://github.com/airbnb/lottie-ios/blob/84009f19580f9ad770a3b706e1ddfefc8a7a87e6/lottie-swift/src/Public/Animation/AnimationView.swift#L612

There is no need in `@unknown default`  check of `LottieContentMode` on macOS target.


Coeur:
Oh, I learnt something today. Thank you for bringing this, @nekrich, and sorry for breaking the macOS target.

PR #881, opened 2019-04-22
Lottie content mode consistency

nekrich:
Hi 🤗.

This PR will bring equality between `LottieContentMode` on macOS platfrom and `UIView.ContentMode` on iOS platfrom.


Coeur:
Good. And if you add `@objc`, we don't need the other PR. :D
That was my joke, please **don't** add `@objc`. ;)

PR #900, opened 2019-05-06
Allow to use lottie framework inside extensions

lonsky:
Without this flag set, you can see the warning during compiling process:
```
linking against dylib not safe for use in application extensions
```
details: https://developer.apple.com/library/archive/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html

This regression was introduced in the new 3.x "pure-swift" version.
In previous (2.x) version this project setting was set.


yutae:
This PR seems to be duplicated. #893 🤭

thedrick:
This has been merged in #893

PR #906, opened 2019-05-13
Bump to 3.0.6

thedrick:
This bumps Lottie to 3.0.6 and also reformats the podspec file using a Ruby linter. Let me know I should revert that change

@buba447 @Coeur

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - lottie-ios (3.0.5)
+  - lottie-ios (3.0.6)
 
 DEPENDENCIES:
   - lottie-ios (from `../`)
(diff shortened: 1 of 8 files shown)


thedrick:
I verified that this builds correctly for iOS, MacOS, and TvOS using CocoaPods and Carthage

Coeur:
Hello,
Yes, it builds and could be merged, but note that:
- I'm not a fan of the `%w` syntax, as it means our eyes have to read multiple syntaxes for the same thing (%w when array elements don't have any space, and classic single quotes for other arrays)
- The commit also introduces a mix of syntaxes between s.license and s.author: one uses `symbol: value` and the other uses `key => value`.
 Sure, overall we use less characters, but perhaps we shouldn't make it harder for Ruby neophytes.

thedrick:
@Coeur sounds good to me, I'll revert those changes

thedrick:
@Coeur updated!

PR #910, opened 2019-05-14
Enhancement: Modified to allow for defined repeat counts

daemedeor:
A patch to allow for manually setting whether it infinite loops or a specific amount (as well as backwards as well). Didn't want to include reading the animation defined repeat count as well since that's a bigger change and I need to test longer (also not as sure on how to read the right animation count yet).


daemedeor:
I didn't know how to read through the JSON to provide support based on the animation object being read in (or if there were multiple animations defined on different objects, what then?). However, I figured the quicker need is to at least be able to define the looping behavior somewhere, if not on the animation itself, at least manually. I get the sense that it's likely to need to change: action(forKey:) in AnimationContainer but didn't want to make a huge change since I don't have any repeating animations available offhand and a project to use this with.

Issue #849, opened 2019-04-03
Make all the Codable models public


joekim:
## Check these before submitting:
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Feature Request

For projects I'm working on we are using more and more information in Lottie files to construct user interfaces. A lot of useful information is decoded by the `Animation` model but most of it is not public. For example, `markers` are on the `Animation` object but they cannot be iterated over for use with the `func progressTime(forMarker named: String) -> AnimationProgressTime?` method. What if we made all the Codable models and properties public? I'd be happy to make a PR for it.

## Which Version of Lottie are you using?

Lottie 3.0

## What Platform are you on? 

- [x] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift

## Code Example

```    
   func printAnimations(values: TemplateValues) {
        let markerNames =  animationView.animation!.markerNames!.sorted()
        for markerName in markerNames{
            print("\(markerName): \(animationView.progressTime(forMarker: markerName)!)")
        }
    }
````

Coeur:
Then, please make a PR. :)

calda:
I'd prefer to keep these `internal`, to avoid growing the public API surface of the library (e.g. changing these types would become source-breaking)

PR #918, opened 2019-05-21
Add markerNames property

dpt:
This adds a new property `markerNames` to Animation which returns an array of all marker names.

There's a couple of C&P fixes in there too.

### DIFF
--- a/lottie-swift/src/Private/Model/Animation.swift
+++ b/lottie-swift/src/Private/Model/Animation.swift
@@ -57,6 +57,13 @@ public class Animation: Codable {
   let markers: [Marker]?
   let markerMap: [String : Marker]?
   
+  /// Return all marker names, in order, or an empty list if none are specified
+  public var allMarkerNames: [String]
+  {
+    guard let markers = markers else { return [] }
+    return markers.map{$0.name}
+  }
+  
   enum CodingKeys : String, CodingKey {
     case version = "v"
     case type = "ddd"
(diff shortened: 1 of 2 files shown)


dpt:
#849 is related.

lottie-swift/src/Private/Model/Animation.swift
```diff
@@ -57,6 +57,13 @@ public class Animation: Codable {
   let markers: [Marker]?
   let markerMap: [String : Marker]?
   
+  /// Return all marker names, in order, or an empty list if none are specified
+  public var allMarkerNames: [String]
```
buba447: What do you think about renaming this to just `markerNames`?

dpt: Sure. Updated.

lottie-swift/src/Private/Model/Animation.swift
```diff
@@ -57,6 +57,13 @@ public class Animation: Codable {
   let markers: [Marker]?
   let markerMap: [String : Marker]?
   
+  /// Return all marker names, in order, or an empty list if none are specified
+  public var markerNames: [String]
```
thedrick:
```suggestion
  public var markerNames: [String] {
```

lottie-swift/src/Private/Model/Animation.swift
```diff
@@ -57,6 +57,13 @@ public class Animation: Codable {
...
+  {
```
thedrick:
```suggestion
```

lottie-swift/src/Private/Model/Animation.swift
```diff
@@ -57,6 +57,13 @@ public class Animation: Codable {
...
+    guard let markers = markers else { return [] }
+    return markers.map{$0.name}
```
thedrick:
```suggestion
    return markers.map { $0.name }
```

dpt:
Are any other changes needed?

Issue #924, opened 2019-05-29
Lottie Scaling Problem


wirnico:
## Check these before submitting:
- [] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [] Feature Request
- [x] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.0

## What Platform are you on? 

- [] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [] Objective-C

## Expected Behavior 
I have a view which can be dragged so that the header is getting shrinked. In the header I am using an AnimationView with Lottie 3.0

## Actual Behavior
Once I was on Lottie 2.5 the LottieView was shrinking immediately, but now since the update on 3.0 it lacks.
Is it a setting error I have made or is it a bug in Lottie 3.0?

## Code Example
I create a view in the Storyboard and give it the class AnimationView. Module is Lottie.
In the Swift File I wrote just this:
animationView.animation = Animation.named("analysis_saving")

## Animation JSON

I attached the Lottie File and also a scrrencapture
[Lottie Scaling Problem.zip](https://github.com/airbnb/lottie-ios/files/3232113/Lottie.Scaling.Problem.zip)

buba447:
How are you changing the scale of the view?

wirnico:
In the video I attached you can see a container in the header where the AnimationView is implemented. It has constraints to the top and bottom and no height or width. If the user swipes up the container gets compressed and with it the AnimationView.

gregligierko:
// Perhaps related... I posted this a few days ago on StackOverflow. //

Title: 'How can I disable the content scale animation, when changing frame size?'

I am using Lottie 3.0 (Swift 4.2) mixed in an Objective-C project. I control the progress "manually":

       animationView.currentFrame = p

When I change the AnimationView instance's .frame (CGRect) in order to resize the view, then the the view moves instantly to the new position and size, but the contents/assets scale over time (maybe 0.2-0.5s) from the old to the final size. This scaling animation seems to have a linear progress.

I tried both with UIView.performWithoutAnimation and without it (I can see actions are disabled in Lottie's code anyway):

        UIView.performWithoutAnimation {
            animationView.frame = rect
        }
[...]

PR #932, opened 2019-05-31
Fix for Implicit Animations during layout

buba447:
When changing the frame of an AnimationView in an animation block the view would not respect the duration of timing curve of the animation and fell back on 0.25 / linear animation. 

This is because UIView animations use a slightly different system than the CALayer implicit animations and do not set the CATransaction's duration or timing curve.

I added a fix that pulls the current animations duration and timing curve and properly sets the CATransaction for layout.

Before:
implicitAnimations_Broke

After:
implicitAnimations_Fixed

Mentioned : https://github.com/airbnb/lottie-ios/issues/924

### DIFF
--- a/lottie-swift/src/Public/Animation/AnimationView.swift
+++ b/lottie-swift/src/Public/Animation/AnimationView.swift
@@ -33,19 +33,19 @@ public enum LottieLoopMode {
 }
 
 extension LottieLoopMode: Equatable {
-    public static func == (lhs: LottieLoopMode, rhs: LottieLoopMode) -> Bool {
-        switch (lhs, rhs) {
-        case (.repeat(let lhsAmount), .repeat(let rhsAmount)),
-             (.repeatBackwards(let lhsAmount), .repeatBackwards(let rhsAmount)):
-            return lhsAmount == rhsAmount
-        case (.playOnce, .playOnce),
-             (.loop, .loop),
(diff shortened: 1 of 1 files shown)


lottie-swift/src/Public/Animation/AnimationView.swift
```diff
@@ -33,19 +33,19 @@ public enum LottieLoopMode {
...
+    switch (lhs, rhs) {
+    case (.repeat(let lhsAmount), .repeat(let rhsAmount)),
+         (.repeatBackwards(let lhsAmount), .repeatBackwards(let rhsAmount)):
+      return lhsAmount == rhsAmount
+    case (.playOnce, .playOnce),
+         (.loop, .loop),
+         (.autoReverse, .autoReverse):
+      return true
+    default:
+      return false
```
buba447: This is just an indent fix.

cipolatica:
Hello, it seems the fix for this issue is no longer in version 3.1.4. Was removed with commit 43ac4b72ef53e6ff9b6a918b8b6a15d78346d178. why was this removed? can it be readded? thanks

Pash237:
It still doesn't work when using spring damping UIView animations :(
`animateWithDuration:delay:usingSpringWithDamping:initialSpringVelocity:options:animations:completion:`

PR #937, opened 2019-06-06
swift package.swift to work at least with iOS

StatusQuo:


alextudge:
Hey! Any reason the package defines a minimum deployment target of ios12? It works with 10 when installed via Cocoapods!

PR #946, opened 2019-06-14
Improve compatibility with Swift Package Manager for iOS

daveanderson:
Does not resolve the inability to do a Swift Package Manager build for macOS but is a step in the right direction.

### DIFF
--- a/lottie-swift/src/Public/MacOS/AnimationSubview.swift
+++ b/lottie-swift/src/Public/MacOS/AnimationSubview.swift
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+#if os(OSX)
 import AppKit
 
 /// A view that can be added to a keypath of an AnimationView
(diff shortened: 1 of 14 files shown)


daveanderson:
Similar to #939 but does not change the locations of files

buba447:
Hey @daveanderson sorry Im just now getting around to looking at this! First, thanks for your help contributing to Lottie!

So whats the right path forward with this. Im not familiar with Swift Package Manager, is there no way to occlude files from the build based on platform?

StatusQuo:
Hi @buba447 no actually if os annotation is only available in source files.

donnywdavis:
If you could separate the files into different targets then you could exclude based on platform. Then you would just have to do the file level `#if` annotations for shared code. Here's an example of how `Quick` does that, https://github.com/Quick/Quick/blob/master/Package.swift. Sorry if not relevant, just saw the example and and wanted to mention it. 🙂

Issue #905, opened 2019-05-13
Vertical centering text issue


NavjotPanesar:
Hey guys, 

I seem to be running into vertical centering issues on lottie-ios. The below animation looks find on web and android, I was wondering if someone could lend a hand in making this work :)

## Check these before submitting:
- [X] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [X] This issue isn't related to another open issue

## This issue is a:
- [X] Non-Crashing Bug (Visual or otherwise)
- [ ] Crashing Bug
- [ ] Feature Request
- [ ] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.0

## What Platform are you on? 

- [ ] MacOS
- [X] iOS

## What Language are you in?
- [X] Swift
- [ ] Objective-C

## Expected Behavior 
![Screen Shot 2019-05-13 at 10 35 03 AM](https://user-images.githubusercontent.com/6888609/57630014-d7433580-756a-11e9-9a57-09fefdd67741.png)

## Actual Behavior
![Screen Shot 2019-05-13 at 10 29 54 AM](https://user-images.githubusercontent.com/6888609/57630024-dc07e980-756a-11e9-8237-2e181abc29d8.png)

## Code Example
```
let lottieView = AnimationView(name: "testfile")
lottieView.frame = CGRect(x: 50, y: 200, width: 200, height: 200)
lottieView.backgroundColor = UIColor.white
self.view.addSubview(lottieView)
lottieView.play()
```

## Animation JSON

[basic_text_export.TXT](https://github.com/airbnb/lottie-ios/files/3173457/basic_text_export.TXT)

buba447:
Thanks for filing the bug. Text support is still a very new feature. Ill add this to the list of text bugs and will update this issue when I find a fix!

illusate:
I'm having the same issue.

ferico55:
I've been trying to do something about this
I think the issue was that we're not put the layer's anchor point into consideration

note that this is a rough fix, I'm that there is a better way to do this, can you please suggest?

ferico55:
since #956 has been merged and new release ([3.1.1](https://github.com/airbnb/lottie-ios/releases/tag/3.1.1) I will close this issue

buba447:
Just put up a PR that should fix this. https://github.com/airbnb/lottie-ios/pull/1035

[...]

PR #956, opened 2019-06-20
Fix text vertical alignment

ferico55:
Possible fix for #905 
we're not considering layer's anchor point (were inferred from text justification) and then need to consider the scaling too


buba447:
Tjhis is great! Thanks for grabbing this.

ferico55:
sure thing @buba447 I'm very glad to help, will look for more ways to contribute, please guide me as well

michaelfransen:
If the anchor point is set to the bottom of a text layer, should it be rendered bottom aligned? So far, I have not found a way to make bottom aligned text that is multiline.

PR #961, opened 2019-06-24
Fix layer usage for macOS platform

nekrich:
Fixes failing build on macOS platform.


buba447:
Thanks for grabbing this!

PR #962, opened 2019-06-25
Nil out completion block once animation finishes

thedrick:
We were having issues where this completion block was being called multiple times. My assumption is that it should be getting niled out after it's called once, but let me know if I'm mistaken.

### DIFF
--- a/lottie-swift/src/Private/Utility/Helpers/AnimationContext.swift
+++ b/lottie-swift/src/Private/Utility/Helpers/AnimationContext.swift
@@ -46,7 +46,7 @@ class AnimationCompletionDelegate: NSObject, CAAnimationDelegate {
   var ignoreDelegate: Bool = false
   var animationState: AnimationContextState = .playing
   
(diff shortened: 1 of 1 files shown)


buba447:
Can you give me a repo for the multiple completion block call? Are you checking if finished flag?

buba447:
@thedrick Ive looked into this further, this will definitely break app background behavior.
There was a bug that caused completion blocks to be called twice that I fixed with https://github.com/airbnb/lottie-ios/pull/926 
Do you think this is related?  Any chance I could get a repo for the bug?

thedrick:
@buba447 I think this was actually our fault. The spinning Airbnb animation that we play to switch between Host / Guest should play once before we dismiss the modal. To do that, we used the completion block on the animation, but that modal is only hidden and never actually removed from the view hierarchy or deallocated which was causing the completion block to get called every time the animation looped and finished. I've fixed it by properly deallocating the view and removing it from the hierarchy so that the completion block gets deallocated when we are done with it.

PR #974, opened 2019-07-03
Resolve stop() not rewinding animation visually

### LINKED ISSUE #970: Stop not rewinding animation visually
- [✅ ] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [✅ ] This issue isn't related to another open issue

## This issue is a:
- [✅ ] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [] Feature Request
- [✅ ] Regression (Something that once worked, but doesn't work anymore)

Lottie 3.1.0

## What Platform are you on? 

- [] MacOS
- [✅ ] iOS

## What Language are you in?
- [✅ ] Swift
- [] Objective-C

## Expected Behavior 

Calling stop on AnimationView should stop and rewind animation to the first frame (both internal variables and visually)

## Actual Behavior

Calling stop on AnimationView stops the animation and seems to rewind internal variables but visually the animation is paused, even if we call forceDisplayUpdate() afterwards.
This was working on 2.5.2

## Code Example

let animationView = AnimationView(...)
animationView.loopMode = .loop
animationView.play()
DispatchQueue.main.asyncAfter(deadline: .now() + ...) { animationView.stop() }

ferico55:
Trying to resolve #970

`forceDisplayUpdate` is called too soon
the `CATransaction` has not finished doing the transaction yet (just committed and not done yet) therefore the `forceDisplayUpdate()` called too soon, so I propose to force display on the completion block of the CATransaction
This results when we're trying to draw, `CALyer.presentation()` still holds old value of `currentFrame`

### DIFF
--- a/lottie-swift/src/Public/Animation/AnimationView.swift
+++ b/lottie-swift/src/Public/Animation/AnimationView.swift
@@ -734,7 +734,9 @@ final public class AnimationView: LottieView {
     CATransaction.setDisableActions(true)
     animationLayer?.currentFrame = newFrame
     CATransaction.commit()
-    animationLayer?.forceDisplayUpdate()
+    CATransaction.setCompletionBlock {
+        self.animationLayer?.forceDisplayUpdate()
+    }
   }
   
   @objc override func animationWillMoveToBackground() {


buba447:
Nice catch! Thanks for grabbing this!

PR #977, opened 2019-07-09
Fix alpha mask image rendering for gradients

### LINKED ISSUE #884: Gradients fill inverted when rendered
I am having an issue where, when testing on iOS, the gradients are displaying as inversed of how they were originally set up. This seems to be causing loading errors as well.

Gradients are supported but I'm using gradients with a drop in opacity to create a fadeout. Is this the issue? I didn't see anything while looking through the supported features that would suggest it wouldn't. Any help would be greatly appreciated.

- [X] This issue isn't related to another open issue

## This issue is a:
- [] Crashing Bug
- [] Feature Request

Lottie 2.0

## What Platform are you on? 

- [] MacOS
- [X] iOS

## Expected Behavior 
abduction-pizza

## Actual Behavior

abduction-iOS

## Animation JSON
[abduction-test.zip](https://github.com/airbnb/lottie-ios/files/3109330/abduction-test.zip)

amlynarczyk:
The mask image colors on iOS and macOS (probably tvOS also, but I can't check that) should be inverted, so for alpha == 1 the actual color to draw should be black. This fix resolves #884

### DIFF
--- a/lottie-swift/src/Private/NodeRenderSystem/Nodes/OutputNodes/Renderables/GradientFillRenderer.swift
+++ b/lottie-swift/src/Private/NodeRenderSystem/Nodes/OutputNodes/Renderables/GradientFillRenderer.swift
@@ -43,7 +43,7 @@ class GradientFillRenderer: PassThroughOutputNode, Renderable {
       if alpha < 1 {
         drawMask = true
       }
-      if let color = CGColor(colorSpace: maskColorSpace, components: [1.0 - alpha, 1]) {
+      if let color = CGColor(colorSpace: maskColorSpace, components: [alpha, 1]) {
         alphaLocations.append(colors[i])
         alphaColors.append(color)
       }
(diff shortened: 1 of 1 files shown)


amlynarczyk:
@buba447 can you review that change?

buba447:
Hmmm. Im not sure whats going on here. I just did a test with the current version of Lottie-ios and it seems that gradients with transparency is working just fine. This PR will break gradient rendering in most normal cases. Perhaps there is some other error that is causing the gradient to invert in the issue you linked to. Theres definitely an error but it's not this. 

Check out my screenshots below:
After Effects Gradient with alpha:
Screen Shot 2019-07-10 at 10 29 20 AM
Screen Shot 2019-07-10 at 10 31 26 AM

Rendered by Lottie-ios:
Screen Shot 2019-07-10 at 10 33 56 AM

Rendered with the change in this PR:
Screen Shot 2019-07-10 at 10 35 21 AM

buba447:
[GradientTest.zip](https://github.com/airbnb/lottie-ios/files/3378292/GradientTest.zip)

Heres a file that I used to test the gradient rendering.

amlynarczyk:
You are right, the problem is affecting only vertical gradients with alpha. I think that it's because of the flipped geometry and adding
`let flipVertical = CGAffineTransform(a: 1, b: 0, c: 0, d: -1, tx: 0, ty: CGFloat(maskContext.height))
      maskContext.concatenate(flipVertical)` transform to maskContext before applying inContext.ctm seams to fix it, but I need a more complicated gradient example to check that,

PR #980, opened 2019-07-11
Fix alpha mask image rendering for gradients

amlynarczyk:
From what I tested, the mask image is rendered vertically flipped. In previous attempt I had a symmetrical gradient, that's why inverting colors worked, but the real issue is not with colors of rendered mask image, but with that flipped coordinates.


amlynarczyk:
@buba447 here is another try, I'm attaching my test gradient.

[test.json.zip](https://github.com/airbnb/lottie-ios/files/3382128/test.json.zip)

amlynarczyk:
From what I tested, the mask image is rendered vertically flipped. In previous attempt I had a symmetrical gradient, that's why inverting colors worked, but the real issue is not with colors of rendered mask image, but with that flipped coordinates.

buba447:
Ahh that makes a lot of sense! Nice catch! Thanks for grabbing this.

PR #983, opened 2019-07-11
public font list and fonts

coreyd303:
- making fonts public allows for much better debugging as well as font injection at run time allowing for the font to be changed between animation runs assuming the animation view is re-inited


buba447:
Im not sure how making FontList public helps in this case.

This doesnt _exactly_ provide the ability to change fonts at runtime. Any method of changing fonts using this would be a massive workaround. 
I would suggest adding a `fontForKeyPath` to `AnimationTextProvider` and piping the fonts through that method to keep uniformity.

buba447:
https://github.com/airbnb/lottie-ios/pull/1243 adds a font provider to Lottie. Closing this.

PR #986, opened 2019-07-16
fix issue in CompoundBezierPath

erlendboe:
The changed code follows defensive programming, and does not change the behaviour of the original code in any way - I hope you will accept this change 😄 

After upgrading from Lottie 2 to Lottie 3, our testers reported crashes that were hard to reproduce. Luckily I managed to get it to crash in the debugger, and found out that this was the cause.


rocxteady:
Hello. Do you think this commit can fix my issue? https://github.com/airbnb/lottie-ios/issues/1019
Thanks

erlendboe:
Hi @rocxteady
Your crash log looks very similar to the crashes that we saw, so the answer is yes

rocxteady:
Hi @erlendboe

Thank you for your response. The crashes you saw only happened on iOS 10 like ours?

PR #998, opened 2019-07-29
Update accessibility label in every case

kernandreas:
Hello.

I did run today in a situation in which the `AnimatedSwitch` did not have the accessibility value set.

After looking into the code I saw that the value is only updated if the state is changed with an animation. So this does not happen in all cases (e.g. after initialization).

This PR moves the call of the `updateAccessibilityLabel()` method before the `guard` statement.


buba447:
Thanks for grabbing this!

PR #1006, opened 2019-08-12
Fix animation completion when app moves to background

rohitpal440:
**Issue:**

While the animation view is waiting to play animation and app moves to background and the comes to foreground... play completion block were not getting called.

**Fix**

These are the scenario when Animationview is waiting to play animation and app moves to background  --> foreground

- for `backgroundBehavior == .pauseAndRestore`: call the completion block after restoring the context and playing animation
- for other state call completion immediately

### DIFF
--- a/lottie-swift/src/Public/Animation/AnimationView.swift
+++ b/lottie-swift/src/Public/Animation/AnimationView.swift
@@ -764,13 +764,16 @@ final public class AnimationView: LottieView {
       case .pause:
         removeCurrentAnimation()
       case .pauseAndRestore:
-        currentContext.closure.ignoreDelegate = true
+        currentContext.closure.ignoreDelegate = !waitingToPlayAimation
         removeCurrentAnimation()
         /// Keep the stale context around for when the app enters the foreground.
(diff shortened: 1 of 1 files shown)


lottie-swift/src/Public/Animation/AnimationView.swift
```diff
@@ -764,14 +764,17 @@ final public class AnimationView: LottieView {
       case .pause:
         removeCurrentAnimation()
       case .pauseAndRestore:
-        currentContext.closure.ignoreDelegate = true
+        currentContext.closure.ignoreDelegate = !waitingToPlayAimation
         removeCurrentAnimation()
         /// Keep the stale context around for when the app enters the foreground.
         self.animationContext = currentContext
```
buba447: The completion block of an animation _should_ be called after restored as-is. The context is stored, along with its completion block, here. When the app resumes the animation and its completion block is started again and should be called.
If that is not happening then there is a larger bug in the animation context. This solution is more of a bandaid that will further mask the real bug.

PR #992, opened 2019-07-19
Set minimum iOS version to 9.0 for project and Package.swift

eneko:
While the [iOS framework target has its minimum version required set to iOS 9.0](https://github.com/airbnb/lottie-ios/blob/531eb22d47a9439a2344fdaf96832189f2c6e925/Lottie.xcodeproj/project.pbxproj#L2355), both the project and `Package.swift` were set to iOS 12.0.

This update allows for Lottie to be used via SwiftPM on iOS 9 projects.

PR #1010, opened 2019-08-20
Update Package.swift base platform to match podspec

maniramezan:
Take back the support to iOS 9.0, tvOS 9.0, and watchOS 2.0. For adding support to build macOS requires some structural changes in the files so commented it out for not.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,8 +3,7 @@ import PackageDescription
 
 let package = Package(
     name: "Lottie",
-    platforms: [.iOS(.v12)],
-    // platforms: [.iOS("9.0"), .macOS("10.10"), tvOS("9.0"), .watchOS("2.0")],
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -3,8 +3,7 @@ import PackageDescription
 
 let package = Package(
     name: "Lottie",
-    platforms: [.iOS(.v12)],
-    // platforms: [.iOS("9.0"), .macOS("10.10"), tvOS("9.0"), .watchOS("2.0")],
+    platforms: [.iOS(.v9), /*.macOS(.v10_10), */.tvOS(.v9), .watchOS(.v2)],
```
buba447: Why is macOS commented out?

maniramezan: Getting macOS works, requires changing some code changes and move things around so for now I commented it out as it wasn't also originally supported.

maniramezan:
This is already fixed in another [PR](https://github.com/airbnb/lottie-ios/pull/992) and merged in. Going to close this.

PR #1017, opened 2019-08-27
Add method to CompatibleAnimationView to change color values

fabiankr:
This PR adds a method to `CompatibleAnimationView` that makes it possible to change color values from Objective-C. The Objective-C compatibility wrapper misses the methods to set value providers. I assume this is because it'd require adding wrappers for the value providers which require Swift closures.

Why does it matter?
In a project that is mainly Objective-C we wanted to use Lottie for `UIBarButtonItem` instances that animate. However, when a modal view is presented those buttons wouldn't dim correctly when the `tintAdjustmentMode` was set to `UIViewTintAdjustmentModeAutomatic`. With this simple change it's possible to tint the animations with the correct tint color.

It'd be easy to add other methods for these static value providers (e.g. float, size and point), but color was the only one we needed.

### DIFF
--- a/lottie-swift/src/Private/Utility/Primitives/ColorExtension.swift
+++ b/lottie-swift/src/Private/Utility/Primitives/ColorExtension.swift
@@ -69,13 +69,13 @@ extension Color {
   }
   
   var cgColorValue: CGColor {
-    var colorspace : CGColorSpace?
-
+    let colorspace : CGColorSpace
     if #available(iOS 9.3, *) {
-        // Always use P3 colorspace for now
-        colorspace = CGColorSpace.init(name: CGColorSpace.displayP3)
+      // Always use P3 colorspace for now
+      colorspace = CGColorSpace.init(name: CGColorSpace.displayP3) ?? CGColorSpaceCreateDeviceRGB()
+    } else {
+      colorspace = CGColorSpaceCreateDeviceRGB()
     }
-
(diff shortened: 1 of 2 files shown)


lottie-swift/src/Private/Utility/Primitives/ColorExtension.swift
```diff
@@ -69,6 +69,13 @@ extension Color {
   }
   
   var cgColorValue: CGColor {
-    return CGColor(colorSpace: CGColorSpaceCreateDeviceRGB(), components: [CGFloat(r), CGFloat(g), CGFloat(b), CGFloat(a)]) ?? Color.clearColor
+    var colorspace : CGColorSpace?
+
+    if #available(iOS 9.3, *) {
+        // Always use P3 colorspace for now
+        colorspace = CGColorSpace.init(name: CGColorSpace.displayP3)
+    }
```
buba447: For easier to follow code try this:

```
let colorspace : CGColorSpace
if #available(iOS 9.3, *) {
    // Always use P3 colorspace for now
    colorspace = CGColorSpace.init(name: CGColorSpace.displayP3)
} else {
	colorspace = CGColorSpaceCreateDeviceRGB()
}
return CGColor(colorSpace: colorspace, components: [CGFloat(r), CGFloat(g), CGFloat(b), CGFloat(a)]) ?? Color.clearColor
```

ksuther:
Thanks for the suggestion. CGColorSpace.init() returns an optional so it still requires checking for a null return value, but otherwise using that structure now.

PR #1063, opened 2019-10-28
Added value provider for gradient colors properties

enriquebk:
This PR implements a value provider (`ColorsValueProvider`) that can be assigned to a key path property 'Colors'. This allows users to easily set custom gradient colors to `GradientFillNode` or `GradientStrokeNode` objects.

Usage example:

```swift
let colorsKeyPath = AnimationKeypath(keypath: "Back.Ellipse 1.Gradient Fill 1.Colors")
let colorsValueProvider = ColorsValueProvider([UIColor.red.lottieColorValue,
                                               UIColor.blue.lottieColorValue])
animationView.setValueProvider(colorsValueProvider, keypath: colorsKeyPath)
```

The `ColorsValueProvider` can be initialized with a `Color` array or a `ColorsValueBlock` and it can also receive the color locations inside the gradient.

Also this PR adds a check inside the `render` method in `GradientFillRenderer` to prevent out of bounds errors if the user adds fewer colors than the number of Colors set inside the lottie file.

### DIFF
--- a/lottie-swift/src/Public/DynamicProperties/ValueProviders/ColorsValueProvider.swift
+++ b/lottie-swift/src/Public/DynamicProperties/ValueProviders/ColorsValueProvider.swift
@@ -1,0 +1,133 @@
+//
+//  ColorsValueProvider.swift
+//  lottie-swift
+//
+//  Created by Enrique Bermúdez on 10/27/19.
+//
+
+import Foundation
+import CoreGraphics
+
+/// A `ValueProvider` that returns a Gradient Color Value.
+public final class ColorsValueProvider: AnyValueProvider {
+    
+    /// Returns a [Color] for a CGFloat(Frame Time).
+    public typealias ColorsValueBlock = (CGFloat) -> [Color]
+    /// Returns a [Double](Color locations) for a CGFloat(Frame Time).
+    public typealias ColorLocationsBlock = (CGFloat) -> [Double]
+    
+    /// The colors values of the provider.
+    public var colors: [Color] {
+        didSet {
+            hasUpdate = true
+        }
+    }
+    
+    /// The color location values of the provider.
+    public var locations: [Double] {
+        didSet {
+            hasUpdate = true
+        }
+    }
+    
+    /// Initializes with a block provider.
+    public init(block: @escaping ColorsValueBlock,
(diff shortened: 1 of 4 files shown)


lottie-swift/src/Public/DynamicProperties/ValueProviders/ColorsValueProvider.swift
```diff
@@ -0,0 +1,133 @@
...
+        var colorValues = [Double]()
+        var alphaValues = [Double]()
+        var shouldAddAlphaValues = false
+        
+        for i in 0..<colors.count {
+
+            if colors[i].a < 1 { shouldAddAlphaValues = true }
+            
+            let location = locations.count > i ? locations[i] :
+                calculateLocation(withStartValue: locations.last ?? 0.0,
```
buba447: I'm not sure why this is necessary. The location block and the color block should just return the data needed in the format expected. Recalculating the data returned by the blocks every frame will be a hit on performance.

lottie-swift/src/Public/DynamicProperties/ValueProviders/ColorsValueProvider.swift
```diff
@@ -0,0 +1,133 @@
...
+//  lottie-swift
+//
+//  Created by Enrique Bermúdez on 10/27/19.
+//
+
+import Foundation
+import CoreGraphics
+
+/// A `ValueProvider` that returns a Gradient Color Value.
+public final class ColorsValueProvider: AnyValueProvider {
```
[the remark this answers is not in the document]
enriquebk: 👍

lottie-swift/src/Public/DynamicProperties/ValueProviders/ColorsValueProvider.swift
enriquebk: Oh, I didn’t consider that, thanks for letting me know. Definitely the `calculateLocation` method needs to be removed, but I think that the data blocks will still need to be formatted since the gradient array has a very specific format. When initializing using `init(_colors, locations)` the formatting can be done once during the init instead of `value(frame)`.

buba447:
LGTM!

Issue #1066, opened 2019-10-31
Lottie Swift with Xcode 11.2 Crashes when creating AnimationView 


gabors:
@thedrick 

- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [] This issue isn't related to another open issue

## This issue is a:
- [] Non-Crashing Bug (Visual or otherwise)
- [x] Crashing Bug
- [] Feature Request
- [] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.1.3 release, nut current master branch also crashes
Using Lottie as a SwiftPM package

## What Platform are you on? 

- [] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [] Objective-C

## Actual Behavior

<img width="1253" alt="Screen Shot 2019-10-31 at 16 19 04" src="https://user-images.githubusercontent.com/3587589/67990015-3882c200-fbfa-11e9-9c3f-8b6931e4fb03.png">

## Code Example

let x = AnimationView(name: "test")

## Animation JSON
Any JSON animation from your example project

danielsaidi:
I just noticed this as well.

mborsten:
I get this crash too with Xcode 11.2. It happens when setting an animation on the animation view, not when creating the animation view.

makoni:
Same for me. That happened with Xcode 11.2 beta too

mborsten:
This crash does not seem related to 13.2. It happens on simulator and devices running 13.2 and 12.4.2.

danielsaidi:
It's a compiler issue, right? I can run our prod app with the latest Lottie version when it's built with the 5.1 compiler, but not with the latest compiler.

gabors:
> It's a compiler issue, right? I can run our prod app with the latest Lottie version when it's built with the 5.1 compiler, but not with the latest compiler.

Well yeah, if you build with Xcode 11.1 or earlier you're fine. 
Issue is using Xcode 11.2 which you need if you want to debug on iOS 13.2.

thedrick:
Could someone set up a sample project or link me to an animation that is causing this crash? I can't reproduce this on my end.

rouviere:
I am having the same problem. Anything Lottie on Xcode 11.2 will not run. Here is a link to a repository: https://github.com/rouviere/lottie

thedrick:
[...]

PR #1069, opened 2019-11-01
Update to 3.1.4

thedrick:
This updates the version to 3.1.4. I'm also hoping there's some change that was made recently that will fix the crash people are seeing 🙏

### DIFF
--- a/Example/Podfile.lock
+++ b/Example/Podfile.lock
@@ -1,5 +1,5 @@
 PODS:
-  - lottie-ios (3.1.3)
+  - lottie-ios (3.1.4)
 
 DEPENDENCIES:
   - lottie-ios (from `../`)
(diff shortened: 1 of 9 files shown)


buba447:
We should wait until the current pending pull requests are reviewed and merged. Im back from my honeymoon today so I will go through them tomo and check it all out.

Hold off on merging this.

buba447:
Alright @thedrick im going to merge this and publish the release. Just a heads up [this crash](https://github.com/airbnb/lottie-ios/issues/1066) will not be fixed by this release. The crash is caused by a bug with swift packaging manager and there is currently an open bug filed for it with Apple [here](https://bugs.swift.org/browse/SR-11564). There is a workaround by setting a build flag.
The bug only happens when importing lottie (or any swift package that uses generics) using swift packaging manager.

Cheers!

PR #1078, opened 2019-11-15
AnimationView remove Final class

shelbygold:
LOTAnimation was able to be inherited, that was changed with the new update to AnimationView. We need to remove final from the animationView class to continue to allow this class to be inherited.


buba447:
Instead of subclassing it is suggested to wrap animation view. Subclassing can lead to errors with the animation view's lifecycle and animation.

PR #1079, opened 2019-11-15
AnimationView -final

shelbygold:
Final taken from AnimationView to make it inheritable.


buba447:
Instead of subclassing it is suggested to wrap animation view. Subclassing can lead to errors with the animation view's lifecycle and animation.

buba447:
Closing this.

shelbygold:
Do you have any information on how to wrap instead of subclassing. I have been subclassing LOTAnimation for years I would really like to not have to rewrite a bunch of code. If you could help I would really appreciate it.

PR #1089, opened 2019-11-26
Revert "Updated Swift Package to include MacOS"

buba447:
Reverts airbnb/lottie-ios#1088

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,22 +2,17 @@
 import PackageDescription
 
 let package = Package(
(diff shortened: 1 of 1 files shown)


dreampiggy:
What's the Reason lottie on SwiftPM does not support tvOS and macOS ?

Though the CocoaPods version support it.

I'm the downstream dependency https://github.com/SDWebImage/SDWebImageLottiePlugin

PR #1108, opened 2020-01-06
Fix animation not playing when dynamically using negative speeds

Joao360:
Basically when I was giving **dynamic** negative animation speeds, through **lottie-react-native**, the animation wouldn't play at all.

After debugging it through **lottie-react-native** and **lottie-ios** I found out the issue was on not using the absolute value on the calculation of **layerAnimation.beginTime**.

After making this change, the animation would start to play in reverse again.


buba447:
Nice catch!

PR #1111, opened 2020-01-18
refactor: Avoid structs with arrays

Rapsssito:
I would like to contribute my grain of sand to the overall performance improvement project.

### DIFF
--- a/lottie-swift/src/Private/NodeRenderSystem/Nodes/PathNodes/EllipseNode.swift
+++ b/lottie-swift/src/Private/NodeRenderSystem/Nodes/PathNodes/EllipseNode.swift
@@ -82,7 +82,7 @@ final class EllipseNode: AnimatorNode, PathNode {
     
     let cp = half * EllipseNode.ControlPointConstant
     
-    var path = BezierPath(startPoint: CurveVertex(point: q1,
+    let path = BezierPath(startPoint: CurveVertex(point: q1,
                                                   inTangentRelative: CGPoint(x: -cp.width, y: 0),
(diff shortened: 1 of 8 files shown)


Rapsssito:
@buba447, I would like to first test the performance improvement that this PR might do. **Is there any heavy-load animation against which I can test this?**

buba447:
This looks like a good start. Have you done any performance testing? It would be great to see some numbers.

As far as performance issues and testing goes, there are two performance issues with Lottie:
1. Serializing the `Animation` data model. When loading a large json into an animation the performance is really bad. This is because Lottie uses the swift Codable, which isnt performant for large files.
2. Playback performance. Currently playback performance is eating a LOT of CPU. This happens no matter what the size of animation file is. So you can take any animation file, load it up and play it on loop to see the performance impact. A larger animation will not really make a difference. As far as I can tell this issue comes from swifts under-the-hood memory reference counting for structs. Which this PR does some to help, but the real issue might be with the keyframe group generic type. Theres a lookupmap for generics that I theorize is causing most of the issues.

buba447:
Otherwise LGTM! Are you done with this PR @Rapsssito ?

Rapsssito:
@buba447, I'll try to get some real numbers and then I'll let you know.

Rapsssito:
@buba447, I was not able to get any significant numbers testing it on real devices or simulators. However, as you said, this may affect the **Playback performance** issue positively!

The PR is ready!

Rapsssito:
After further testing, I have discovered that this PR does not improve performance. I will close this one and open a new one that shows a much more noticeable performance improvement.

Issue #1113, opened 2020-01-21
Crashing when the views window goes away


zenangst:
Hey folks!

We are seeing some crashes occurring when the applications root window goes away.
(Or is about to be replaced by a new window)
The way our application works is when the user logs in, we create a new window for
the application to use, we do this to keep the code-path clean and tidy and to make sure
that we present everything in the right context. However, because of this architectural choice,
we are seeing crashes when this happens.

Have you encountered this kind of crash before (left a stack trace at the bottom).

The instruction pointer stops in the `deinit` method for the `AnimationView`.

<img width="526" alt="image" src="https://user-images.githubusercontent.com/57446/72810966-f7c9a080-3c5e-11ea-8775-bdbc5f813422.png">

https://github.com/airbnb/lottie-ios/blob/master/lottie-swift/src/Public/Animation/AnimationView.swift#L622

According to the docs, you shouldn't have to invoke `removeObserver` in `deinit`

> If your app targets iOS 9.0 and later or macOS 10.11 and later, you don't need to unregister an observer in its dealloc method.

https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver#

Is there a particular reason to explicitly do this?

## This issue is a:
- [x] Crashing Bug

## Which Version of Lottie are you using?

Lottie 3.1.5

## What Platform are you on? 

- [x] iOS/tvOS

## What Language are you in?
- [x] Swift

## Actual Behavior
The application crashes

## Stacktrace

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x7faf2f60816)
    frame #0: 0x00007fff3fc0a553 libobjc.A.dylib`objc_autorelease + 19
    frame #1: 0x00007fff39b82cd7 UIKitCore`-[UIView(Hierarchy) window] + 27
    frame #2: 0x00007fff39b9aa15 UIKitCore`-[UIView(UIViewGestures) removeAllGestureRecognizers] + 440
    frame #3: 0x00007fff39b6acf3 UIKitCore`-[UIView dealloc] + 607
  * frame #4: 0x000000010fbaade3 Lottie`AnimationView.__deallocating_deinit(self=0x00007faf2f605aa0) at AnimationView.swift:623:3
    frame #5: 0x000000010fbaae1b Lottie`@objc AnimationView.__deallocating_deinit at <compiler-generated>:0
[...]

PR #1114, opened 2020-01-21
Improve animation view

zenangst:
- Remove `removeObserver` call in `AnimationView.deinit`
- Add guard to `animationMovedToWindow()` to ensure that state change don't happen to orphan `AnimationView`'s

Reference issue: #1113

### DIFF
--- a/lottie-swift/src/Public/Animation/AnimationView.swift
+++ b/lottie-swift/src/Public/Animation/AnimationView.swift
@@ -820,6 +820,11 @@ final public class AnimationView: LottieView {
   }
   
   override func animationMovedToWindow() {
+    /// Don't update any state if both the `superview` and `window` is `nil`
+    guard window != nil && superview != nil else {
+      return
+    }
+
     if window != nil {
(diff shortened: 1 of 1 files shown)


lottie-swift/src/Public/Animation/AnimationView.swift
```diff
@@ -824,6 +820,11 @@ final public class AnimationView: LottieView {
   }
   
   override func animationMovedToWindow() {
+    /// Don't update any state if both the `superview` and `window` is `nil`
+    guard window != nil && superview != nil else {
+      return
+    }
```
emrcftci: Please use early-return instead of new lines.

`guard window != nil && superview != nil else {
    return
  }`

zenangst: Hey @emrcftci, pushed another commit fixing the syntax :)

emrcftci:
It looks good for me PTAL @buba447

buba447:
LGTM! Nice catch.

PR #1118, opened 2020-01-27
refactor: Keyframe<T> Contiguous Arrays

Rapsssito:
Another grain of sand to the overall performance improvement project.

After some testing on real devices and simulators, I was able to reduce 3% RAM usage.


lottie-swift/src/Private/NodeRenderSystem/NodeProperties/ValueProviders/GroupInterpolator.swift
```diff
@@ -15,30 +15,19 @@ final class GroupInterpolator<ValueType>: AnyValueProvider where ValueType: Inte
   }
   
   func hasUpdate(frame: CGFloat) -> Bool {
-    for interpolator in keyframeInterpolators {
-      if interpolator.hasUpdate(frame: frame) {
-        return true
-      }
-    }
-    return false
+    let updated = keyframeInterpolators.first(where: {$0.hasUpdate(frame: frame)})
```
buba447, line 18: Nice!

PR #1135, opened 2020-02-28
Fixes zero based CALayer for Precomposition.

migs647:
Our team had a number of animations that were partially drawing that were working before. Narrowed down the commit that broke to 6ede9ee. Because the bounds were not set before hand, all of the layers bounds are set to zero. This brings back not setting the bounds before the layers are configured in order to avoid clipping.

### DIFF
--- a/lottie-swift/src/Private/LayerContainers/CompLayers/PreCompositionLayer.swift
+++ b/lottie-swift/src/Private/LayerContainers/CompLayers/PreCompositionLayer.swift
@@ -28,8 +28,9 @@ final class PreCompositionLayer: CompositionLayer {
     }
     self.frameRate = frameRate
     super.init(layer: precomp, size: CGSize(width: precomp.width, height: precomp.height))
+    bounds = CGRect(origin: .zero, size: CGSize(width: precomp.width, height: precomp.height))
     contentsLayer.masksToBounds = true
-    contentsLayer.bounds = CGRect(origin: .zero, size: CGSize(width: precomp.width, height: precomp.height))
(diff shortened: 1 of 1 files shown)


lottie-swift/src/Private/LayerContainers/CompLayers/PreCompositionLayer.swift
```diff
@@ -28,8 +28,9 @@ final class PreCompositionLayer: CompositionLayer {
     }
     self.frameRate = frameRate
     super.init(layer: precomp, size: CGSize(width: precomp.width, height: precomp.height))
+    bounds = CGRect(origin: .zero, size: CGSize(width: precomp.width, height: precomp.height))
     contentsLayer.masksToBounds = true
-    contentsLayer.bounds = CGRect(origin: .zero, size: CGSize(width: precomp.width, height: precomp.height))
+    contentsLayer.bounds = bounds
```
buba447: Maybe I am reading this wrong, but the code seems to be nearly identical to what it was before. How does declaring the frame as a variable before setting it change anything?

dreampiggy: The below code use the local variable bounds value. The left one does not update them.

buba447: Oh I see! Silly me. Thanks for grabbing this!

lamjay2704: Look great

PR #1140, opened 2020-03-05
Added possibility to draw animation frame via CALayer.

Shedward:
Removed perspective and added some backdoors to enable rendering frame via CALayer.render(in context: CAContext) to avoid pressure on main thread.


Shedward:
Sorry, meant to make it in my fork.

J-Arji:
how can use it 
can you say with example?

PR #1143, opened 2020-03-08
Fixed potential memory leak due to strong delegate reference in AnimationContext

tarekabdallah:


yosimasu:
I use this to patch lottie-ios 3.1.9.
It fixes my issue.

My Issue:
I use lotte-react-native to make a loading view component.
When I try this component in a device(iPhone XS, iOS 14.5) back and forth, it will crash at about 10 times.

PR #1149, opened 2020-03-18
Fix method :-animationMovedToWindow judgment logic

liulichao123:
Method :- logic bug in animationMovedToWindow:

> `guard window != nil && superview != nil else { return } `
>     `if window != nil {`
>     `updateAnimationForForegroundState()`
> `} else { `
>      `//  the following will never be executed, resulting in the current Lottie not recording the previous state`
>       `updateAnimationForBackgroundState()`
> `}`

It should be changed to:
> `guard superview != nil else { return }`
>     `if window != nil {`
>       `updateAnimationForForegroundState()`
>     `} else {`
>       `updateAnimationForBackgroundState()`
>    ` }`

Note: When A viewA owns superViewB, it removes the superViewB from the window. At this point, viewA still owns superViewB and triggers the viewA method: -didmovetowindow

### DIFF
--- a/lottie-swift/src/Public/Animation/AnimationView.swift
+++ b/lottie-swift/src/Public/Animation/AnimationView.swift
@@ -826,8 +826,9 @@ final public class AnimationView: LottieView {
   }
   
   override func animationMovedToWindow() {
-    /// Don't update any state if both the `superview` and `window` is `nil`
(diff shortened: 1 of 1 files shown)


buba447:
Nice catch!

PR #1156, opened 2020-03-25
fix keypath traversal

RolandasRazma:
correct me if I'm wrong, but I think layer by key path lookup is wrong

### DIFF
--- a/lottie-swift/src/Private/Utility/Extensions/AnimationKeypathExtension.swift
+++ b/lottie-swift/src/Private/Utility/Extensions/AnimationKeypathExtension.swift
@@ -98,14 +98,13 @@ extension KeypathSearchable {
       return nil
     }
     
-    if nextKeypath.nextKeypath != nil {
-      /// Now check child keypaths.
-      for child in childKeypaths {
-        if let layer = child.layer(for: keyPath) {
-          return layer
-        }
+    /// Now check child keypaths.
+    for child in childKeypaths {
+      if let layer = child.layer(for: nextKeypath) {
+        return layer
       }
     }
+    
     return nil
   }
   


RolandasRazma:
if this PR is correct, wonder about https://github.com/airbnb/lottie-ios/blob/master/lottie-swift/src/Private/Utility/Extensions/AnimationKeypathExtension.swift#L75 - that looks wrong as well

jonasman:
I confirm this fixes the keypath issue

jonasman (approved):

RolandasRazma:
what about https://github.com/airbnb/lottie-ios/pull/1156#issuecomment-604274077

isimpson (approved):
This definitely fixes an issue when adding subviews to a Keypaths using:
`public func addSubview(_ subview: AnimationSubview, forLayerAt keypath: AnimationKeypath)`

Can we please have this merged and released asap.

isimpson (approved):

isimpson:
@thedrick do you think this fix could be merged as its blocking us using that latest release?

calda:
This seems like a reasonable change. To merge this, I think we'd need to:
 1. Rebase and resolve merge conflicts
 2. Add a test case to the regression suite. You can add animation JSON files to `Tests/Samples`, and configure `AnimationKeypath`s for the test case [here](https://github.com/airbnb/lottie-ios/blob/master/Tests/SnapshotConfiguration.swift#L42).

I'm closing this PR because it's gone stale, but feel free to rebase and re-open after making updates.

isimpson:
Seems to have been fixed here: https://github.com/airbnb/lottie-ios/pull/1587

Issue #975, opened 2019-07-05
call stop() to reset animation without reverse animation in visual!


Calabash-Boy:
## Check these before submitting:
- [✅] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [] This issue isn't related to another open issue

## This issue is a:
- [✅] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [] Feature Request
- [] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.1.1

## What Platform are you on? 

- [] MacOS
- [✅] iOS

## What Language are you in?
- [✅] Swift
- [] Objective-C

## Expected Behavior 
I want to play the animation from 0 to 0.5, then stop it, the animationView should go back to start frame without animation

## Actual Behavior
Actually, when i call stop() in the play(fromProgress....), a reverse animation came out, this is weird;

## Code Example
```
var aniView: AnimationView!
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        aniView = AnimationView(name: "star")
        aniView.frame = CGRect(x: 50, y: 50, width: 100, height: 100)
        aniView.backgroundColor = UIColor.lightGray
        self.view.addSubview(aniView)
    }
    
    override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
        aniView.play(fromProgress: 0, toProgress: 0.5, loopMode: nil) { (finished) in
            self.aniView.stop()
        }
    }
```

## Animation JSON
```
[...]

PR #1157, opened 2020-03-25
Fix AnimationView.Stop()

7jason:
This fixes https://github.com/airbnb/lottie-ios/issues/970. That issue was not actually resolved.

And also fixes https://github.com/airbnb/lottie-ios/issues/975

Before this change, calling stop() on an AnimationView that is playing did not work properly. It would not return to the first frame of the animation. And when it did, it would "reverse" the animation.

The reversing is because [this PR](https://github.com/airbnb/lottie-ios/pull/974) put CATransaction.setCompletionBlock _outside_ the CATransaction begin/commit block. CATransaction.setCompletionBlock needs to occur between CATransaction.begin and CATransaction.commit.

This PR moves that setCompletionBlock to the correct place.

After moving that code block, the animation would still not return to frame 0 when stop is called, but when a subsequent animation is play()'d, the animation would return to frame 0 _without reversing_.

It seemed clear to me, based on this behaviour, that some of the CATransaction stuff or the setNeedsDisplay() system is not working properly.

To remedy this, I have put a flush() in stop().

This fixes the problem. But I realize that the flush shouldnt need to be there. There are some deeper bugs in lottie-ios, but _at least this fixes the symptom_.

Tagging @buba447 @dcramps @ferico55 @thedrick

### DIFF
--- a/lottie-swift/src/Public/Animation/AnimationView.swift
+++ b/lottie-swift/src/Public/Animation/AnimationView.swift
@@ -373,6 +373,7 @@ final public class AnimationView: LottieView {
   public func stop() {
     removeCurrentAnimation()
     currentFrame = 0
+    CATransaction.flush()
   }
   
   /**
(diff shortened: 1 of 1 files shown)


liulichao123:
I also found this problem before, and it seems to have something to do with the keyword ‘@NSManaged’
# 1130

PR #1166, opened 2020-03-31
fix text not being updated if textProvider is set after animation

RolandasRazma:
fix text not being updated if textProvider is set after animation. This doesn't fix case where text change - you will need to set new (or same) provider again


buba447 (approved):

calda:
@RolandasRazma, this seems like a reasonable change. We can probably merge this after rebasing / resolving merge conflicts.

calda:
Closing since there are unresolved merge conflicts. Feel free to re-open once the merge conflicts are resolved.

PR #1173, opened 2020-04-04
 Enables support for distributing Lottie as a binary library

### LINKED ISSUE #1081: Module compiled with Swift 5.1 cannot be imported by the Swift 5.1.2 compiler
## This issue is a:
- [] Crashing Bug

lottie-ios 3.1.5, lottie-react-native 3.3.2

## What Platform are you on? 

- [] MacOS
- [] iOS
- Xcode Version 11.2.1 (11B500)

## What Language are you in?
- [] Swift
- React Native

## Actual Behavior
`Module compiled with Swift 5.1 cannot be imported by the Swift 5.1.2 compiler` when building. Have removed Pods/ and installed again, tried different toolchains.

Unable to run project does anyone know how to fix this?

Thanks!

devioustree:
This fixes #1081.

**Solution**

To allow frameworks to be used by multiple Swift version, we need to set:

```
BUILD_LIBRARY_FOR_DISTRIBUTION = YES
```

This isn't necessary for a Swift package (since they are build from source), but it's useful for Carthage or other means of distributing binary frameworks.

**Testing**

I built `Lottie.framework` in Xcode 11.3.1. Then in Xcode 11.4 I created a test project and linked to the generated project. It built and linked successfully.

### DIFF
--- a/Lottie.xcodeproj/project.pbxproj
+++ b/Lottie.xcodeproj/project.pbxproj
@@ -2327,6 +2327,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				APPLICATION_EXTENSION_API_ONLY = YES;
+				BUILD_LIBRARY_FOR_DISTRIBUTION = YES;
 				CODE_SIGN_IDENTITY = "";
 				CODE_SIGN_STYLE = Manual;
 				DEFINES_MODULE = YES;
@@ -2355,6 +2356,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				APPLICATION_EXTENSION_API_ONLY = YES;
+				BUILD_LIBRARY_FOR_DISTRIBUTION = YES;
 				CODE_SIGN_IDENTITY = "";
 				CODE_SIGN_STYLE = Manual;
 				DEFINES_MODULE = YES;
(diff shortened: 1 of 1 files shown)


Cyberbeni:
I think you just need to run `carthage bootstrap` in your build phase and it will rebuild the library if it was built with a different Swift version. At least when downloading the prebuilt binary, Swift version is checked to be compatible: https://github.com/Carthage/Carthage#swift-binary-framework-download-compatibility
If that doesn't work, then it's a Carthage issue.

Cyberbeni:
This was also discussed in the CocoaPods repo: https://github.com/CocoaPods/CocoaPods/issues/9232#issuecomment-566713762

Cyberbeni:
Read the discussion I linked.

devioustree:
Is your concern that the performance would be degraded?

Cyberbeni:
That is one of the concerns. 3.x is already bad enough to warrant people opening issues and using 2.x.
But the point is still that `BUILD_LIBRARY_FOR_DISTRIBUTION` is meant for libraries that are distributed separately from apps. Carthage just provides a better version of DerivedData, so Xcode gets confused by the amount of code less often.

Seeing that this repository doesn't provide prebuilt binaries (that can still be used with the same Xcode version that they were built with, even without this flag), this could also introduce differences between the same version on different machines even if they are using the same Xcode version currently. Sure, that doesn't happen with every Xcode release but trust me, it's not fun to go crazy about a mysterious bug.

calda:
This imposes a large amount of constraints on the public API of a framework, so I don't think we should enable it for Lottie.

From "[Library Evolution in Swift](https://www.swift.org/blog/library-evolution/)":

> Library evolution support is turned off by default. Frameworks that are always built and distributed together, such as Swift Package Manager packages or binary frameworks that are internal to your app, should not be built with library evolution support.
>
> Library evolution support should only be used when a framework is going to be built and updated separately from its clients.

PR #1184, opened 2020-04-22
FIX:Cache bug in class LRUAnimationCache

liulichao123:
When adding a cache, check the cache size and remove the earliest cache(cacheMap & lruList)


buba447:
Was fixed here https://github.com/airbnb/lottie-ios/pull/1119

PR #1210, opened 2020-06-12
AnimationSelectionViewController added

antonioanchondo:
Added a Table View Controller to the example iOS code allowing the animations to be selected from a list and thus removing the need for hardcoded animation names and recompile to see the different animations.


calda:
Thanks for this suggestion. Resolved in #1435.

PR #1211, opened 2020-06-12
Add isAnimationQueued to determine if animation will start when this view is added to a window

jlukas:
In Lottie v2, isAnimationPlaying would return true if the animation isn't playing, but would start as soon as the view is added to a window. In Lottie v3, isAnimationPlaying returns true only if an animation is actually playing.

Rather than modify how isAnimationPlaying works, I've added isAnimationQueued to detect the state where the animation isn't playing, but will when the view is added to a window. This allows the app I work on to upgrade from Lottie v2 to v3 and preserve the existing behaviour.

### DIFF
--- a/lottie-swift/src/Public/Animation/AnimationView.swift
+++ b/lottie-swift/src/Public/Animation/AnimationView.swift
@@ -110,6 +110,11 @@ final public class AnimationView: LottieView {
     return animationLayer?.animation(forKey: activeAnimationName) != nil
   }
   
+  /// Returns `true` if the animation will start playing when this view is added to a window.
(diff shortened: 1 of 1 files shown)


pbatg:
Hi @keakalamudo, is anything else needed to merge this commit?

FYI @jlukas is now at a different employer and may no longer have access to this GitHub account.

Cee:
+1. Is there any update?

pbatg:
@buba447 Could you help review this PR?

buba447:
LGTM!

buba447 (approved):

PR #1215, opened 2020-06-23
Work Around Xcode 12 Beta 1 Compilation Issues

danl3v:
Fixes a compilation issue while archiving on Xcode 12 Beta 1

### DIFF
--- a/lottie-swift/src/Private/Utility/Extensions/AnimationKeypathExtension.swift
+++ b/lottie-swift/src/Private/Utility/Extensions/AnimationKeypathExtension.swift
@@ -217,15 +217,15 @@ extension String {
     }
     if let index = self.firstIndex(of: "*") {
       // Wildcard search.
-      let prefix = self.prefix(upTo: index)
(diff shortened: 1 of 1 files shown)


museum-future:
Unless this fix works with other versions of xcode, i would not pull this into the main branch.  This patch needs to be explicity defined for xcode 12+

danl3v:
Hi @museum-future. Thanks for taking a look! `Substring` has the same interface as `String`, so the behavior will be the same for prior versions. `Substring` just provides a memory optimization. See https://developer.apple.com/documentation/swift/substring

That said, happy to wrap the patch up if you like

agiokas:
It would be great if this PR would get reviewed and merged.

danl3v:
No idea! Someone from Airbnb needs to approve and merge it!

buba447:
Lgtm thanks for grabbing this!

danl3v:
It is possible in a future xcode beta that these changes will be unnecessary. Will open a revert if this is the case down the line

PR #1246, opened 2020-08-15
fix: empty layerID causes crash

abigaleypc:
If the layerID is not set, the app would crash.

### DIFF
--- a/lottie-ios/Classes/Models/LOTLayerGroup.m
+++ b/lottie-ios/Classes/Models/LOTLayerGroup.m
@@ -38,7 +38,9 @@
                                       withAssetGroup:assetGroup
                                        withFramerate:framerate];
     [layers addObject:layer];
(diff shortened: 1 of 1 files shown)


abigaleypc:
https://dlied5sdk.myapp.com/music/release/upload/t_mm_file_publish/2682081.zip 这个资源文件是通过AE直接导出来的，它会让iPhone设备crash，原因是没有layerID这个key, 无法从layerID获取值，但源码里却有读取layerID的逻辑，这将导致设备crash。

abigaleypc:
https://dlied5sdk.myapp.com/music/release/upload/t_mm_file_publish/2682081.zip 

This resource file was exported directly through AE, which would cause the iOS app to crash.
By default, lottie will read layerID from resource file, while it doesn't have the logic to handle with empty layerID, which would cause the app to crash.

Issue #1266, opened 2020-09-28
imageAssets can be nil when AnimationImageProvider.imageForAsset(asset:) is nil..


mqzkim:
## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Feature Request

## Which Version of Lottie are you using?
Lottie 3.1.8

## What Platform are you on? 
- [x] iOS

## What Language are you in?
- [x] Swift

## Expected Behavior 
Lottie assets could have some resources that AnimationImageProvider doesn't have.
and Android Lottie checks if the AnimationImageProvider have or not.
https://github.com/airbnb/lottie-android/blob/7371eabe326ac70991cf237da423b03ee52caba8/lottie/src/main/java/com/airbnb/lottie/manager/ImageAssetManager.java#L81

## Actual Behavior
`setImage nil from imageProvider.imageForAsset(asset: asset)`
https://github.com/airbnb/lottie-ios/blob/37c6075e949d395d83dd669fd01668ba88fb5774/lottie-swift/src/Private/LayerContainers/Utility/LayerImageProvider.swift#L44

## Code Example
PR: https://github.com/airbnb/lottie-ios/pull/1268
```
if let asset = imageAssets[imageLayer.imageReferenceID],
   let image = imageProvider.imageForAsset(asset: asset) {
  imageLayer.image = image
}
```

PR #1268, opened 2020-09-28
imageLayer.image no changes when imageForAsset is nil

mqzkim:
issue: https://github.com/airbnb/lottie-ios/issues/1266

Lottie assets could have some resources that AnimationImageProvider doesn't have.
also Android Lottie checks if the AnimationImageProvider have or not.


calda:
I think we likely want `imageLayer.image` to always match the result of calling `imageProvider.imageForAsset(asset: asset)`, even when it returns `nil`.

Issue #1117, opened 2020-01-27
Consider using ZippyJSON as a drop in improvement to Codable Performance Issues?


kyle-fox:
## Check these before submitting:
- [X] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [X] This issue isn't related to another open issue

Instead of trying to find time to make improvements yourself, or writing your own JSON parsing again, it might be worth testing out Lottie after just dropping in [ZippyJSON](https://github.com/michaeleisel/zippyjson) as a dependency? I came across it in [this thread](https://forums.swift.org/t/rearchitecting-jsonencoder-to-be-much-faster/28139). I know it's not an ideal solution, but it could serve as a placeholder to get better performance while waiting for a more permanent solution. Unfortunately, I haven't had the chance to test this myself, but when I do, I'll follow up here. Hoping your test suite would be a better place to run this anyway!

liulichao123:
The swift version is more time-consuming to load Lottie than the OC(2.5.3) version, which I found to be caused by the different JSON parsing, especially when the APP started, loading multiple Lottie animations, which seriously slowed down the startup speed.
I hope this problem can be improved

sheff1422:
@kyle-fox Unfortunately that is not completely drop in solution. I am working on that for the second day. And ZippyJSON behaves strange when decoding a heterogeneous list of objects for a given family.
I am trying to get in touch with ZippyJSON supporter and get more info. Rn It throws some exceptions.

kyle-fox:
Closing, since it doesn't seem to be worth the change. Feel free to reopen if that changes!

michaeleisel:
hi, author of zippyjson here. i've gotten a number of issues/prs from people concerning uses of lottie with it, so it seems popular and i'm happy to offer support. zippyjson is intended to be a drop-in replacement, and if you see any difference feel free to report it. sheff's planning to test the new release of zippyjson which has a fix for that issue. hopefully that's the last remaining blocker

kyle-fox:
Might as well re-open, since it seems to be a possibility, at least? Sorry that I don't have time to do any further testing at the moment. Thanks for the info Michael!

sheff1422:
@kyle-fox I will check if that is working for me this week.

zabolotiny:
[...]

PR #1275, opened 2020-10-17
Use ZippyJSONDecoder for JSON Decoding

michaeleisel:
Following discussion in https://github.com/airbnb/lottie-ios/issues/1117

ZippyJSONDecoder is a ~3-4x faster drop-in replacement of JSONDecoder

### DIFF
--- a/lottie-ios.podspec
+++ b/lottie-ios.podspec
@@ -33,6 +33,7 @@ For the first time, designers can create and ship beautiful animations without a
   s.tvos.exclude_files = 'lottie-swift/src/Public/MacOS/**/*'
   s.osx.exclude_files = 'lottie-swift/src/Public/iOS/**/*'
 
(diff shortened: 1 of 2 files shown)


michaeleisel:
How can I run lottie-ios's unit tests with this?

buba447 (commented):
Lottie-ios is currently built without any third party libraries and for the foreseeable future we are going to keep it that way.

For some context:
Lottie is not just a cocoapod, it is also built as a static library, swift library, and additionally available on carthage. A third party library would have to be added to all of the builds, which would require quite a bit of work and a change of build process. Additionally added dependencies further complicate many of the projects that use lottie.

Sorry for the trouble!

michaeleisel:
What if the consumer could customize which JSON decoder was used, meaning they'd do their own management of ZippyJSON?

PR #1282, opened 2020-11-04
fix: Fixed imageProvider not loading animation image

eharrison:
When setting an image path to the AnimationView, the image would not load because the AnimationLayer.imageProvider was not updated. This fix updates the imageProvider in the AnimationLayer with the one set in the AnimationView.

How to test:
Unzip the file below, open animation/with_image.json with the AnimationView and set animationView.imageProvider = FilepathImageProvider(filepath: "THE_PATH_TO_IMAGES_FOLDER")
[animation-external-image.lottie.zip](https://github.com/airbnb/lottie-ios/files/5488179/animation-external-image.lottie.zip)


lottie-swift/src/Public/Animation/AnimationView.swift
```diff
@@ -402,6 +402,7 @@ final public class AnimationView: LottieView {
   
   /// Reloads the images supplied to the animation from the `imageProvider`
   public func reloadImages() {
+    animationLayer?.imageProvider = imageProvider
```
buba447, line 405: This should go in the ImageProvider setter method. Up around line 94.

```
 public var imageProvider: AnimationImageProvider {
    didSet {
    animationLayer?.imageProvider = imageProvider
      reloadImages()
    }
  }
```

buba447 (changes requested):

lottie-swift/src/Public/Animation/AnimationView.swift
eharrison: Done :)

buba447 (approved):

PR #1283, opened 2020-11-09
Specify data collection practices of this SDK

bachand:
Apple is requiring application to publish "nutrition facts" related to data privacy.

Lottie is a widely used SDK.

Let's make it easy for consumers of Lottie to know how including Lottie in their application affects what they need to report to Apple.

The good news is that it should have no effect!

While I was here I made a line break above `### Objective-C Support`. This section was technically below `## Installing Lottie` so this seemed like a nice clarification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,8 @@ If you have doubts, please, check the following links:
 
 After successfully retrieved the package and added it to your project, just import `Lottie` and you can get the full benefits of it.
 
+-----
+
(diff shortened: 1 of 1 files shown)


buba447:
Thanks for grabbing this Michael! Also good to see ya!

LGTM

buba447 (approved):

bachand:
You too @buba447 . Hope you are well :)

PR #1294, opened 2020-11-30
Modify: update mask imp follow Swift 3.1.9 Version

778477:
as [Swift 3.1.9 MaskContainerLayer UpdateWithFrame:](https://sourcegraph.com/github.com/airbnb/lottie-ios/-/blob/lottie-swift/src/Private/LayerContainers/CompLayers/MaskContainerLayer.swift#L130) Mask Layer render path use a large frame.


778477:
@thedrick  please take a look this mr. thx

PR #1295, opened 2020-11-30
Video layers support

NSGolova:
**New feature request - Video Layers**

Video layers are similar to image layers. It is a precompiled file with some content to be integrated into animation.
They are useful for creating rich animations and presentations.

Added support for video layers and video providers. The video will be played from the first frame when it appears. Video playback is not affected by animation speed or current progress(it can be added if needed).

Also added such features:
- video looping -- If you set boolean `loopVideos`, the animation will play visible videos in a loop after finish or pause.
- transition on video substitution from video provider -- The video provider can change the content at any moment and this change will be animated with a simple fade.

Sample video layer in AE. The filename is used as a composition name because "bodymovin" is not exporting videos as assets.
Screenshot 2020-11-30 at 12 10 48 PM

Sample animation:
Comp 1_3_3

Same animation in Lottie:
ezgif-2-0841c2271a7d


radulov:
@buba447, please review the new feature =)

calda:
Thanks for the suggestion. Since Lottie doesn't have support for this on [other platforms](https://airbnb.io/lottie/#/supported-features), I don't think we should support this on iOS at this time.

Issue #1254, opened 2020-09-09
Crash in AnimationView.stop


davidpasztor:
## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Crashing Bug
## Which Version of Lottie are you using?

Lottie 3.1.8

## What Platform are you on? 

- [x] iOS 13.5

## What Language are you in?
- [x] Swift

## Expected Behavior 

I should be able to call `AnimationView.stop()` without it crashing.

## Actual Behavior

When stopping the animation, I get a crash in `AnimationView.stop()` on the last line, `CATransaction.flush()`. Sadly this crash is quite intermittent, so I am not sure how exactly to reproduce it.
>Thread 1: EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)
The code is called on the main thread.

## Code Example
```swift
/// SwiftUI wrapper for `AnimationView`
@available(iOS 13, *)
public struct AnimationViewWrapper: UIViewRepresentable {

    private let animation: Lottie.Animation?
    private let loopMode: LottieLoopMode
    @ObservedObject private var animate: ObservableContainer<Bool>

    /// Designated initialiser
    /// - parameter animationName: the name of the animation to load
    /// - parameter animate: controls whether the animation is playing or not. Setting this property to `true` starts the animation, while setting it to `false` stops it.
    /// - parameter loopMode: how the animation should behave once it starts playing
    public init(animationName: String, animate: ObservableContainer<Bool>, loopMode: AnimationLoopMode = .playOnce) {
        self.animation = Animation.named(animationName, bundle: Bundle(for: BundleLocator.self))
        self.loopMode = loopMode.lottie
        self.animate = animate
    }

    public func makeUIView(context: UIViewRepresentableContext<Self>) -> UIView {
        // Wrapping `animationView` in a `UIView` with `NSLayoutConstraint`s is necessary for `animationView` to follow the size set on the SwiftUI wrapper
        let view = UIView()

        let animationView = AnimationView(animation: animation)
        animationView.loopMode = loopMode
        animationView.contentMode = .scaleAspectFit

        animationView.translatesAutoresizingMaskIntoConstraints = false
        view.addSubview(animationView)

        NSLayoutConstraint.activate([
[...]

PR #1313, opened 2021-01-29
Fix: remove flush

Kaelzs:
I found a [pull request](https://github.com/airbnb/lottie-ios/pull/1157) to fix AnimationView stop problem. It says `stop()` not working and adds `CATransaction.flush()` in the stop method to fix it.

In my opinion, adding flush in this function is not a good way to fix it, because it will cause all the UI elements to render (or layout). and may lead to some [crash problem](https://github.com/airbnb/lottie-ios/issues/1254).

After reading the animation source code, I finally found the reason why animation couldn't revert to the first frame after `stop()` is that `AnimationContainer.display()` is still called once after stopping, and at this time the AnimationContainer gets the `currentFrame` of the copy value in `presentation()`, which is not zero at this time, to resolve this problem, I think we can check if the animation exists. If the AnimationContainer does not have an animation, then its presentation is meaningless and we should not use the value of the `curerntFrame` in the presentation. Since the `needsDisplay(forKey:)` in AnimationContainer for Lottie-customized key will always return true, we don't need to worry about the display function not called.

and here comes my pull request.

### DIFF
--- a/lottie-swift/src/Private/LayerContainers/AnimationContainer.swift
+++ b/lottie-swift/src/Private/LayerContainers/AnimationContainer.swift
@@ -193,7 +193,14 @@ final class AnimationContainer: CALayer {
   
   public override func display() {
     guard Thread.isMainThread else { return }
-    var newFrame: CGFloat = self.presentation()?.currentFrame ?? self.currentFrame
+    var newFrame: CGFloat
+    if let animationKeys = self.animationKeys(),
+      !animationKeys.isEmpty {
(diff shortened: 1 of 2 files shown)


haojianzong:
+1, stopping the animation shouldn't have a side effect to for flushing CATransaction, which causes an animation glitch for host view on the App I am working on.

buba447 (approved):
LGTM Thanks for grabbing this!

buba447 (approved):
LGTM Thanks for grabbing this.

PR #1317, opened 2021-02-10
Fixed a deprecation warning where some protocols were inheriting from class instead of AnyObject per Xcode's suggestion

brunnerjosh:
I was updating my iOS app to be able to run on Xcode 12.5 (beta) and noticed that Apple is now flagging the use of class inherited protocols as deprecated in favor of inheriting from AnyObject. In my iOS app, we flag warnings as errors so now our app won't compile. I wanted to be able to pitch in and help modernize the code so I went ahead and fixed this deprecation here.

Screen Shot 2021-02-10 at 1 40 35 PM

After a little bit of research, I came across this forum Swift.org [post](https://forums.swift.org/t/class-only-protocols-class-vs-anyobject/11507/4) which seems to point to the discussion where this deprecation discussion started.

### DIFF
--- a/lottie-ios.podspec
+++ b/lottie-ios.podspec
@@ -8,7 +8,7 @@
 
 Pod::Spec.new do |s|
   s.name             = 'lottie-ios'
-  s.version          = '3.2.1'
+  s.version          = '3.2.2'
   s.summary          = 'A library to render native animations from bodymovin json. Now in Swift!'
 
   s.description = <<-DESC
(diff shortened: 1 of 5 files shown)


apps4everyone:
@buba447 would a 3.2.2 Release possible? Thank you

brunnerjosh:
@buba447 @apps4everyone I bumped the version to 3.2.2. If you'd like me to not bump the version in this PR, I can remove that commit. Thanks!

buba447:
Closing this as another PR made this change and also caught a few other instances.

Issue #1325, opened 2021-03-19
Animations always use screen scale as the rendering scale, which isn't always preferred


JoeSzymanski:
## Check these before submitting:
- [X] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [X] This issue isn't related to another open issue

## This issue is a:
- [X] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [X] Feature Request
- [] Regression (Something that once worked, but doesn't work anymore)

While investigating performance issues around large, complex animations in Lottie 3 (all of which worked fine in Lottie 2.5.x), I discovered that the drawing context in Lottie 3 was significantly larger than the same animation running in Lottie 2. After some digging, I discovered that Lottie 3 will use the screen's scale as the rendering scale, which can significantly impact performance of large (i.e. screen-sized) animations. For something like gradients, this can make animations almost impossible to run in a performant manner.

Lottie 3 doesn't expose any way to modify the render scale of the animation layer, so there's no way to tune the tradeoffs between the accuracy of the drawing versus the performance. With a simple test of forcing the render scale to 1, the animations play extremely well.

The render scale should be exposed in some way to allow the developer to choose to run with a different render scale than the screen for performance reasons.

JoeSzymanski:
I will be looking into create a PR for this, but I need to spend some time figuring out the optimal way to manage exposing the render scale, since I don't think the animation layer should be exposed directly, but I need to make sure the correct scale is used in all cases.

absolute-heike:
Great find @JoeSzymanski!!

I was playing around with this idea and this is what I found:
* In ObjC (2.5.2) `contentsScale` is 1.0 for all sublayers
* Setting `contentsScale` to `1.0` in lottie-swift has **no** visual impact at all
* `contentsScale = 1.0` has tremendous impact on Performance. Some numbers with a complex test animation:
  * Swift contentsScale 3.0: ~70% CPU Utilization
  * Swift contentScale 1.0: ~40% CPU Utilization
  * ObjC (defaults to scale 1.0): ~37% CPU Utilization
  
I know that documentation tells us to adjust the contentsScale of a programatically created Layer. But what do we gain by setting contentsScale to 3.0? Except for hotter iPhone? I haven't found any differences so far.

absolute-heike:
contentsScale 3.0 vs contentsScale 1.0 comparison

[...]

PR #1327, opened 2021-03-22
Add support for modifying renderScale

JoeSzymanski:
This fixes the issues in #1325 by allowing the render scale to be set by the caller. Setting the render scale to 1 appears to significantly improve performance of animations with little visible changes in quality.

Allowing this to be set by the caller seemed to be the safest path forward, as it won't change existing behavior at all. An optional value was used to allow for the default setting of "self.screenScale" to be accessed properly, as it's not 100% clear that the self access for default values would be availble.

In testing some of our most complex animations, this resolved the issues seen related to performance changes between Lottie 2.x and Lottie 3.x.

### DIFF
--- a/lottie-swift/src/Public/Animation/AnimationView.swift
+++ b/lottie-swift/src/Public/Animation/AnimationView.swift
@@ -74,6 +74,19 @@ final public class AnimationView: LottieView {
       }
     }
   }
+
+  /**
+   Sets the rendering scale to use for animations. If set to nil, the render scale will be
+   set to the screen scale.
+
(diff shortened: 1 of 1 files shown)


absolute-heike:
I played around with `renderScale` and (except for the performance impact) didn't find any regressions when not setting `contentsScale` at all. We can probably remove it alltogether 🤔 

https://github.com/airbnb/lottie-ios/pull/1328

JoeSzymanski:
Closing this as I think #1328 is probably a better way to resolve this issue.

Issue #1314, opened 2021-02-03
High memory and CPU overhead


loganblevins:
## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [] This issue isn't related to another open issue

## This issue is a:
- [x] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [x] Feature Request
- [] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.2.0

## What Platform are you on? 

- [] MacOS
- [x] iOS 14.4 (likely does not matter)

## What Language are you in?
- [x] Swift
- [] Objective-C

## Expected Behavior 

Low memory and CPU overhead.

## Actual Behavior

High memory and CPU overhead
iPhone 11 Pro Max gets physically hot while sitting on a screen with a [stock Lottie animation](https://lottiefiles.com/43885-laptop-working)

![Screen Shot 2021-02-03 at 3 54 46 PM](https://user-images.githubusercontent.com/12498682/106808795-d7397200-6638-11eb-9e9f-1d6eb92008e9.png)

## Animation JSON

https://lottiefiles.com/43885-laptop-working

loganblevins:
@buba447 As I understand this is an open-source project without full time maintenance, a proper fix for performance could be time intensive and not possible in the short term.

Is there a suitable workaround? Maybe a version I can downgrade to while remaining compatible with Swift 5 (and preferably SPM)?

If not, is there anything I can do to help out?

Thanks Brandon!

Tr736:
We also discovered this issue. A temp fix is to use the OBJ-C lottie version. in 2.x.x

RamblinWreck77:
IIRC buba mentioned the issue had to do with swift structs consuming more memory/CPU than he expected, and that a rewrite of several sections would be necessary. 

IMO this library is so useful big companies should sponsor @buba447 to work on it full time, or Apple should should add native support lol

hassanalinali:
I am also having very high memory consumption. On an iPhone 12 Pro that has more RAM than the standard, the app with 2 Lottie animations barely has 5 fps, while memory consumption is at 1.5GB and CPU at 90%. This is unusable

anishparajuli555:
In the same boat myself!

calda:
[...]

Issue #1060, opened 2019-10-19
Wrong Gradients & Bad Performance when showing them


Rapsssito:
I am experiencing 1 FPS on UI when more than one animation with gradients is being reproduced on iOS. After some profiling, turns out that **`CGContextDrawLinearGradient`** is taking more than 90% of CPU time. Is there a missing swift optimization?
Furthermore, it looks like some gradients are not shown completely. I have attached a JSON that shows both errors.

## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Non-Crashing Bug (Visual or otherwise)

## Which Version of Lottie are you using?

Lottie 3.0

## What Platform are you on? 
- [x] iOS

## What Language are you in?
- [x] Swift

## Expected Behavior 
![hexgood](https://user-images.githubusercontent.com/44206249/67143541-c5c62e00-f26c-11e9-9db8-333f40e8685d.PNG)

## Actual Behavior
![hexbad](https://user-images.githubusercontent.com/44206249/67143543-c8288800-f26c-11e9-81d0-cfb6af597c1f.PNG)
_Note the white corner on the bottom left._

## Animation JSON
[Animation JSON](https://github.com/airbnb/lottie-ios/files/3746457/menuSecondaryButton.zip)

This might be related to this issue: https://github.com/airbnb/lottie-ios/issues/895

buba447:
You are experiencing two issues here.

1. The gradient is correctly displaying, its the **shape** that the gradient is filling that is incorrect. How is this shape being created? Is it a polygon with rounded corners or a shape layer?
2. The gradient performance is slow, drawing gradients is expensive, but 1fps makes me think something else is wrong. Mind sharing the after effects file so I can debug when I have time?

Rapsssito:
I think the issue is **indeed related to the gradient generation**. I have attached my AE file. It has three compositions, all look the same except for the gradient colors, but **only the composition _"comp 2"_ shows this issue**.
[menuMainButton.zip](https://github.com/airbnb/lottie-ios/files/3877181/menuMainButton.zip)

buba447:
I will look at this over the weekend!

Rapsssito:
Thank you! If you need anything from my end, tell me.

buba447:
[...]

Issue #895, opened 2019-04-26
Performance regression on Lottie 3


manolosavi:
## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Non-Crashing Bug (Visual or otherwise)
- [ ] Crashing Bug
- [ ] Feature Request
- [x] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?
Lottie 3.0.4

## What Platform are you on? 
- [ ] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [ ] Objective-C

## Expected Behavior 
Same or improved performance as previous versions of Lottie when running animations

## Actual Behavior
Running an animation on a loop using Lottie 2.5.3 consumes around 16% CPU, while running that same animation on Lottie 3.0.4 consumes around 43%

## Animation JSON
[Attached file](https://github.com/airbnb/lottie-ios/files/3123084/lottie.animation.zip)

buba447:
Duplicate of https://github.com/airbnb/lottie-ios/issues/832 and https://github.com/airbnb/lottie-ios/issues/828

manolosavi:
@buba447 for what its worth we also tried installing off of `master` today and we didn't see any difference in CPU usage

buba447:
I imagine there wouldnt be any difference. I havent had a chance to deep dive into the performance issues yet. The issues will be updated when I do though!

alexbredy:
Same issue here. I am updating the `currentProgress` property of a Lottie animation on scrollViewDidScroll (creating a parallax offset) and performance dropped a lot since we migrated from v2 to v3. CPU consumption jumps making the frame rate drop.

manolosavi:
@buba447 hi! just wanted to let you know that I just tried out version 3.1.2 and there are still issues in this version. The same screen shows CPU utilization of ~45% on 3.1.2 vs ~18% on 2.5.3, and the reported "energy impact" is "very high" vs "high".

buba447:
Yeah I still haven't had a chance to work on the performance issues. They are caused by some limitations in the swift language cause by automatic memory management in structs. To fix the issue requires a pretty hefty refactor that I haven't had the resources to tackle. (Im working on lottie in my personal free time)
We had to bump to 3.1 because of an API change.
[...]

PR #1328, opened 2021-03-22
Fix Swift performance issues. Remove renderScale and contentsScale

absolute-heike:
Related to #1325 #895 #1060 #1314 (and possibly others)

lottie-swift sets `layer.contentsScale` on every sublayer. ObjC didn't do that and this seems to explain the performance regression from 2.x to 3.x. Removing that code, I haven't found any regressions whatsoever while at the same time fixing the performance issues.

Profiling these changes gave the following insights:
* Swift contentsScale 3.0: ~70% CPU Utilization
* **Swift contentScale 1.0: ~40% CPU Utilization**
* ObjC (defaults to scale 1.0): ~37% CPU Utilization

### DIFF
--- a/lottie-swift/src/Private/LayerContainers/CompLayers/CompositionLayer.swift
+++ b/lottie-swift/src/Private/LayerContainers/CompLayers/CompositionLayer.swift
@@ -143,7 +143,7 @@ class CompositionLayer: CALayer, KeypathSearchable {
   }
   
   func updateRenderScale() {
-    self.contentsScale = self.renderScale
+    /// To be overridden by subclass
   }
 }
 
--- a/lottie-swift/src/Private/NodeRenderSystem/RenderLayers/ShapeContainerLayer.swift
+++ b/lottie-swift/src/Private/NodeRenderSystem/RenderLayers/ShapeContainerLayer.swift
@@ -66,7 +66,6 @@ class ShapeContainerLayer: CALayer {
   }
   
   func updateRenderScale() {
-    self.contentsScale = self.renderScale
     renderLayers.forEach( { $0.renderScale = renderScale } )
   }
   
--- a/lottie-swift/src/Private/NodeRenderSystem/RenderLayers/ShapeRenderLayer.swift
+++ b/lottie-swift/src/Private/NodeRenderSystem/RenderLayers/ShapeRenderLayer.swift
@@ -87,9 +87,4 @@ final class ShapeRenderLayer: ShapeContainerLayer {
     }
     renderer.render(ctx)
   }
-  
-  override func updateRenderScale() {
-    super.updateRenderScale()
-    shapeLayer.contentsScale = self.renderScale
-  }
 }


lottie-swift/src/Private/LayerContainers/CompLayers/TextCompositionLayer.swift
```diff
@@ -91,8 +91,6 @@ final class TextCompositionLayer: CompositionLayer {
   override func displayContentsWithFrame(frame: CGFloat, forceUpdates: Bool) {
     guard let textDocument = textDocument else { return }
     
-    textLayer.contentsScale = self.renderScale
```
absolute-heike, line 94: I didn't test animations with Text Layers. Probably makes sense to check on that before merging

absolute-heike: Okay, so for text there seems to be a notable visible difference when removing contentsScale. I will have a look at how to set contentsScale for TextLayers 🤓

absolute-heike:
@buba447 Seems like @JoeSzymanski found the issue for the performance regression. We are unsure about the best way to tackle this. 

I could not distinguish regular Lottie animations on a pixel level when removing contentsScale (Which is not what I would have expected, so I am wondering whether I overlooked something 😅 ). Only TextLayers become blurry, which is why I removed screenScale everywhere except for TextLayers.

Joes approach was to make renderScale publicly available. https://github.com/airbnb/lottie-ios/pull/1327
Which seems like a cleaner solution from a code point of view but has some downsides for usability. Most users probably won't know about this setting and you have to trade performance for blurry texts ¯\\_(ツ)_/¯

leberwurstsaft:
According to the docs, there is no need to change this value for layers backing views.

https://developer.apple.com/documentation/quartzcore/calayer/1410746-contentsscale

> The default value of this property is 1.0. For layers attached to a view, the view changes the scale factor automatically to a value that is appropriate for the current screen. For layers you create and manage yourself, you must set the value of this property yourself based on the resolution of the screen and the content you are providing. Core Animation uses the value you specify as a cue to determine how to render your content.

joeldrotleff:
This is working great for me. I had severe performance issues in loading and playback that went away completely after using this PR. Much thanks!

buba447:
Nice catch here! Just a few comments to clean this up so we can get it merged.

The original purpose of the `renderScale` property was to provide every layer type the ability to set its own render scale accordingly. This pr removes the renderScale all together.  Instead you should be able to remove the calls to `self.contentsScale = self.renderScale` inside of `updateRenderScale()` for `CompositionLayer` `ShapeContainerLayer` and `ShapeRenderLayer`. This should fix the render issues while keeping a clean way for lottie objects to know about the render scale (such as the text layer)

absolute-heike:
@buba447 Great suggestions. Looks much cleaner now 😊

ErrorPro:
The fix also worked for me: was having a performance issue with Gradients. Any idea when this gonna be merged? @buba447

buba447 (approved):

charlesanim499:
Any thoughts on @marksands comment? @buba447

calda:
Hi all, I reverted this PR since it introduced some visual regressions (#1407). We now have snapshot tests that run on every PR (#1432) and should help catch these sorts of issues in the future.

PR #1352, opened 2021-05-14
Add support for P3 Wide Color Gamut

### LINKED ISSUE #958: [Feature Request] Wide Gamut color space
## Useful Data
**Lottie Version:** 3.1.0
**Platform:** iOS
**Language:** Swift

## Expected Behavior 
When an animation view is initialized, there should be a parameter that tells Lottie about which color space should be used so that we can render it in wide-gamut instead of sRGB.
(The iOS API is this one: https://developer.apple.com/documentation/uikit/uicolor/1648568-init)

## Actual Behavior
All the colors are interpreted in sRGB. It results in a less brilliant animation.

## Samples
The sRGB version (the current one)
lottie_sRGB

the P3 version (with better colors)
lottie_p3

kyle-fox:
Fixes #958

For this PR, I intentionally went with a very simple, global setting for controlling P3 color settings, so I could solicit feedback from the maintainers before digging further! I believe a global setting will be satisfactory for most use cases, but there isn't an existing location for this type of global configuration, as far as I can tell - this simple singleton is based on `LRUAnimationCache`. It would also be necessary to configure the setting in `AnimationColorSpace` as early in the app lifecycle as possible, as this code is simple enough that it won't handle re-rendering any views to update their color space - that probably warrants an addition to the documentation, which I can also provide after merging. I chose to use an intermediary `UIColor` object before converting to `CGColor`, because that API is available in iOS 10, whereas `CGColorSpace` doesn't seem to support P3 color until later versions of iOS.

Open to any direction from the maintainers! With the recent performance improvements to Lottie Swift, this change brings us one step closer to being able to upgrade from v2.5.2 😄

### DIFF
--- a/lottie-swift/src/Private/Utility/Primitives/ColorExtension.swift
+++ b/lottie-swift/src/Private/Utility/Primitives/ColorExtension.swift
@@ -7,6 +7,7 @@
 
 import Foundation
 import CoreGraphics
+import UIKit
 
 extension Color: Codable {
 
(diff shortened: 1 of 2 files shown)


calda:
Thanks for this contribution @kyle-fox. This has merge conflicts that need to be resolved, so I'm am going to close this for now. I am open to merging this if you'd like to rebase this and resolve the merge conflicts, and make a few updates:

 1. We can put the configuration flag for this in the [`LottieConfiguration`](https://github.com/airbnb/lottie-ios/blob/master/Sources/Public/LottieConfiguration.swift) type
 2. We would need a test case in the regression suite that validates this new behavior. You can add animation JSON files to Tests/Samples and then run the test suite to generate the snapshots.
 3. We would need to support this on macOS as well. As is, I think this won't work on macOS since it depends on UIKit.

PR #1358, opened 2021-05-26
Support animations in asset catalogs (NSDataAsset)

dflems:
This was supported in the ObjC version of Lottie. Would be nice to have here as well:

https://github.com/airbnb/lottie-ios/blob/44bda1253360b7ea5698f058f8483e42e26e65bd/lottie-ios/Classes/Private/LOTComposition.m#L38-L42

### DIFF
--- a/lottie-swift/src/Private/Model/Extensions/Bundle.swift
+++ b/lottie-swift/src/Private/Model/Extensions/Bundle.swift
@@ -1,0 +1,21 @@
+import Foundation
+#if os(iOS) || os(tvOS) || os(watchOS) || targetEnvironment(macCatalyst)
+import UIKit
+#endif
+
+extension Bundle {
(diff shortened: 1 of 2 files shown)


lottie-swift/src/Private/Model/Extensions/Bundle.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+#if os(iOS) || os(tvOS) || os(watchOS) || targetEnvironment(macCatalyst)
+import UIKit
+#endif
```
dflems, line 4: usually I just use `#if canImport(UIKit)` but this is the pattern used elsewhere in the project, so I felt it was best to be consistent

dflems:
@buba447

thedrick (approved):
Seems reasonable to me!

PR #1365, opened 2021-06-10
Use scanHexInt64 instead of the deprecated scanHexInt32 in iOS 13

BalestraPatrick:
If you build Lottie-Swift with a deployment target of iOS 13 and treat warnings as errors, the error `'scanHexInt32' was deprecated in iOS 13.0` comes up. 

This PR should fix it without side effects.


shepting (approved):
Looks good to me.

thedrick (approved):
thanks for the fix @BalestraPatrick !

PR #1368, opened 2021-06-16
Fix project after #1358

dflems:
I accidentally broke this in #1358. I developed/verified it in the example app via cocoapod, so I didn't add it to the main project.

This repo could really use some CI jobs to do build / test / validation 😬 

@buba447 @thedrick 

cc @AYastrebov


thedrick (approved):
oh thank you 😬  yeah we could definitely do with some CI here 😅

PR #1377, opened 2021-07-08
Fix multi-platform Swift Package

### LINKED ISSUE #1057: Unable to use Lottie with MacOS app
## Lottie-iOS Issue
Unable to use Lottie for an macOS app. I'm using latest **macOS Catalina** + **Xcode 11.1**
I'm trying to use it with a **SwiftUI** app (if that helps)

Please make it compatible latest Xcode / macOS version

## This issue is a:
- [ ] Crashing Bug
- [x] Feature Request

Lottie 3.1.3

## What Platform are you on? 

- [x] MacOS
- [ ] iOS

## What Language are you in?
- [x] Swift
- [ ] Objective-C

## Expected Behavior 
Compiles without errors

## Actual Behavior
There are no issues installing the library, but unable to compile, saying ``ld: framework not found Lottie``. I tried installing using **Package Manager** method and **Cocoa Pods** method. Neither works.

Screenshot 2019-10-11 at 09 05 29

**_BTW installing and compiling for iOS app yields no errors or warnings._**

## How to reproduce

- Create new project -> macOS -> App
- Add the Lottie library using preferred method (e.g Package Manager)
- Hit Run (Command +R)

## Animation JSON
Unable to make it work. Not at the animation step yet.

msealand:
This fixes #1057, removing the need for macOS projects that use Swift Package Manager to use a separate branch.  SPM only requires that all file names are unique across the entire package, regardless of target platform.  This fix simply renames the few platform specific files, appending either `-iOS` or `-macOS` depending on which platform they are for.  There's no need to exclude them completely since they all have `#if os(...)` blocks around their platform specific code.

I updated Example project with the new file paths and verified that it still worked.  In the macOS target of the Example project, I also updated the code signing settings so that it can be run without signing.


hannesoid:
I briefly tested this and it seems to work - the resulting package can be used on iOS and macOS

Who has the permissions to review & merge this?

calda:
@msealand, this should be fixed by #1428. Does that change work for you?

PR #1384, opened 2021-08-04
feat: add the black and white color mode of lottie json animation;

taoyuxuan:


calda:
Thanks for the suggestion @taoyuxuan -- this seems somewhat nice so I'm not sure we'd want to make this part of Lottie's public API. It may be better to update the json of individual animations to make them use grayscale colors, or use the `AnimationKeypath` API to update the colors to be grayscale.

PR #1391, opened 2021-09-03
Integrate macOS macros for multi-platform support

### LINKED ISSUE #1361: Support multi-platform Xcode Projects (iOS and macOS) in Swift Package Manager

## This issue is a:
- [] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [x] Feature Request

Lottie 3.0

## What Platform are you on? 
- [x] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [] Objective-C

## Expected Behavior 

For a SwiftUI cross-platform application with multiple Xcode targets (iOS AND macOS), I expect to be able to include Lottie for iOS and macOS in the one Xcode project.

i.e. these two:

https://github.com/airbnb/lottie-ios (branch: `master`)
https://github.com/airbnb/lottie-ios (branch: `lottie/macOS-spm`)

## Actual Behavior

Xcode won't allow the unique repo (https://github.com/airbnb/lottie-ios) to be added a second time, so I'm forced to choose one of the two available branches.

Workaround: I can fork the airbnb repo, thereby creating the second repo for Xcode to recognise. 

Better solutions: Either

1.  Move the macOS branch of the `airbnb/lottie-ios` repo to its own `airbnb/lottie-macos` repo, or
2. Support cross-platform within the existing `airbnb/lottie-ios` repo.

Thank you!

chuynadamas:
Hello,

First all awesome work with this project :) -  This PR fix #1361 I added the macro ```#else``` for the duplicated files to be able to use in ```macOS``` and ```iOS``` , this implies that we have to use the same macros to use in shared code

```swift
#if os(iOS) 
//Code for iOS
#else
//Code for other platforms
#endif
``` 

In my particular case, this fix suits my ```CI/CD``` because I don't need to launch a simulator and I can test directly using ```macOS``` configuration.

Also I see the PR #1377 is using a different approach, so the owners can pick whatever they think is better for the library! 
Thanks in advance.


Moriquendi:
👍 Can someone from the Lottie team merge it please?

calda:
@chuynadamas, this should be fixed by #1428. Does that change work for you?

PR #1395, opened 2021-09-15
Fix layout jump

### LINKED ISSUE #1393: AnimationView position glitches on device rotation when inside UIStackView
- [] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)

## This issue is a:
- [] Crashing Bug
- [] Feature Request

Lottie 3.2.3

## What Platform are you on? 

- [] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [] Objective-C

AnimationView is added to UIStackView, along with some other elements.
On device rotation when changing UIStackView axis the animation jumps to some other place that it should be. 
See the below video with AnimationView and UIImageView as reference on what I expected.

https://user-images.githubusercontent.com/78353113/132045307-b007a8a6-b196-411f-b0a8-90657710bd90.mov
https://user-images.githubusercontent.com/78353113/132045323-eeefeffa-88ca-4113-a584-6dc69d0da548.mov

dunkmann00:
When rotating the device I was noticing that the Lottie AnimationView was not always smoothly animating into the new position. After some digging it appears this is because of how `layoutAnimations` was handling setting up the animation.

Currently, if `layoutAnimations` is called multiple times, we will get the ending position as the 'from' value instead of the actual value on screen. This results in a potential visual 'jump' in the view.

This PR uses the layer's presentation layer rather than the model layer to setup the animation. By using the presentation layer's values, we get the values for where the layer is currently on screen, regardless of how many times `layoutAnimations` ends up getting called. This ensures the animation always appears to run smoothly.

I believe this should fix #1393.

### DIFF
--- a/Sources/Public/Animation/AnimationView.swift
+++ b/Sources/Public/Animation/AnimationView.swift
@@ -784,14 +784,14 @@ final public class AnimationView: LottieView {
       let positionAnimation = animation.copy() as? CABasicAnimation ?? CABasicAnimation(keyPath: "position")
       positionAnimation.keyPath = "position"
       positionAnimation.isAdditive = false
-      positionAnimation.fromValue = animationLayer.presentation()?.position
+      positionAnimation.fromValue = (animationLayer.presentation() ?? animationLayer).position
       positionAnimation.toValue = position
       positionAnimation.isRemovedOnCompletion = true
 
       let xformAnimation = animation.copy() as? CABasicAnimation ?? CABasicAnimation(keyPath: "transform")
       xformAnimation.keyPath = "transform"
       xformAnimation.isAdditive = false
-      xformAnimation.fromValue = animationLayer.presentation()?.transform
+      xformAnimation.fromValue = (animationLayer.presentation() ?? animationLayer).transform
       xformAnimation.toValue = xform
       xformAnimation.isRemovedOnCompletion = true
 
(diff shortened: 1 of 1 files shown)


dunkmann00:
@calda bump

calda:
@dunkmann00, would you mind sharing a before/after gif or video recording of this issue? That would help me understand what the problem was (the video links in #1393 no longer seem to open properly)

dunkmann00:
@calda sure thing! I'll put something together later today/tomorrow

dunkmann00:
https://user-images.githubusercontent.com/11963090/145898123-deaf369d-d87d-4071-8523-a48caed60c81.mov


https://user-images.githubusercontent.com/11963090/145898153-1cd7fe2a-2dd6-49ed-9886-f2321f69f937.MP4

calda:
Nice, thanks! Much better.

Sources/Public/Animation/AnimationView.swift
```diff
@@ -784,14 +784,14 @@ final public class AnimationView: LottieView {
       let positionAnimation = animation.copy() as? CABasicAnimation ?? CABasicAnimation(keyPath: "position")
       positionAnimation.keyPath = "position"
       positionAnimation.isAdditive = false
-      positionAnimation.fromValue = animationLayer.position
+      positionAnimation.fromValue = animationLayer.presentation()?.position
```
calda, line 787: It may make sense to handle the case where `presentation()` is `nil`, thoughts?

```suggestion
      positionAnimation.fromValue = (animationLayer.presentation() ?? animationLayer).position
```

calda (approved):

Sources/Public/Animation/AnimationView.swift
calda: I went ahead and made this change in https://github.com/airbnb/lottie-ios/pull/1395/commits/3b57a052226915417ca27862a7758a363a40a74f

dunkmann00:
Great, thanks!

Gam3356 (commented):
Willie Parish IPHONE 12 Pro max

PR #1397, opened 2021-10-06
Avoid automatic animations for opacity changes

absolute-heike:
This fixes a bug for layers, that switch opacity from 1 to 0.

Right now there is an automatic animation added, which leads to a Ghosting like effect


calda (approved):

calda:
@absolute-heike, thanks for this fix! Would you mind rebasing to resolve merge conflicts? Then I'd be happy to merge this.

absolute-heike:
@calda done ☺️

PR #1400, opened 2021-10-11
Fixed missing default layer index

liliomk-supercharge:
Copied the behavior from the Android implementation.

Reference:
[LayerParser.java](https://github.com/airbnb/lottie-android/blob/7d80f43caeb1822a1a34e361b2708a8ba01bb745/lottie/src/main/java/com/airbnb/lottie/parser/LayerParser.java#L81)
`layerId`


calda:
Resolved by #1433

Issue #883, opened 2019-04-23
3D orientation / rotation


atom22:
Hello guys,

My After Effect animation is making use of 3D layers in order to apply rotations on X and Y axis. The rotations are well applied in the bodymovin json (it's ok in preview and in html too), but when imported in iOS these are not working (Lottie 3.0). Same with the orientation transformation.

Is it an unsupported transformation (doesn't look like it is following your list) ? How can I achieve it ?

buba447:
3d Layers are not supported in iOS or Android.

proxemics:
Any updates on this as a possible feature update? I had asked on the Slack and it sounds like this is now possible on iOS and maybe even planned? This is definitely a feature that comes up a lot in my work and would be an amazing addition!

proxemics:
Just curious if this feature is now possible or planned with some of the recent updates to iOS? Thanks!

AlanCourt:
I'm really want this feature too.

isblvgara:
is there a way to transform the 3D Paths into keyframes or something similar?

pawel3ala:
hey guys, does anybody have some info if this feature is supported in the latest version `3.4.x` ?

PR #1402, opened 2021-10-13
add support for 3D rx, ry, rz layer transformations

jonklein:
Adds support for 3D layer transformations as discussed in #883.

I've manually tested this on a few pre-made assets that were rendering differently in the web player vs this native player - please let me know if there's guidance on a more comprehensive approach to testing the rendering!


calda:
Thanks for your contribution @jonklein. I'm going to close this PR since it has merge conflicts that would need to be resolved, but I'd welcome this sort of change if you want to resolve the merge conflicts and re-open it.

To merge this, we would also need to add some animations to the regression suite that use this feature (gifs in the PR body would be helpful as well). You can add animation JSON files to `Tests/Samples` and then run the test suite to generate the snapshots.

PR #1404, opened 2021-10-19
Update OC Compatiblity

mlch911:


calda:
Are you having issues with the current implementation of `CompatibleAnimation`? If so, could you file an issue discussing this, or re-open the PR with a description of why the changes are necessary?

PR #1408, opened 2021-10-27
Add support for tvOS in Package.swift file

daltonclaybrook:
This PR adds support for tvOS in the Package.swift file. tvOS is supported for CocoaPods and Carthage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ xcuserdata/
 
 # other
 build/
+.swiftpm/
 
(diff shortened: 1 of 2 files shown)


daltonclaybrook:
@thedrick Have you had a chance to look at this PR? Seems safe considering tvOS is support in the podspec.

calda:
Fixed in https://github.com/airbnb/lottie-ios/pull/1428

PR #1413, opened 2021-11-07
Issue 1412 cannot parse json in cae of no indecies

chizhavko:
We use lotty not as vector graphics, but as a sequence of JPG images. And in this JSON we do not have their indices. Each picture is a separate object, it is also the initial frame, and this same frame is its own index. Due to the non-index option, we have no way to parse JSON.

In my fix we have to parse ind value, but if there no value we can use as default value start frame which can replace index because for img sequences each image == startFrame == index

### DIFF
--- a/lottie-swift/src/Private/LayerContainers/Utility/CompositionLayersInitializer.swift
+++ b/lottie-swift/src/Private/LayerContainers/Utility/CompositionLayersInitializer.swift
@@ -22,7 +22,7 @@ extension Array where Element == LayerModel {
     var childLayers = [LayerModel]()
     
     for layer in self {
-      let layerIndex = layer.index ?? Int(layer.inFrame)
+      let layerIndex = layer.index ?? Int(layer.inFrame) // In case of using image sequencies not vector data json
       if layer.hidden == true {
         let genericLayer = NullCompositionLayer(layer: layer)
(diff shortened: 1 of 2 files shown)


lottie-swift/src/Private/LayerContainers/Utility/CompositionLayersInitializer.swift
```diff
@@ -22,18 +22,19 @@ extension Array where Element == LayerModel {
...
       } else if let shapeLayer = layer as? ShapeLayerModel {
         let shapeContainer = ShapeCompositionLayer(shapeLayer: shapeLayer)
         compositionLayers.append(shapeContainer)
-        layerMap[layer.index] = shapeContainer
+        layerMap[layerIndex] = shapeContainer
       } else if let solidLayer = layer as? SolidLayerModel {
         let solidContainer = SolidCompositionLayer(solid: solidLayer)
         compositionLayers.append(solidContainer)
-        layerMap[layer.index] = solidContainer
+        layerMap[layerIndex] = solidContainer
```
jorgisimo90, line 37: All

jorgisimo90 (approved):

calda:
Should be fixed in #1433, let me know if that didn't fix your issue

PR #1419, opened 2021-11-23
Made animationName public so it can be changed from anywhere

AngCosmin:
Hello 👋,

I think it will be great if the property `animationName` will be `public` so it can be changed from anywhere.


tiam1karimi (approved):

Sources/Public/Animation/AnimationView.swift
```diff
@@ -685,7 +685,7 @@ final public class AnimationView: LottieView {
   var animationLayer: RootAnimationLayer? = nil
 
   /// Set animation name from Interface Builder
-  @IBInspectable var animationName: String? {
```
calda, line 688: One issue with this API is that `animationName` and `animation` can become out of sync. That is:

```swift
animationView.animationName = "Animation A"
animationView.animation = .named("AnimationB")
print(animationView.animationName) // incorrectly prints "Animation A"
```

Since you can get the equivalent behavior by just setting `animation = Animation.named(...)`, I think we should leave this internal.

PR #1426, opened 2021-12-07
Fixed animation pixelation. Fixed frame in AnimationView init

Iyuna:


calda (approved):
Thanks for this fix @lyuna!

Could you also rebase to resolve the merge conflicts? We recently enabled Swift linting in CI, so for checks to pass you'll likely need to run bundle install && bundle exec rake format:swift after making code changes.

lottie-swift/src/Private/LayerContainers/CompLayers/CompositionLayer.swift
```diff
@@ -143,7 +143,7 @@ class CompositionLayer: CALayer, KeypathSearchable {
   }
   
   func updateRenderScale() {
-    /// To be overridden by subclass
```
calda, line 146: What is this change for? It looks like at least some of the subclasses set `self.contentsScale = self.renderScale` in their `override func updateRenderScale()` implementations -- is there a reason this needs to be implemented in both places?

calda: Oh interesting, is this just a revert of #1328? If so then this change makes sense

mbiparship:
Thank you so much for taking on this issue! I just gave it a try and it works very well for me.

calda:
@lyuna, thanks for bringing this to our attention. I posted a fix for this in #1436. Let me know if that works for you.

PR #1427, opened 2021-12-08
Update CI to build project and example app

calda:
This PR updates this repo's CI to build the Lottie package for iOS, and iOS example app.

They key changes are the additions to `Rakefile` and `.github/workflows/main.yml` (https://github.com/airbnb/lottie-ios/pull/1427/commits/c13e13efbabf9237c74eeceed952e72b81ced1e7)
 - we run `bundle exec rake build:package` and `bundle exec rake build:example` in CI
 - these commands use `xcodebuild` to build the corresponding scheme in the new top-level `Lottie.xcworkspace`
 
The new top-level `Lottie.xcworkspace` includes the Lottie package (referenced directly through the SPM `Package.swift`) and the Example Xcode project (`Example/Example.xcodeproj`).

Screen Shot 2021-12-07 at 4 15 07 PM


Rakefile
```diff
@@ -1,16 +1,28 @@
+namespace :build do
+  desc 'Builds the Lottie package for iOS'
+  task :package do
+    sh 'xcodebuild build -scheme Lottie -destination generic/platform=iOS'
+  end
+
+  desc 'Builds the iOS Lottie Example app'
+  task :example do
+    sh 'xcodebuild build -scheme "Example (iOS)" -destination "platform=iOS Simulator,name=iPhone 8"'
```
erichoracek, line 9: Perhaps as a follow-up, it would be nice to use `xcbeautify` here https://github.com/thii/xcbeautify which we should be able to install using Mint

erichoracek (approved):
Good stuff, LGTM with a few small suggestions.

Rakefile
calda: Great idea, will look in to this as a follow-up

PR #1428, opened 2021-12-08
Add SPM support for macOS and tvOS, build for all supported platforms in CI

### LINKED ISSUE #1361: Support multi-platform Xcode Projects (iOS and macOS) in Swift Package Manager

calda:
This PR adds Swift Package Manager support for macOS and tvOS, and updates this repo's CI to build the Lottie package and Example app for macOS and tvOS. This should resolve #1361.

This PR depends on #1427.

The gang's all here (and built in CI now!):

Screen Shot 2021-12-08 at 10 04 01 AM


Package.swift
```diff
@@ -3,13 +3,10 @@ import PackageDescription
 
 let package = Package(
   name: "Lottie",
-  // TODO: Fix SPM support for macOS, tvOS, and watchOS.
-  // https://github.com/airbnb/lottie-ios/issues/1361
-  // platforms: [.iOS("9.0"), .macOS("10.10"), tvOS("9.0"), .watchOS("2.0")],
-  platforms: [.iOS(.v9)],
+  platforms: [.iOS("9.0"), .macOS("10.10"), .tvOS("9.0")],
```
erichoracek, line 6: Perhaps as a follow-up, what do you think about starting a `CHANGELOG` in the same style as the one we have for Epoxy? Might be helpful to start capturing some of these

erichoracek (approved):
Good stuff!

Package.swift
calda: Good idea, although I'm not totally sure how we'd want to reference the existing releases (maybe something like "releases 3.2.3 and earlier are tracked [here](https://github.com/airbnb/lottie-ios/releases)")?

Just curious, what do you think are the benefits of having a `CHANGELOG.md` over, say, having release notes on the GitHub release tab (which seems to be how Lottie has handled release notes so far).

PR #1432, opened 2021-12-09
Add snapshot tests

calda:
This PR adds snapshot tests that run in CI. These tests currently give us 55% code coverage, which is a great start!


Package.swift
```diff
@@ -1,12 +1,20 @@
-// swift-tools-version:5.0
+// swift-tools-version:5.4
 import PackageDescription
 
 let package = Package(
   name: "Lottie",
-  platforms: [.iOS("9.0"), .macOS("10.10"), .tvOS("9.0")],
+  platforms: [.iOS("11.0"), .macOS("10.10"), .tvOS("11.0")],
```
calda, line 6: I had to bump the minimum platform target from iOS 9 to iOS 11, because the Snapshot Testing library requires iOS 11+. I think this is probably ok, since even iOS 11 is from 2017.

Tests/SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
...
+// MARK: - SnapshotTests
+
+class SnapshotTests: XCTestCase {
+
+  /// Snapshots all of the sample animation JSON files visible to this test target
+  func testLottieSnapshots() {
+    #if !os(iOS)
+    // We only run snapshot tests on iOS, since running snapshot tests
+    // for macOS and tvOS would triple the number of snapshot images
+    // we have to check in to the repo.
```
calda, line 17: This currently produces 10 MB of snapshot images

Package.swift
erichoracek: This seems fine to me 👍, but we should probably capture this change in release notes and also update podspecs so we have a consistent deployment target

Tests/SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
+// Created by Cal Stephens on 12/8/21.
+// Copyright © 2021 Airbnb Inc. All rights reserved.
+
+@testable import Lottie
```
erichoracek, line 4: nit: our style guide recommends splitting these out below the other imports with a newline

Tests/SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    // for macOS and tvOS would triple the number of snapshot images
+    // we have to check in to the repo.
+    XCTFail("Snapshot tests should only be ran on iOS")
+    #endif
+
+    for sampleAnimationURL in Bundle.module.urls(forResourcesWithExtension: "json", subdirectory: nil)! {
+      let sampleAnimationName = sampleAnimationURL.lastPathComponent.replacingOccurrences(of: ".json", with: "")
+
+      guard let animation = Animation.named(sampleAnimationName, bundle: .module) else {
+        continue
```
erichoracek, line 25: If we fail to decode the image should we throw an error? We wouldn't want this to fail silently I'd imagine

Tests/SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
...
+
+class SnapshotTests: XCTestCase {
+
+  /// Snapshots all of the sample animation JSON files visible to this test target
+  func testLottieSnapshots() {
+    #if !os(iOS)
+    // We only run snapshot tests on iOS, since running snapshot tests
+    // for macOS and tvOS would triple the number of snapshot images
+    // we have to check in to the repo.
+    XCTFail("Snapshot tests should only be ran on iOS")
```
erichoracek, line 18: If we'd like to prevent this function from running we should be able to throw an error which will both fail the test and prevent the below from running

Tests/SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
...
+      for percent in [0, 0.25, 0.5, 0.75, 1.0] {
+        let animationView = AnimationView(animation: animation)
+        animationView.frame.size = animation.snapshotSize
+        animationView.currentProgress = CGFloat(percent)
+
+        assertSnapshot(
+          matching: animationView,
+          as: .image,
+          named: "\(sampleAnimationName) (\(Int(percent * 100))%",
+          testName: "Sample")
```
erichoracek, line 37: Do you think we should use `"\(SnapshotTests.self).\(#function)` as the test name? not sure what the intent is here with "Sample" but it feels like we may have collisions if we add more types of snapshots later on

Tests/SnapshotTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    // If this is a landscape aspect ratio, we clamp the width
+    if width > height {
+      let newWidth = min(CGFloat(width), maxDimension)
+      let newHeight = newWidth * (CGFloat(height) / CGFloat(width))
+      return CGSize(width: newWidth, height: newHeight)
+    }
+
+    // otherwise, this is either a square or portrait aspect ratio,
+    // in which case we clamp the height
+    else {
```
erichoracek, line 60: nit: this else is unneeded as the if has an early return

erichoracek (approved):
@calda Good stuff, LGTM with a few small suggestions/comments

Tests/SnapshotTests.swift
calda: Great point, I fixed this in #1433 (https://github.com/airbnb/lottie-ios/pull/1433/files#diff-c63a7ac51a7f84e89040a1f0e17375e14f26cc41720b2e3444385d8c7416dce0R25)

Tests/SnapshotTests.swift
calda: aha good point, that sounds like a better approach

Tests/SnapshotTests.swift
calda: Sure, we can remove this. The default implementation is just `"\(#function)"`, so `Tests/__Snapshots__/Sample.Switch_States-25.png` would become `Tests/__Snapshots__/testLottieSnapshots.Switch_States-25.png`.

Tests/SnapshotTests.swift
calda: somewhat pedantically I think `if` / `else` makes sense here since the two branches are semantically like "peers" to each other (you are right that this its technically redundant though)

Tests/SnapshotTests.swift
calda: TIL, didn't know about this rule

calda: Aha, the linter was passing because it wasn't set up to run on this folder. Fixed, thanks for catching this.

PR #1434, opened 2021-12-10
Parallelize swift linters in CI

calda:
This PR updates this repo's CI config to run SwiftLint and SwiftFormat in parallel.

The lint step is currently the step that takes the longest to run -- parallelizing this should reduce the wall-clock time of running CI.

Screen Shot 2021-12-10 at 9 42 51 AM


calda:
Much better:

Screen Shot 2021-12-10 at 9 53 22 AM

Rakefile
```diff
@@ -59,10 +59,19 @@ namespace :test do
 end
 
 namespace :lint do
+  task swift: ['run:swiftlint', 'run:swiftformat']
+
   desc 'Lints swift files'
-  task :swift do
-    sh 'mint run SwiftLint lint Sources Tests Example Package.swift --config script/lint/swiftlint.yml --strict'
-    sh 'mint run SwiftFormat Sources Tests Example Package.swift --config script/lint/airbnb.swiftformat --lint'
+  namespace :run do
```
erichoracek, line 65: I think it's valid to have a namespace that's the same as a task name, so maybe we could just nest these under a swift namespace to make them more discoverable?

erichoracek (approved):
Good stuff, LGTM with one small suggestion

Rakefile
calda: aah good idea, didn't know that was allowed

PR #1435, opened 2021-12-10
Update iOS example app to support browsing all available sample animations

calda:
This PR updates the iOS example app to support browsing all available sample animations:


erichoracek (approved):
Good stuff! Love the use of Epoxy.

PR #1436, opened 2021-12-10
[#1407] Fix contentScale regression

### LINKED ISSUE #1407: Animation is pixelated
My Lottie file (animated or non-animated) that I created with the online SVG to Lottie converter (https://lottiefiles.com/svg-to-lottie/convert) appears pixelated, while the Web (and Android) version appear sharp like a vector image/animation should.

Steps to reproduce:
- Create shape in Sketch or Figma with a gradient fill
- Export to SVG
- Convert to Lottie using https://lottiefiles.com/svg-to-lottie/convert

## This issue is a:

LottieFiles for Mac v1.2.3 (27)
SVG to Lottie converter on: https://lottiefiles.com/svg-to-lottie/convert

## What Platform are you on? 

- [x] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift

## Expected Behavior 
Objects look sharp, no pixelation.

## Actual Behavior

Screen Shot 2021-10-26 at 16 28 06
Screen Shot 2021-10-26 at 16 28 18

## Code Example
In this Zip, the Sketch file and the generated SVG file are included, so you can test it out yourself.
[color-blob.zip](https://github.com/airbnb/lottie-ios/files/7418987/color-blob.zip)

## Animation JSON

```
{"v":"5.6.6","ip":0,"op":1,"fr":60,"w":208,"h":226,"layers":[{"ind":19265,"nm":"surface74460","ao":0,"ip":0,"op":60,"st":0,"ty":4,"ks":{"ty":"tr","o":{"k":100},"r":{"k":0},"p":{"k":[0,0]},"a":{"k":[0,0]},"s":{"k":[133.33,133.33]},"sk":{"k":0},"sa":{"k":0}},"shapes":[{"ty":"gr","hd":false,"nm":"surface74460","it":[{"ty":"gr","hd":false,"it":[{"ty":"sh","ks":{"k":{"i":[[0,0],[18.14,-34.81],[19.5,37.41],[-36.94,0]],"o":[[36.94,0],[-18.14,34.81],[-19.5,-37.41],[0,0]],"v":[[78.38,0],[152.25,141.75],[4.5,141.75],[78.38,0]],"c":true}}},{"ty":"gf","s":{"k":[77.7,42.31]},"e":{"k":[77.7,192.33]},"t":1,"h":{"k":0},"a":{"k":0},"g":{"p":3,"k":{"k":[0,0.65,0.39,0.39,0.4,0.81,0.64,0.29,1,0.4,0.12,0.62,0,1,0.4,1,1,1]}},"hd":false,"o":{"k":100}},{"ty":"tr","o":{"k":100},"r":{"k":0},"p":{"k":[0,0]},"a":{"k":[0,0]},"s":{"k":[100,100]},"sk":{"k":0},"sa":{"k":0},"hd":false}]},{"ty":"tr","o":{"k":100},"r":{"k":0},"p":{"k":[0,0]},"a":{"k":[0,0]},"s":{"k":[100,100]},"sk":{"k":0},"sa":{"k":0},"hd":false}]}]}],"meta":{"g":"LF SVG to Lottie"}}
```

calda:
This PR reverts #1328, which introduced some visual regressions (#1407).

Now that we have snapshot tests, I added the sample animation from #1407 to the regression suite. As expected, the tests now fail when applying the changes from #1328.

| Before | After #1328 | Diff |
| ---- | ----- | ---- |
| reference_1_A031A718-AB90-4922-9227-055A8B298024 | failure_2_EA2237E9-6B6B-454C-AEF8-FC4EC1CCA83E | difference_3_D5FECE5B-3DF7-4074-8658-EEC767FDEA57 |

Fixes #1407.


erichoracek (approved):
Looks good to me. Snapshot tests help a bunch for things like this!

calda:
The sample animation added in this PR appears to render in a slightly nondeterministic way (the edges alias seem slightly differently depending on the test environment). In https://github.com/airbnb/lottie-ios/pull/1436/commits/be7b7b68e00cb2789e413c0581dc23d2a436c6a1 I reduced the precision for this test to `0.9`, which makes the tests pass. 

I also double checked that the tests still fail if I revert https://github.com/airbnb/lottie-ios/pull/1436/commits/ba98fd92e11fc032ff6e68a35142b4ac451b7c09, so even at `0.9` precision the test case is still sensitive enough to detect that sort of issue.

absolute-heike:
Hey @calda :)

Just wanted to let you know, that IMO this is still not the optimal fix for that issue. We are now rendering everything at 3x again, which of course looks better but comes at a very high rendering performance cost.

The issue lies in the scaling behavior of ShapeRenderLayer. More precisely the path being added to the context. Scaling the whole layer after that will result in pixelated path renderings (see https://github.com/airbnb/lottie-ios/blob/master/Sources/Private/NodeRenderSystem/RenderLayers/ShapeRenderLayer.swift#L89). Refactoring that to use ShapeLayers (Like in the old ObjC version) fixed the issue. I didn't open a PR for it yet, because there was a bigger refactoring necessary and I didn't get to migrate GradientStrokes yet.

That fix has since been rolled out to millions of users to play vastly complex animations and we haven't seen any regressions compared to ObjC 2.5 up until now.

Performance is also off the charts with that build:
* ~90% Performance Boost compared to **Lottie 2.5 (ObjC)** for some animations on older devices
* >10X performance improvement compared to **Lottie 3 (Swift)**

What do you think? Should I clean that commit up a bit, open a PR (even without the GradientStrokes being migrated) and we can figure out together what needs to be done in order to get that merged?

calda:
Thanks for your thoughts and effort in this space @absolute-heike. Agree that there is lots of room for performance improvements.

My top priority for the current animation rendering implementation is to avoid any visual regressions. We recently invested in this space by setting up snapshot tests (#1432 -- before this, the repo had no automatic tests). Coverage is still a bit lower than I'd like (only ~55%), but this is a good start.

I don't personally have much knowledge or expertise about the current animation rendering implementation, but I think I'm comfortable with changes / refactors as long as the specific area being changed has good test coverage and the snapshot tests pass without any issues (speaking of snapshot tests, it would be super helpful if you could contribute your test animations to the regression suite if possible!).

On the performance side, I'm currently working on prototyping a new experimental rendering engine that leverages Core Animation more extensively (e.g. everything is a `CALayer` that animates with `CAAnimation`s). Hopefully this will drastically reduce (or even eliminate?) CPU / main thread overhead of Lottie animations (but its still extremely early, so no guarantees yet). My broad thinking here is that investing in a new engine will let us make changes and improvements with higher confidence that we aren't causing issues with the existing implementation (since the new experimental engine is behind a flag and can be enabled/disabled for each individual `AnimationView`).

buba447:
@calda Hello! Nice to meet you! I see there's a lot of activity here recently! This is awesome. I thought I would introduce myself- I'm Brandon, the author of lottie-ios. Just wanted to ping you with a couple of things:

1. I wrote a document [here](http://airbnb.io/lottie/#/ios-contributor) that outlines the entire iOS rendering pipeline. Reading through it should help clear up much of how the rendering system works, some interesting gotchas, and why some of the decisions were made. 

2. We've had snapshot testing for quite some time now! We keep it on a private repo, because the snapshot tests run on a library of animations from many top tier companies. We have to keep it private because of the IP. The animation library ensures that new builds won't break animations form some of Lotties major supports and consumers.  Im happy to migrate to the newer visual regression testing, but some thought will have to be put into how we handle private IP and animations. This part is a must have for the repo. We can talk more about this over email if you would like access to it.

 Feel free to reach out to me if you have any questions: brandon@withrow.io

calda:
Thanks @buba447! Very helpful. Glad to hear we have internal access to a pretty large regression suite.

PR #1443, opened 2021-12-14
Add barebones support for FillLayer and ShapeLayer to experimental rendering engine

calda:
This PR depends on #1442.

This PR adds barebones support for `FillLayer` and `ShapeLayer` in the experimental rendering engine. There isn't any support for actual animations yet, but this implementation is able to correctly render the first frame of some simple animations.

For example, here's the `Zoom.json` sample file that I've been testing with:

| Existing rendering engine | Experimental rendering engine |
| ----- | ------ |
| Simulator Screen Shot - iPhone 12 Pro Max - 2021-12-14 at 11 35 31 | Simulator Screen Shot - iPhone 12 Pro Max - 2021-12-14 at 11 36 03 |


calda:
The rest of these are obviously not looking so good 🤣

Tests/SnapshotTests.swift
```diff
@@ -104,6 +92,53 @@ class SnapshotTests: XCTestCase {
   /// The list of snapshot image files in `Tests/__Snapshots__`
   private let snapshotURLs = Bundle.module.urls(forResourcesWithExtension: "png", subdirectory: nil)!
 
+  /// Captures snapshots of `sampleAnimationURLs` and compares them to the snapshot images stored on disk
+  private func compareSampleSnapshots(
+    usingExperimentalRenderingEngine: Bool = false,
+    testName: String = #function) throws
+  {
+    #if os(iOS)
```
erichoracek, line 100: nit: I believe these conditions support bangs, so I think we could do:
```swift
if !os(iOS)
throw SnapshotError.unsupportedPlatform
#endif
```

erichoracek (approved):
Good stuff!

PR #1447, opened 2021-12-16
Update tests to support snapshotting in-flight CAAnimations

calda:
This PR updates our tests to support snapshotting in-flight `CAAnimation`s. This can be done by setting up the `CAAnimation` at the given progress value and then snapshotting `view.layer.presentation()`.


Sources/Private/Experimental/Extensions/KeyframeGroup+CAKeyframeAnimation.swift
```diff
@@ -60,8 +64,27 @@ extension KeyframeGroup {
...
+    //   time points between the start and end times.
+    //
+    if keyTimes.first != 0.0 {
+      keyTimes.insert(0.0, at: 0)
+      values.insert(values.first!, at: 0)
+    }
+
+    if keyTimes.last != 1.0 {
+      keyTimes.append(1.0)
+      values.append(values.last!)
```
calda, line 82: This was quite the tricky issue I ran in to -- the documentation says "the first value in the array must be 0.0 and the last value must be 1.0". 

This doesn't seems _strictly required_ at runtime (because it worked correctly at runtime in the previous PR) **BUT** the `layer.presentation()` copy seems to silently ignore any keyframes that don't follow this rule.

Tests/SnapshotConfiguration.swift
```diff
@@ -0,0 +1,64 @@
+// Created by Cal Stephens on 12/15/21.
+// Copyright © 2021 Airbnb Inc. All rights reserved.
+
+// MARK: - SnapshotConfiguration
+
+/// Snapshot configuration for an individual test case
+struct SnapshotConfiguration {
```
calda, line 7: Moved this to its own file and added more doc comments -- no real changes otherwise

Sources/Private/Experimental/Extensions/KeyframeGroup+CAKeyframeAnimation.swift
```diff
@@ -44,9 +44,13 @@ extension KeyframeGroup {
     context: LayerAnimationContext)
     -> CAKeyframeAnimation
   {
+    precondition(!keyframes.isEmpty)
```
erichoracek, line 47: Add a message to this precondition?

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -38,20 +38,28 @@ final class ExperimentalAnimationLayer: CALayer {
 
   // MARK: Internal
 
-  // TODO: Add more configuration, like `loopMode`
-  func playAnimation() {
+  /// Timing-related configuration to apply to this layer's child `CAAnimation`s
+  struct TimingConfiguration {
```
erichoracek, line 42: Is this effectively implementing `CAMediaTiming`? Wondering if we would want to make this explicit via a conformance or at least document it

Sources/Private/Experimental/Extensions/KeyframeGroup+CAKeyframeAnimation.swift
erichoracek: Interesting! Core Animation is full of so many weird things like this 😅

Sources/Private/Experimental/Layers/ShapeItemLayer.swift
```diff
@@ -49,6 +49,17 @@ final class ShapeItemLayer: CAShapeLayer {
     fatalError("init(coder:) has not been implemented")
   }
 
+  /// Called by CoreAnimation to create a shadow copy of this layer
+  /// More details: https://developer.apple.com/documentation/quartzcore/calayer/1410842-init
+  override init(layer: Any) {
+    guard let layer = layer as? Self else {
+      fatalError("init(layer:) incorrectly called with \(type(of: layer))")
```
erichoracek, line 56:
```suggestion
      fatalError("\(Self).init(layer:) incorrectly called with \(type(of: layer))")
```

Sources/Private/Experimental/Layers/SolidLayer.swift
```diff
@@ -33,6 +33,17 @@ final class SolidLayer: CALayer {
...
+  /// Called by CoreAnimation to create a shadow copy of this layer
+  /// More details: https://developer.apple.com/documentation/quartzcore/calayer/1410842-init
+  override init(layer: Any) {
+    guard let layer = layer as? Self else {
+      fatalError("init(layer:) incorrectly called with \(type(of: layer))")
+    }
+
+    solidLayer = layer.solidLayer
+    super.init(layer: layer)
+  }
```
erichoracek, line 45: Seems like we're repeating this logic a number of times, wondering if we could generalize?

Tests/Snapshotting+presentationLayer.swift
```diff
@@ -0,0 +1,41 @@
...
+        window.addSubview(view)
+
+        // Calling CATransaction.flush() makes Core Animation synchronously update
+        // the content being displayed on-screen, which gives our view a presentation layer.
+        CATransaction.flush()
+
+        // Now that the view is on-screen, it has a presentation layer:
+        guard let presentationLayer = view.layer.presentation() else {
+          fatalError("Presentation layer does not exist and cannot be snapshot")
+        }
```
erichoracek, line 32: Is it possible to throw here so the test fails rather than crashing? Or does `pullback` not rethrow?

Tests/Snapshotting+presentationLayer.swift
```diff
@@ -0,0 +1,41 @@
...
+    precision: Float = 1,
+    traits: UITraitCollection = .init())
+    -> Snapshotting<UIView, UIImage>
+  {
+    SimplySnapshotting
+      .image(precision: precision, scale: traits.displayScale)
+      .pullback { (view: UIView) -> UIImage in
+        // Place the view in an on-screen window and then
+        // have Core Animation update the screen synchronously
+        let window = UIWindow()
```
erichoracek, line 21: Should we give the window a bounds equal to the view.bounds? Seems like a size-zero window could cause some weirdness

Tests/Snapshotting+presentationLayer.swift
```diff
@@ -0,0 +1,41 @@
...
+  /// Captures an image of the input `UIView`'s `layer.presentation()`,
+  /// which reflects the current state of any in-flight animations.
+  static func imageOfPresentationLayer(
+    precision: Float = 1,
+    traits: UITraitCollection = .init())
+    -> Snapshotting<UIView, UIImage>
+  {
+    SimplySnapshotting
+      .image(precision: precision, scale: traits.displayScale)
+      .pullback { (view: UIView) -> UIImage in
```
erichoracek, line 18: It's not entirely clear what pullback does from reading this—I assume it overrides the logic that goes from view to image but then does that augment the behavior from `.image` above? Might be helpful to explain exactly what's happening here

Tests/SnapshotTests.swift
```diff
@@ -140,11 +108,22 @@ class SnapshotTests: XCTestCase {
             useNewRenderingEngine: usingExperimentalRenderingEngine))
 
         animationView.frame.size = animation.snapshotSize
-        animationView.currentProgress = CGFloat(percent)
+        animationView.layoutIfNeeded()
```
erichoracek, line 111: Add a comment explaining what this layout accomplishes?

erichoracek (approved):
LGTM! Just a few small questions/suggestions. Lots of subtlety but glad you were able to get this working 🎉

Sources/Private/Experimental/Layers/SolidLayer.swift
calda: Yeah I'm really not a fan this layer shadowing API -- the requirements that you have to `override init(layer: Any)` and must call `super.init(layer: layer)` (or else things silently break) makes this tricky to generalize. 

I tried some approaches to generalize this in a way that individual layer subclasses only have to implement a method like:

```swift
func copy(from self: Self) { ... }
```

where the casting is handled automatically by the base implementation

but:
 1. this can't me an instance method, because it has to run before calling `super.init`
 2. this can't be a static copying method of the shape `(Self) -> Self`, because then you can't call `super.init(layer: Any)`.
 3. you also can't implement an `override init` from a protocol extension, so there's no feasible way to provide a default implementation of the init afaict
 
We could potentially extract out the `layer as? Self` casting, but its already pretty lighweight and I couldn't come up with a spelling that felt good enough to be better than a standard `guard`.

Any thoughts or ideas? I would certainly love to improve this if we can think of something.

Tests/Snapshotting+presentationLayer.swift
calda: haha you and me both 😆 

I was following the implementation of the framework's existing `UIView` -> `UIImage` snapshotting implementation, which is written like this:

```swift
return SimplySnapshotting.image(precision: precision, scale: traits.displayScale).asyncPullback { view in
  // ...
])
 ```

Tests/Snapshotting+presentationLayer.swift
calda: `pullback` is not `rethrows` so we have to either `fatalError` or return an empty `UIImage`

an empty `UIImage` would probably be better here, since it would show up as a diff failure for the specific problematic test case and not block everything else from continuing

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
calda: interesting idea -- I don't think all of the `CAMediaTiming` properties are applicable here (e.g. `duration`) but mentioning it in the doc comment can't hurt

Sources/Private/Experimental/Extensions/KeyframeGroup+CAKeyframeAnimation.swift
calda: image

Sources/Private/Experimental/Layers/SolidLayer.swift
erichoracek: If you investigated and there weren't any clean options to factor out seems totally fine to me to keep as is 👍

PR #1449, opened 2021-12-17
Add interactive scrubbing support to experimental rendering engine

calda:
This PR adds interactive scrubbing support to the experimental rendering engine. This works by setting the root layer's `speed` to 0, and then applying a `timeOffset` to each of the `CAAnimation`s.

I also simplified how the animation setup works, now that we have a better idea of the requirements.

### DIFF
--- a/Sources/Private/Experimental/ExperimentalAnimationLayer.swift
+++ b/Sources/Private/Experimental/ExperimentalAnimationLayer.swift
@@ -47,20 +47,41 @@ final class ExperimentalAnimationLayer: CALayer {
     var timeOffset: TimeInterval = 0
   }
 
-  func playAnimation(timingConfiguration: TimingConfiguration) {
+  /// The timing configuration that is being used for the currently-active animation
+  private var timingConfiguration: TimingConfiguration?
+
+  /// Sets up `CAAnimation`s for each `AnimationLayer` in the layer hierarchy
+  func setupAnimation(timingConfiguration: TimingConfiguration) {
+    self.timingConfiguration = timingConfiguration
+
     let context = LayerAnimationContext(
       startFrame: animation.startFrame,
       endFrame: animation.endFrame,
       framerate: CGFloat(animation.framerate))
 
+    // Remove any existing animations from the layer hierarchy
     for animationLayer in animationLayers {
+      animationLayer.removeAllAnimations()
+    }
+
+    // Set up the new animations with the current `TimingConfiguration`
+    for animationLayer in animationLayers {
+
+      // Necessary here and not on animation (why??)
+      animationLayer.speed = timingConfiguration.speed
+
+      // TODO: instead of collecting and applying the animations here,
+      // it would be nice to just build them and apply them in the subclass
+      // (passing `TimingConfiguration` through the context struct)
+
       for caAnimation in animationLayer.animations(context: context) {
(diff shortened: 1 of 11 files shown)


Sources/Private/Experimental/Layers/ShapeLayer.swift
```diff
@@ -14,19 +14,6 @@ final class ShapeLayer: CALayer {
     self.shapeLayer = shapeLayer
     super.init()
 
-    // TODO: On the model side, these fields are part of a `LayerModel` superclass.
-    // We should put this somewhere where it doesn't have to be duplicated by
-    // all layer types.
-    transform = CATransform3D.makeTransform(
-      anchor: shapeLayer.transform.anchorPoint.keyframes.first!.value.pointValue,
```
calda, line 21: be gone!

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -103,9 +122,15 @@ final class ExperimentalAnimationLayer: CALayer {
 
 extension ExperimentalAnimationLayer: RootAnimationLayer {
 
-  var currentFrame: CGFloat {
-    get { 0 }
-    set { /* Currently unsupported */ }
+  var currentFrame: AnimationFrameTime {
+    get {
+      0 // TODO: how do we retrieve the realtime animation progress?
```
erichoracek, line 127: I assume `presentation().timeOffset` doesn't work here?

calda: I tried that and it was always `0`

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -166,3 +191,19 @@ extension ExperimentalAnimationLayer: RootAnimationLayer {
...
+// MARK: - CALayer + allSublayers
+
+extension CALayer {
+  /// All of the layers in the layer tree that are descendants from this later
+  @nonobjc
+  fileprivate var allSublayers: [CALayer] {
+    var allSublayers: [CALayer] = []
+
+    for sublayer in sublayers ?? [] {
+      allSublayers += [sublayer] + sublayer.allSublayers
```
erichoracek, line 204: nit: seems like this will create less throwaway arrays
```suggestion
      allSublayers.append(sublayer)
      allSublayers.append(contentsOf: sublayer.allSublayers)
```

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
calda: It seems like the idiomatic way to track realtime animation progress with Core Animation is to create a placeholder `CABasicAnimation` that animates a `"progress"` key or something like that ([discussion on stack overflow](https://stackoverflow.com/questions/18827973/core-animation-progress-callback/19024971#19024971), which is basically exactly how this works in the existing animation engine)

erichoracek (approved):

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
calda: good point, hadn't thought of that

PR #1451, opened 2021-12-17
Support loopMode and accessing realtime animation progress

calda:
This PR adds support for configuring the `loopMode` of the new rendering engine, and accessing the realtime animation progress

### DIFF
--- a/Example/iOS/AnimationPreviewViewController.swift
+++ b/Example/iOS/AnimationPreviewViewController.swift
@@ -119,7 +119,7 @@ class AnimationPreviewViewController: UIViewController {
   }
 
   private func updateLoopMode(to loopMode: LottieLoopMode) {
-    animationView.loopMode = .playOnce
+    animationView.loopMode = loopMode
     animationView.play()
     configureSettingsMenu()
   }
--- a/Example/iOS/AnimationPreviewViewController.swift
+++ b/Example/iOS/AnimationPreviewViewController.swift
@@ -101,27 +101,27 @@ class AnimationPreviewViewController: UIViewController {
             title: "Autoreverse",
             state: animationView.loopMode == .autoReverse ? .on : .off,
             handler: { [weak self] _ in
-              self?.animationView.loopMode = .autoReverse
-              self?.animationView.play()
-              self?.configureSettingsMenu()
+              self?.updateLoopMode(to: .autoReverse)
             }),
           UIAction(
             title: "Loop",
             state: animationView.loopMode == .loop ? .on : .off,
             handler: { [weak self] _ in
-              self?.animationView.loopMode = .loop
-              self?.animationView.play()
-              self?.configureSettingsMenu()
+              self?.updateLoopMode(to: .loop)
             }),
           UIAction(
             title: "Play Once",
             state: animationView.loopMode == .playOnce ? .on : .off,
             handler: { [weak self] _ in
-              self?.animationView.loopMode = .playOnce
-              self?.animationView.play()
-              self?.configureSettingsMenu()
+              self?.updateLoopMode(to: .playOnce)
             }),
         ]))
+  }
+
+  private func updateLoopMode(to loopMode: LottieLoopMode) {
+    animationView.loopMode = .playOnce
+    animationView.play()
+    configureSettingsMenu()
   }
 
 }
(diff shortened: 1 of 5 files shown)


Example/iOS/AnimationPreviewViewController.swift
```diff
@@ -90,4 +89,39 @@ class AnimationPreviewViewController: UIViewController {
...
+      menu: UIMenu(
+        title: "Loop Mode",
+        children: [
+          UIAction(
+            title: "Autoreverse",
+            state: animationView.loopMode == .autoReverse ? .on : .off,
+            handler: { [weak self] _ in
+              self?.animationView.loopMode = .autoReverse
+              self?.animationView.play()
+              self?.configureSettingsMenu()
```
erichoracek, line 106: nit: factor out into a closure that takes a param above?

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -75,6 +80,13 @@ final class ExperimentalAnimationLayer: CALayer {
     for animationLayer in animationLayers {
       animationLayer.setupAnimations(context: context)
     }
+
+    // We also set up a placeholder animation to track the current animation progress
+    let animationProgressTracker = CABasicAnimation(keyPath: #keyPath(animationProgress))
+    animationProgressTracker.configureTiming(with: context)
+    animationProgressTracker.fromValue = 0
+    animationProgressTracker.toValue = 1
+    add(animationProgressTracker, forKey: key)
```
erichoracek, line 89: Is this going to add some overhead to these animations as this will ping the main thread at the device frame rate? Wonder if we could do something wild like have a size 0 layer animating its opacity and then getting its `presentationLayer().opacity`?

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -75,6 +80,13 @@ final class ExperimentalAnimationLayer: CALayer {
     for animationLayer in animationLayers {
       animationLayer.setupAnimations(context: context)
     }
+
+    // We also set up a placeholder animation to track the current animation progress
+    let animationProgressTracker = CABasicAnimation(keyPath: #keyPath(animationProgress))
+    animationProgressTracker.configureTiming(with: context)
```
erichoracek, line 86: nit: factor out into a method?

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
erichoracek, line 89: Why do we need to pass this key through from the caller vs. having a static key? Might be helpful to document

erichoracek (approved):
LGTM! Just a few small questions

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
calda: Good question, since this doesn't have `needsDisplay: true` I assume it doesn't actually run on the main thread every frame. Let me double-check this though.

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
calda: good idea

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
calda: `AnimationView.isAnimationPlaying` is implemented by checking `animationLayer?.animation(forKey: activeAnimationName) != nil`, where `activeAnimationName` is the `key` of the placeholder `CABasicAnimation` tracking the animation progress. My original thinking was that following this pattern with the `ExperimentalAnimationLayer` (and having both layers behave in the same way) would simplify things, but I think this is a pretty confusing / non-obvious way to do this so will probably change it.

calda: When I add a `didSet` tp `animationProgress` and set a breakpoint, it doesn't get called while the animation is playing. CPU usage is also stays at 0% while playing the animation.

erichoracek: Nice! Great to hear

PR #1452, opened 2021-12-17
Set up correct timing functions for CAKeyframeAnimations

calda:
This PR updates our `CAKeyframeAnimation` setup to use the correct `timingFunctions`, as specified by `Keyframe.inTangent` and `Keyframe.outTangent`

| Before (completely linear, gross!) | After |
| ---- | ----- |
|  |  |


erichoracek (approved):
Nice!

PR #1453, opened 2021-12-20
Fix Carthage support

### LINKED ISSUE #1448: Dependency "lottie-ios" has no shared framework schemes using Carthage
## This issue is a:
- Can't Install Lottie with Carthage 

Lottie 3.3.0
github "airbnb/lottie-ios" "master"

## What Platform are you on? 

-  iOS

## What Language are you in?
- Swift

## Error 
https://github.com/airbnb/lottie-ios/tree/42dc2f40bbb7cbcff8866b369beea62683a50f36
Screenshot 2021-12-17 at 3 16 51 PM 

https://github.com/airbnb/lottie-ios/tree/cac9ede96fc134a2f96f923fdbb0cfa80246215a
Screenshot 2021-12-17 at 3 25 25 PM

**Kindly fix these issues ASAP.**

calda:
This PR fixes Carthage support, which I broke in #1427 when I restructured the project to be primarily a SPM package.

I also added a `bundle exec rake test:carthage` command that will run in CI from now on, which validates that Carthage can build the package correctly.

The key problem was that Carthage support requires a shared framework scheme, which is not created by SPM. To support this, I set up a `LottieCarthage` framework scheme that `@_exported import`s the `Lottie` SPM package. This causes the Carthage build to succeed again, but has the unfortunate side-effect of requiring Carthage consumers to `import LottieCarthage` instead of `import Lottie`.

Resolves #1448.


calda:
Admin merging since folks are out for the holidays

PR #1454, opened 2021-12-21
Rename Carthage framework from "LottieCarthage" to just "Lottie"

calda:
This PR renames the Carthage framework from `LottieCarthage` (added in #1453) to just `Lottie`.

The Carthage framework `@_exported import`s the SPM implementation target, which has to have a different name from the Carthage framework itself. (This was why the Carthage framework was previously called `LottieCarthage`).

For this to work, this PR renames the existing SPM `Lottie` implementation target to `LottieCore`. Similarly to the `Lottie` Carthage framework, we now define a `Lottie` SPM target that `@_exported import`s `LottieCore`.


script/test-carthage/CarthageTest/ViewController.swift
```diff
@@ -2,7 +2,7 @@
 // Copyright © 2021 Airbnb Inc. All rights reserved.
 
 import UIKit
-import LottieCarthage
+import Lottie
```
calda, line 5: Much better!

calda:
Admin merging since folks are out for the holidays

PR #1456, opened 2021-12-21
Support more ShapeLayer / LayerModel features (Trim, Rectangle, inFrame, outFrame)

calda:
This PR adds more `ShapeLayer` / `LayerModel` features (each in a separate commit):
 - Support for `Trim` (maps to `CALayer.strokeStart` / `CALayer.strokeEnd`) 
 - Support for `Rectangle` `ShapeItem`s
 - Support for top-level `ShapeItem`s in `ShapeLayer` (this previously only supported top-level `Group`s)
 - Support for rendering multiple `CGPath`s in a single `ShapeLayer`
 - Support for `LayerModel.inFrame` / `LayerModel.outFrame` (which toggle layer visibility)
 
This PR depends on #1455.
 
A bunch of animations are working pretty nicely now:

| Public | Private |
| ----- | ----- |
|  |  |


Sources/Private/Experimental/Animations/RectangleAnimation.swift
```diff
@@ -0,0 +1,153 @@
...
+  /// Adds animations for the given `Rectangle` to this `CALayer`
+  func addAnimations(
+    for rectangle: Rectangle,
+    context: LayerAnimationContext)
+  {
+    addAnimation(
+      for: .path,
+      keyframes: rectangle.size.keyframes,
+      value: { sizeKeyframe in
+        let size = sizeKeyframe.sizeValue * 0.5
```
calda, line 16: Pulled this `CGPath` rendering implementation from https://github.com/airbnb/lottie-ios/blob/master/Sources/Private/NodeRenderSystem/Nodes/PathNodes/RectNode.swift#L83

Sources/Private/Experimental/Animations/VisibilityAnimation.swift
```diff
@@ -0,0 +1,37 @@
...
+import QuartzCore
+
+extension CALayer {
+  /// Adds an animation for the given `inTime` and `outTime` to this `CALayer`
+  func addVisibilityAnimation(
+    inFrame: AnimationFrameTime,
+    outFrame: AnimationFrameTime,
+    context: LayerAnimationContext)
+  {
+    let animation = CAKeyframeAnimation(keyPath: "hidden")
```
erichoracek, line 13: I'm pretty sure that `#keyPath(CALayer.isHidden)` resolves to "hidden" if you want to use it here
```suggestion
    let animation = CAKeyframeAnimation(keyPath: #keyPath(CALayer.isHidden))
```

Sources/Private/Experimental/Animations/VisibilityAnimation.swift
```diff
@@ -0,0 +1,37 @@
...
+    //    there should be three key times.
+    animation.keyTimes = [
+      NSNumber(value: 0.0),
+      NSNumber(value: Double(context.relativeTime(of: inFrame))),
+      NSNumber(value: Double(context.relativeTime(of: outFrame))),
+      NSNumber(value: 1.0),
+    ]
+
+    animation.configureTiming(with: context)
+    add(animation, forKey: "hidden")
```
erichoracek, line 35:
```suggestion
    add(animation, forKey: #keyPath(CALayer.isHidden))
```

Sources/Private/Experimental/Animations/RectangleAnimation.swift
erichoracek: Do you think it makes sense to try to keep these in sync/shared? Or at least add a comment to the effect that we're trying to keep them in sync? Not sure how long you'd imagine we'll have both

erichoracek (approved):
Good stuff! Just a few small suggestions/questions

Sources/Private/Experimental/Layers/ShapeItemLayer.swift
```diff
@@ -111,6 +141,28 @@ extension ShapeItemLayer: AnimationLayer {
...
+
+  private func addAnimations(for trim: Trim, context: LayerAnimationContext) {
+    addAnimation(
+      for: .strokeStart,
+      keyframes: trim.start.keyframes,
+      value: { strokeStart in
+        // Lottie animation files express stoke trims as a numerical percentage value
+        // (e.g. 25%, 50%, 100%) so we divide by 100 to get the decimal values
+        // expected by Core Animation (e.g. 0.25, 0.5, 1.0).
+        CGFloat(strokeStart.cgFloatValue) / 100
```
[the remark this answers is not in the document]
calda: Interesting idea -- one thought here is that the code itself is already very lightweight imo (just a `/ 100`), which makes me think that we may not get much out of having a shared utility for this.

Sources/Private/Experimental/Animations/RectangleAnimation.swift
calda: Good idea, seems reasonable to keep this shared between the two since it's rather large and unlikely to change much

Sources/Private/Experimental/Animations/VisibilityAnimation.swift
calda: You're right! Good thinking

```
(lldb) p #keyPath(CALayer.isHidden)
(String) $R1 = "hidden"
```

Sources/Private/Experimental/Layers/ShapeItemLayer.swift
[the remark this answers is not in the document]
erichoracek: Yeah, I'm mostly thinking about the "idea" of combining percentage normalization so that this comment/logic doesn't need to be sprinkled around everywhere. Just an idea of course, no obligation to take this approach.

PR #1462, opened 2022-01-04
Add demo of AnimatedSwitch and AnimatedButton to Example app

calda:
This PR adds a demo of `AnimatedSwitch` and `AnimatedButton` to the Example app. These don't currently work correctly with the experimental rendering engine -- will work on fixes in a follow-up.


Example/iOS/ViewControllers/ControlsDemoViewController.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    super.init(layout: layout)
+    setItems(items, animated: false)
+
+    title = "Controls Demo"
+  }
+
+  // MARK: Internal
+
+  var items: [ItemModeling] {
```
erichoracek, line 25: nit: if you'd like you can make this an `@ItemModelBuilder` to omit the commas and array literal

Example/iOS/Views/AnimatedButtonRow.swift
```diff
@@ -0,0 +1,111 @@
...
+      else { return false }
+
+      for (lhsRanges, rhsRanges) in zip(lhs.playRanges, rhs.playRanges) {
+        if lhsRanges != rhsRanges {
+          return false
+        }
+      }
+
+      return true
+    }
```
erichoracek, line 47: nit: making `playRanges` a nested `Equatable` struct would probably be shorter than this implementation

Example/iOS/Views/AnimatedSwitchRow.swift
```diff
@@ -0,0 +1,119 @@
...
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  // MARK: Internal
+
+  struct Content: Equatable {
+    var animationName: String
+    var title: String
+    var onTimeRange: (from: CGFloat, to: CGFloat)
+    var offTimeRange: (from: CGFloat, to: CGFloat)
```
erichoracek, line 33: How about using `ClosedRange<CGFloat>` instead? Will that also allow us to remove our equatable implementation?

Example/iOS/ViewControllers/SampleListViewController.swift
```diff
@@ -83,13 +83,28 @@ final class SampleListViewController: CollectionViewController {
...
+    if isTopLevel {
+      items += [
+        LinkView.itemModel(
+          dataID: "Controls Demo",
+          content: .init(animationName: nil, title: "Controls Demo"))
+          .didSelect { [weak self] context in
+            self?.show(ControlsDemoViewController(), sender: context.view)
+          },
+      ]
+    }
```
erichoracek, line 95: nit: Suggest breaking this out into multiple methods that return arrays of items models for readability

erichoracek (approved):
LGTM! Just a few small nitpicks

Example/iOS/Views/AnimatedButtonRow.swift
calda: good idea

Example/iOS/Views/AnimatedSwitchRow.swift
calda: Good idea!

Example/iOS/ViewControllers/SampleListViewController.swift
calda: Good suggestion, much better now https://github.com/airbnb/lottie-ios/pull/1462/commits/48f98df7802f3b9a5693275bb06088815cc00968

PR #1463, opened 2022-01-05
Support animating custom time ranges

calda:
This PR adds support for animating custom time ranges in `ExperimentalAnimationLayer` (it was previously hardcoded to always play from 0% to 100%).

Now `AnimatedControl` and `AnimatedButton` almost completely work:

| Custom time ranges | Controls demo |
| ----- | ----- |
|  |  |

The only lingering issue is with interrupting in-flight animations -- the standard engine resumes the new animation from where the previous animation left off, but the experimental engine always starts from the beginning. Will work on this in a follow-up.

### DIFF
--- a/Sources/Private/Experimental/Layers/AnimationLayer.swift
+++ b/Sources/Private/Experimental/Layers/AnimationLayer.swift
@@ -48,6 +48,7 @@ extension CAAnimation {
     //
     let baseAnimation = self
     baseAnimation.duration = context.animation.duration
+    baseAnimation.speed = (context.endFrame < context.startFrame) ? -1 : 1
 
     // To select the subrange of the `baseAnimation` that should be played,
     // we create a parent animation with the duration of that subrange
@@ -66,7 +67,7 @@ extension CAAnimation {
     let clippingParent = CAAnimationGroup()
     clippingParent.animations = [baseAnimation]
 
-    clippingParent.duration = context.animation.time(forFrame: context.endFrame - context.startFrame)
+    clippingParent.duration = abs(context.animation.time(forFrame: context.endFrame - context.startFrame))
     baseAnimation.timeOffset = context.animation.time(forFrame: context.startFrame)
 
     clippingParent.autoreverses = context.timingConfiguration.autoreverses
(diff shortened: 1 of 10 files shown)


Sources/Private/Experimental/Layers/AnimationLayer.swift
```diff
@@ -17,36 +17,67 @@ protocol AnimationLayer: CALayer {
...
+    //        │   clippingParent   │  ...
+    //        └────────────────────┴───────►
+    //       25%                  75%
+    //  ┌──────────────────────────────────┐
+    //  │           baseAnimation          │
+    //  └──────────────────────────────────┘
+    //  0%                                100%
+    //
+    let clippingParent = CAAnimationGroup()
+    clippingParent.animations = [baseAnimation]
```
erichoracek, line 68: Do we need this clipping parent in the scenario where the full animation is playing? Or could we return `baseAnimation` above if the clipping parent isn't needed?

Sources/Private/Experimental/Layers/AnimationLayer.swift
```diff
@@ -17,36 +17,67 @@ protocol AnimationLayer: CALayer {
...
+    // and/or autoreverse over the clipped subrange.
+    //
+    //        ┌────────────────────┬───────►
+    //        │   clippingParent   │  ...
+    //        └────────────────────┴───────►
+    //       25%                  75%
+    //  ┌──────────────────────────────────┐
+    //  │           baseAnimation          │
+    //  └──────────────────────────────────┘
+    //  0%                                100%
```
erichoracek, line 65: Very helpful diagram!

erichoracek (approved):
Good stuff!

Sources/Private/Experimental/Layers/AnimationLayer.swift
calda: We don't strictly need the parent `CAAnimationGroup` when playing the full animation, but I think having just one codepath makes this setup simpler and easier to reason about. (I don't expect there to be an observable perf difference between the two, since this already has 0 CPU overhead).

PR #1465, opened 2022-01-06
Play new animations from current frame of existing animation

calda:
This PR updates `AnimationView`'s integration with `ExperimentalAnimationLayer` to play new animations from the `currentFrame` of the existing animation. This matches the behavior of the existing rendering engine, and fixes support for interruptible animations in `AnimatedSwitch`.

| Different `LottieLoopMode`s | Interruptible `AnimatedSwitch` animations | 
| ----- | ----- |
|  |  |

### DIFF
--- a/Sources/Private/Experimental/Animations/CAAnimation+TimingConfiguration.swift
+++ b/Sources/Private/Experimental/Animations/CAAnimation+TimingConfiguration.swift
@@ -1,0 +1,73 @@
+// Created by Cal Stephens on 1/6/22.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+import QuartzCore
+
+// MARK: - LayerAnimationContext
+
+// Context describing the timing parameters of the current animation
+struct LayerAnimationContext {
+  /// The animation being played
+  let animation: Animation
+
+  /// The timing configuration that should be applied to `CAAnimation`s
+  let timingConfiguration: ExperimentalAnimationLayer.CAMediaTimingConfiguration
+
+  /// The absolute frame number that this animation begins at
+  let startFrame: AnimationFrameTime
+
+  /// The absolute frame number that this animation ends at
+  let endFrame: AnimationFrameTime
+}
+
+// MARK: - CAAnimation + LayerAnimationContext
+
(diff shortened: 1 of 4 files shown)


Sources/Private/Experimental/Animations/CAAnimation+TimingConfiguration.swift
```diff
@@ -0,0 +1,73 @@
...
+  /// The absolute frame number that this animation begins at
+  let startFrame: AnimationFrameTime
+
+  /// The absolute frame number that this animation ends at
+  let endFrame: AnimationFrameTime
+}
+
+// MARK: - CAAnimation + LayerAnimationContext
+
+extension CAAnimation {
```
calda, line 25: Moved this from `Experimental/Layers/AnimationLayer.swift` to `Experimental/Animations/CAAnimation+TimingConfiguration.swift` -- otherwise no changes

Sources/Public/Animation/AnimationView.swift
```diff
@@ -1001,16 +1001,38 @@ final public class AnimationView: LottieView {
...
+        // duration (so the animation stops playing at `playFrom`).
+        case .playOnce:
+          animationContext.playFrom = currentFrame
+
+        // When looping, we specifically _don't_ want to affect the duration of the animation,
+        // since that would affect the duration of all subsequent loops. We just want to adjust
+        // the duration of the _first_ loop. Instead of setting `playFrom`, we just add a `timeOffset`
+        // so the first loop begins at `currentTime` but all subsequent loops are the standard duration.
+        default:
+          timingConfiguration.timeOffset = currentTime - animation.time(forFrame: animationContext.playFrom)
```
calda, line 1028: A silly issue I was running in to at first: this `timeOffset` codepath works perfectly for any animation mode that loops, but doesn't work for `.playOnce`. If you had a `playOnce` from 25% to 75% where `currentTime` was 50%, then it would play from 50% to 75% and then wrap around to play from 25% to 50% (since it was playing for the "full duration" specified by `endTime - startTime`).

erichoracek (approved):

PR #1468, opened 2022-01-07
Adds support for layer masks, fix some misc layer visibility issues

calda:
This PR adds support for layer masks, and fixes some other miscellaneous issues with layer visibility / ordering


Sources/Private/Experimental/Layers/ShapeLayer.swift
```diff
@@ -19,7 +19,11 @@ final class ShapeLayer: BaseCompositionLayer {
     let groupItems = shapeLayer.items.compactMap { $0 as? Group }
     let otherItems = shapeLayer.items.filter { !($0 is Group) }
 
-    for group in groupItems {
+    // Groups are listed from front to back,
+    // but `CALayer.sublayers` are listed from back to front.
+    let groupsInZAxisOrder = groupItems.reversed()
```
calda, line 24: He's alive!

image

calda: This change also fixes issue with `IconTransitions`: 

image

Sources/Private/Experimental/Layers/BaseCompositionLayer.swift
```diff
@@ -62,18 +64,18 @@ class BaseCompositionLayer: CALayer, AnimationLayer {
...
     transformLayer.addTransformAnimations(for: baseLayerModel.transform, context: context)
 
-    // Add the opacity animation to _only_ the `contentsLayer`
-    //  - The opacity should affect the content rendered directly by this layer,
+    // Add the opacity and visibility animations to _only_ the `contentsLayer`
+    //  - These animations should affect the content rendered directly by this layer,
     //    but _not_ any child `LayerModel`s.
     contentsLayer.addOpacityAnimation(from: baseLayerModel.transform, context: context)
 
+    contentsLayer.addVisibilityAnimation(
```
calda, line 74: This change fixes an issue where layers could disappear unexpectedly (because one of their parents were hidden)

image

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -215,6 +215,13 @@ extension ExperimentalAnimationLayer: RootAnimationLayer {
       animation.frameTime(forProgress: (presentation() ?? self).animationProgress)
     }
     set {
+      // Currently, setting `currentFrame` causes the existing layer hierarchy
+      // to be discarded and then completely rebuilt.
+      // This isn't a problem when setting the animation up just once,
+      // but can max out the CPU (causing frame drops / lag) for very-large animations
+      // when interactively controlling the current frame with a user gesture.
+      // TODO: This should be adjusted to use the existing layer hierarchy
+      //       and control something like `layer.timeOffset`.
```
calda, line 224: Left a note here about performance issues when interactively updating `currentFrame`: when bound to a gesture it rebuilds the entire layer hierarchy on every frame, which becomes a perf problem for large enough animations (this doesn't happen when playing an animation noninteractively, since we construct the layer hierarchy once as then Core Animation takes over).

Should be reasonably straightforward to fix some sort of `layer.timeOffset` approach, but will leave this as-is for now

Sources/Private/Experimental/Layers/MaskCompositionLayer.swift
```diff
@@ -0,0 +1,104 @@
...
+
+extension MaskCompositionLayer {
+  final class MaskLayer: CAShapeLayer {
+
+    // MARK: Lifecycle
+
+    init(mask: Mask) {
+      maskModel = mask
+      super.init()
+      fillColor = "#000000".cgColor
```
erichoracek, line 73: nit: why not `UIColor.black.cgColor`? Seems like we'll pay for string processing overhead here
```suggestion
      fillColor = UIColor.black.cgColor
```

Sources/Private/Experimental/Layers/MaskCompositionLayer.swift
```diff
@@ -0,0 +1,104 @@
...
+
+// MARK: - MaskCompositionLayer
+
+/// The CALayer type responsible for rendering the `Mask` of a `BaseCompositionLayer`
+final class MaskCompositionLayer: CALayer {
+
+  // MARK: Lifecycle
+
+  init(masks: [Mask]) {
+    maskLayers = masks.map { MaskLayer(mask: $0) }
```
erichoracek, line 14: nit:
```suggestion
    maskLayers = masks.map(MaskLayer.init(mask:))
```

erichoracek (approved):
Exciting!

Sources/Private/Experimental/Layers/MaskCompositionLayer.swift
calda: `CGColor.black` is only defined on macOS :(

calda: I'll tweak this to not use a string though, good call

erichoracek: Ah interesting! Yes then perhaps `UIColor.black.cgColor`

calda: We can't import `UIKit` / `UIColor` here because this file is also used on macOS (non-Catalyst) 

`CGColor` initializers are truly a wild-west, there's a nice `CGColor(red:green:blue:alpha:)` initializer that was added in iOS 13 but we're using a minimum deployment target of iOS 11. 

The pre-iOS 13 initializer isn't too bad, I just don't love how you have to force unwrap it 🤷🏻 

```swift
CGColor(
  colorSpace: CGColorSpaceCreateDeviceRGB(),
  components: [0, 0, 0])!
```

may write a "polyfill" init like we're currently doing in the `String.cgColor` extension:

```swift
    if #available(iOS 13.0, tvOS 13.0, *) {
      return CGColor(red: red, green: green, blue: blue, alpha: 1)
    } else {
      return CGColor(
        colorSpace: CGColorSpaceCreateDeviceRGB(),
        components: [red, green, blue])!
    }
```

calda: i cry 😆 😥

image

calda: Fixed: https://github.com/airbnb/lottie-ios/pull/1468/commits/c981d5e601fefc76bdca25bdac784963db3d26dd

PR #1471, opened 2022-01-11
Support gradients, images, and polygons

calda:
This PR adds support for gradients, images, and polygons to the experimental rendering engine

| Gradients (banding artifacts are just from the gif recording) | Images | More shape types, and more gradients! |
| ----- | ----- | ----- |
|  |  |  |

Gradients are easily two orders of magnitude more performant now:

| Legacy rendering engine | Experimental rendering engine |
| ---- | ---- |
| Screen Shot 2022-01-10 at 4 05 58 PM | Screen Shot 2022-01-10 at 4 06 34 PM |


Sources/Private/Experimental/Animations/CustomPathAnimation.swift
```diff
@@ -0,0 +1,20 @@
+// Created by Cal Stephens on 12/21/21.
+// Copyright © 2021 Airbnb Inc. All rights reserved.
+
+import QuartzCore
+
+extension CAShapeLayer {
```
calda, line 6: Just renamed this file

Sources/Private/Experimental/Animations/ShapeAnimation.swift
```diff
@@ -1,28 +1,81 @@
...
+    case let star as Star:
+      addAnimations(for: star, context: context)
+
+    default:
+      return
+    }
+  }
+
+  /// Adds a `fillColor` animation for the given `Fill` object
+  func addAnimations(for fill: Fill, context: LayerAnimationContext) {
```
calda, line 28: Moved these `ShapeItem`-related animations to `ShapeAnimation.swift`, from `ShapeItemLayer.swift`

Sources/Private/Experimental/Layers/GradientRenderLayer.swift
```diff
@@ -0,0 +1,86 @@
...
+///  - The gradient itself is supposed to be rendered infinitely in all directions
+///    (e.g. including outside of `bounds`). This is because `ShapeItemLayer` paths
+///    don't necessarily sit within the layer's `bounds`.
+///
+///  - To support this, `GradientRenderLayer` tracks a `gradientRenderBounds`
+///    that `startPoint` / `endPoint` are calculated relative to.
+///    The _actual_ `bounds` of this layer is padded by a large amount so that
+///    the gradient can be drawn outside of the `gradientRenderBounds`.
+///
+final class GradientRenderLayer: CAGradientLayer {
```
calda, line 22: This subclass lets us properly support drawing gradients outside the `bounds` of the `ShapeItemLayer`, which is necessary because the `path` of the `CAShapeLayer` isn't necessarily within the `bounds` of its parent layer.

In the left image, you can see what happens if you just use `CAGradientLayer` (the top-left corner of the small square is `bounds.origin` (`CGPoint.zero`). This subclass lets us draw the gradient above and to the left of that bounds rectangle, as if the gradient itself is "infinite" (that's the behavior of `CGContext.drawLinearGradient` with `[.drawsAfterEndLocation, .drawsBeforeStartLocation]`, which is used by the existing rendering engine).

| Without this subclass | With this subclass |
| ----- | ---- |
| reference_1_1658A012-E8C2-4AC3-87E9-B6A482880F3F | failure_2_1658A012-E8C2-4AC3-87E9-B6A482880F3F |

Sources/Private/Model/Layers/LayerModel.swift
```diff
@@ -88,7 +88,7 @@ class LayerModel: Codable {
   required init(from decoder: Decoder) throws {
     let container = try decoder.container(keyedBy: LayerModel.CodingKeys.self)
     name = try container.decodeIfPresent(String.self, forKey: .name) ?? "Layer"
-    index = try container.decodeIfPresent(Int.self, forKey: .index) ?? 0
+    index = try container.decodeIfPresent(Int.self, forKey: .index) ?? UUID().hashValue
```
calda, line 91: The experimental rendering engine requires that every layer has a unique `index`, so I changed the fallback from `0` to an `Int` derived from a `UUID`.

Sources/Private/Experimental/Extensions/CALayer+fillBounds.swift
```diff
@@ -3,18 +3,33 @@
...
 
-    bounds = superlayer.bounds
+      bounds = superlayer.bounds
+    }
   }
 }
+
+// MARK: - CustomLayoutLayer
+
+/// A `CALayer` that sets a custom `bounds` relative to its superlayer
```
erichoracek, line 32:
```suggestion
/// A `CALayer` that sets a custom `bounds` and `anchorPoint` relative to its superlayer
```

Sources/Private/Experimental/Animations/StarAnimation.swift
```diff
@@ -0,0 +1,74 @@
...
+      for: .path,
+      keyframes: star.position.keyframes,
+      value: { position in
+        // We can only use one set of keyframes to animate a given CALayer keypath,
+        // so we currently animate `position` and ignore any other keyframes.
+        // TODO: Is there a way to support this properly?
+        BezierPath.star(
+          position: position.pointValue,
+          outerRadius: star.outerRadius.keyframes.first!.value.cgFloatValue,
+          innerRadius: star.innerRadius?.keyframes.first?.value.cgFloatValue ?? 0,
```
erichoracek, line 41: Why do we force unwrap `keyframes.first` some of these and safely unwrap others? Might make sense to unify, and perhaps use `[0]` if we know the count will always be >0

Sources/Private/Experimental/Animations/ShapeAnimation.swift
```diff
@@ -1,28 +1,81 @@
...
+      addAnimations(for: ellipse, context: context)
+
+    case let rectangle as Rectangle:
+      addAnimations(for: rectangle, context: context)
+
+    case let star as Star:
+      addAnimations(for: star, context: context)
+
+    default:
+      return
```
erichoracek, line 23: Are there other shapes that we're missing? Might make sense to add a TODO if so or log if not

Sources/Private/Experimental/Animations/GradientAnimations.swift
```diff
@@ -0,0 +1,129 @@
...
+
+    // To draw the correct gradients, we have to derive a custom `endPoint`
+    // relative to the `startPoint` value. Since calculating the `endPoint`
+    // at any given time requires knowing the current `startPoint`,
+    // we can't allow them to animate separately.
+    guard
+      gradientFill.startPoint.keyframes.count == 1,
+      gradientFill.endPoint.keyframes.count == 1
+    else {
+      fatalError("Animating radial gradient `endPoint` point is currently unsupported")
```
erichoracek, line 83: Seems like we could fail more gracefully here vs. crashing the host app on an unsupported feature—unless you're planning on addressing immediately in a follow-up?

Sources/Private/Experimental/Animations/GradientAnimations.swift
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    let absoluteStartPoint = gradientFill.startPoint.keyframes.first!.value.pointValue
+    let absoluteEndPoint = gradientFill.endPoint.keyframes.first!.value.pointValue
```
erichoracek, line 87: nit: Why not do a guard let above? Seems like we check count and then force unwrap

Sources/Private/Model/Layers/LayerModel.swift
erichoracek: Seems like a kinda roundabout way to get to an `Int.random(…)` — do you have a reason to believe you'll get a better result from UUID?

Sources/Private/NodeRenderSystem/Nodes/PathNodes/PolygonNode.swift
```diff
@@ -88,14 +88,33 @@ final class PolygonNode: AnimatorNode, PathNode {
...
+    numberOfPoints: CGFloat,
+    outerRadius: CGFloat,
+    outerRoundedness inputOuterRoundedness: CGFloat,
+    rotation: CGFloat,
+    direction: PathDirection)
+    -> BezierPath
+  {
     var currentAngle = (rotation - 90).toRadians()
     let anglePerPoint = ((2 * CGFloat.pi) / numberOfPoints)
+    let outerRoundedness = inputOuterRoundedness * 0.01
```
erichoracek, line 117: nit: document this magic number?

Sources/Private/NodeRenderSystem/Nodes/PathNodes/StarNode.swift
```diff
@@ -99,18 +99,40 @@ final class StarNode: AnimatorNode, PathNode {
...
+    rotation: CGFloat,
+    direction: PathDirection)
+    -> BezierPath
+  {
     var currentAngle = (rotation - 90).toRadians()
     let anglePerPoint = (2 * CGFloat.pi) / numberOfPoints
     let halfAnglePerPoint = anglePerPoint / 2.0
     let partialPointAmount = numberOfPoints - floor(numberOfPoints)
+    let outerRoundedness = inoutOuterRoundedness * 0.01
+    let innerRoundedness = inputInnerRoundedness * 0.01
```
erichoracek, line 135: nit: document these magic numbers?

erichoracek (approved):
Exciting! Just a few questions/nits/comments

erichoracek:
> banding artifacts are just from the gif recording

Github supports mp4/movs now, which are much smaller and render more precisely than gifs!

Sources/Private/Model/Layers/LayerModel.swift
calda: `Int.random` sounds better, `UUID` was just the first thing I thought of

Sources/Private/Experimental/Layers/ImageLayer.swift
```diff
@@ -0,0 +1,50 @@
...
+  private let imageLayer: ImageLayerModel
+
+  private func setupSublayers(context: LayerContext) {
+    guard
+      let imageAsset = context.assetLibrary?.imageAssets[imageLayer.referenceID],
+      let image = context.imageProvider.imageForAsset(asset: imageAsset)
+    else { return }
+
+    let imageLayer = CALayer()
+    imageLayer.contents = image
```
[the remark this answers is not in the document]
calda: That's a nice improvement, didn't think to do that (`var contents: Any?` seems like a somewhat cursed API lol)

Sources/Private/Experimental/Layers/ImageLayer.swift
```diff
@@ -0,0 +1,50 @@
...
+
+  // MARK: Private
+
+  private let imageLayer: ImageLayerModel
+
+  private func setupSublayers(context: LayerContext) {
+    guard
+      let imageAsset = context.assetLibrary?.imageAssets[imageLayer.referenceID],
+      let image = context.imageProvider.imageForAsset(asset: imageAsset)
+    else { return }
```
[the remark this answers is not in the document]
calda: Good question, some thoughts:
 1. The legacy rendering engine silently ignores cases where referenced objects don't exist (e.g. parent layer by index, pre-comp layer assets, image assets) so it feels somewhat ok to continue doing that in the new engine.
 2. I think there's a somewhat reasonable expectation that the json will be well-formed, since they're typically output directly by bodymovin. 
Open to changing this if you think it's a good idea (if so we could also update parent layer and pre-comp layer asset lookup to log on failure)

Sources/Private/Experimental/Layers/GradientRenderLayer.swift
```diff
@@ -0,0 +1,86 @@
...
+  private func updateLayout() {
+    anchorPoint = .zero
+
+    bounds = CGRect(
+      x: gradientReferenceBounds.origin.x,
+      y: gradientReferenceBounds.origin.y,
+      width: gradientPadding + gradientReferenceBounds.width + gradientPadding,
+      height: gradientPadding + gradientReferenceBounds.height + gradientPadding)
+
+    transform = CATransform3DTranslate(
```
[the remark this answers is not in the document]
calda: Aha, I somehow missed `CATransform3DMakeTranslation` when I was looking at this earlier. That sounds better.

Sources/Private/Experimental/Layers/GradientRenderLayer.swift
```diff
@@ -0,0 +1,86 @@
...
+///
+final class GradientRenderLayer: CAGradientLayer {
+
+  // MARK: Internal
+
+  /// The reference bounds within this layer that the gradient's
+  /// `startPoint` and `endPoint` should be calculated relative to
+  var gradientReferenceBounds: CGRect = .zero {
+    didSet {
+      updateLayout()
```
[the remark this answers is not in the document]
calda: good idea

Sources/Private/Experimental/Extensions/CALayer+fillBounds.swift
```diff
@@ -3,18 +3,33 @@
...
+protocol CustomLayoutLayer {
```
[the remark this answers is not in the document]
calda: good idea

Sources/Private/Experimental/Animations/ShapeAnimation.swift
calda: This is all of them!

Sources/Private/Experimental/Animations/GradientAnimations.swift
calda: Good question -- I've been following this pattern elsewhere for now ([example](https://github.com/airbnb/lottie-ios/blob/master/Sources/Private/Experimental/Animations/EllipseAnimation.swift#L24). Relatively soon we're going to need to figure out a long term plan for unsupportable features (e.g. features that exist in the legacy rendering engine that just can't be expressed in a `CAKeyframeAnimation`). 

I have a feeling that the new engine can/will support enough features to become the default eventually, and we keep the legacy engine around to be used if necessary. Not totally sure though.

Sources/Private/Experimental/Animations/GradientAnimations.swift
calda: Good point, updated this to `[0]`

Sources/Private/NodeRenderSystem/Nodes/PathNodes/PolygonNode.swift
calda: Not reaally sure why this is necessary -- it was here before (https://github.com/airbnb/lottie-ios/blob/master/Sources/Private/NodeRenderSystem/Nodes/PathNodes/PolygonNode.swift#L92)

Sources/Private/NodeRenderSystem/Nodes/PathNodes/StarNode.swift
calda: https://github.com/airbnb/lottie-ios/pull/1471#discussion_r781712565

Sources/Private/Experimental/Animations/StarAnimation.swift
calda: Good idea, I updated this to just use `[0]` everywhere

Sources/Private/NodeRenderSystem/Nodes/PathNodes/PolygonNode.swift
erichoracek: Ah ok—wonder if it's converting from percentage to fractional again?

Sources/Private/Experimental/Animations/GradientAnimations.swift
erichoracek: I could imagine some sort of system where we make animation construction throwing and recover by falling back to the legacy main-thread renderer

Sources/Private/Experimental/Layers/ImageLayer.swift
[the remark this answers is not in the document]
erichoracek: I definitely agree that silent failures make sense in the case where the JSON is malformed and the error would not be recoverable by the client developer without exporting new JSON. However in cases where we have client engineer programmer error (e.g. the asset wasn't provided by the image provider) I think it would make sense to log as it would enable a client engineer to more easily diagnose why an animation isn't rendering correctly. I could also imagine a system like the [`EpoxyLogger`](https://git.musta.ch/airbnb/epoxy-ios/blob/master/Sources/EpoxyCore/Logging/EpoxyLogger.swift) to support logging warnings in a way that consumers can pipe into their own logging/alerting systems for assertions.

Sources/Private/Experimental/Animations/GradientAnimations.swift
erichoracek: > I've been following this pattern elsewhere for now

As far as I can tell this pattern is explicitly against our [style guide rule](https://github.com/airbnb/swift#preconditions-and-asserts):

> Handle an unexpected but recoverable condition with an assert method combined with the appropriate logging in production. If the unexpected condition is not recoverable, prefer a precondition method or fatalError(). This strikes a balance between crashing and providing insight into unexpected conditions in the wild. Only prefer fatalError over a precondition method when the failure message is dynamic, since a precondition method won't report the message in the crash report.

Ideally we'd be able to clean up this type of behavior at all possible in cases where the errors are recoverable as a framework should never be crashing a consumer app if at all possible.

erichoracek: > I've been following this pattern elsewhere for now

As far as I can tell this pattern is explicitly against our [style guide rule](https://github.com/airbnb/swift#preconditions-and-asserts):

> Handle an unexpected but recoverable condition with an assert method combined with the appropriate logging in production. If the unexpected condition is not recoverable, prefer a precondition method or fatalError(). This strikes a balance between crashing and providing insight into unexpected conditions in the wild. Only prefer fatalError over a precondition method when the failure message is dynamic, since a precondition method won't report the message in the crash report.

Ideally we'd be able to clean up this type of behavior at all possible in cases where the errors are recoverable as a framework should never be crashing a consumer app if at all possible.a

Sources/Private/Experimental/Layers/ImageLayer.swift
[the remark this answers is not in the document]
calda: Adding a logging system like `EpoxyLogger` seems like a good idea

Sources/Private/Experimental/Animations/GradientAnimations.swift
calda: > I could imagine some sort of system where we make animation construction throwing and recover by falling back to the legacy main-thread renderer

This is probably the best approach for making the new engine the default (We'll Get There™)

Agree that we shouldn't be `fatalError`ing here once this is used in production

PR #1473, opened 2022-01-11
Add matrix of Xcode/Swift versions to CI

erichoracek:
This enables us to check that Lottie compiles in Swift 5.3, 5.4, and 5.5, which has regressed with recent changes on `master`.

We also add a local setup action to reduce the setup code that's interspersed through all of the steps in the same style as the one added to Epoxy here https://github.com/airbnb/epoxy-ios/pull/74

Additionally fixes those Swift 5.3/5.4 compilation errors that prevent the project from building on those Xcode versions.


Example/Example.xcodeproj/project.pbxproj
```diff
@@ -507,6 +508,7 @@
 				CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER = YES;
 				CLANG_WARN_UNGUARDED_AVAILABILITY = YES_AGGRESSIVE;
 				CODE_SIGN_ENTITLEMENTS = macOS/macOS.entitlements;
+				CODE_SIGN_IDENTITY = "-";
```
erichoracek, line 511: This change was added by Xcode 13.2

Example/Example.xcodeproj/project.pbxproj
```diff
@@ -665,6 +668,7 @@
 				CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF = YES;
 				CLANG_WARN_OBJC_LITERAL_CONVERSION = YES;
 				CLANG_WARN_OBJC_ROOT_CLASS = YES_ERROR;
+				CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER = YES;
```
erichoracek, line 671: This change was added by Xcode 13.2

Example/iOS/ViewControllers/SampleListViewController.swift
```diff
@@ -112,7 +110,7 @@ final class SampleListViewController: CollectionViewController {
   private func configureSettingsMenu() {
     navigationItem.rightBarButtonItem = UIBarButtonItem(
       title: "Settings",
-      image: .init(systemName: "gear"),
+      image: UIImage(systemName: "gear")!,
```
erichoracek, line 113: Seems like we need to force unwrap here, not sure how this is compiling as the method takes a non-optional image

calda: weird!

Sources/Private/Experimental/Layers/ImageLayer.swift
```diff
@@ -24,12 +24,12 @@ final class ImageLayer: BaseCompositionLayer {
   /// Called by CoreAnimation to create a shadow copy of this layer
   /// More details: https://developer.apple.com/documentation/quartzcore/calayer/1410842-init
   override init(layer: Any) {
-    guard let layer = layer as? Self else {
```
calda, line 27: Interesting, didn't know this used to be unsupported

.github/workflows/main.yml
```diff
@@ -10,58 +10,100 @@ jobs:
   build-package:
     name: "Build Package"
     runs-on: macos-latest
+    strategy:
+      matrix:
+        xcode:
+        - '12.4' # Swift 5.3
```
calda, line 16: Why do we need to support Swift 5.3? Just curious

calda (approved):

calda (approved):
Thanks!

.github/workflows/main.yml
erichoracek: We don't have to, it was just previously supported in `master` based on https://swiftpackageindex.com/airbnb/lottie-ios so I continued to support it

calda: Good to know!

calda:
Maybe we should update `README.md`, where it currently says:

> Lottie is written in Swift 5.4.

To something like:

> Lottie supports Swift 5.3 and later.

Sources/Private/Experimental/Layers/ImageLayer.swift
erichoracek: This was supported, it's the fact that we're using `layer` in the fatalError` that was unsupported

calda: oooh I see

erichoracek:
> Maybe we should update `README.md`, where it currently says:
> 
> > Lottie is written in Swift 5.4.
> 
> To something like:
> 
> > Lottie supports Swift 5.3 and later.

Good point, I added a Swift Package Index badge here 9a7b7e086292fdc1b7a1e4a5c9ed45054b7f6352

PR #1474, opened 2022-01-11
Improve logging / assertions system, remove fatalErrors

calda:
This PR updates our logging system to match Epoxy's `EpoxyLogger`, and cleaned up / centralized some of the handling for features that are not supported by the experimental rendering engine.

### DIFF
--- a/Sources/Public/Animation/AnimationPublic.swift
+++ b/Sources/Public/Animation/AnimationPublic.swift
@@ -69,7 +69,7 @@ extension Animation {
       return animation
     } catch {
       /// Decoding error.
-      LottieLogger.shared.assertionFailure("Error when decoding animation \"\(name)\": \(error)")
+      LottieLogger.shared.warn("Error when decoding animation \"\(name)\": \(error)")
       return nil
     }
   }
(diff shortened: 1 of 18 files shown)


Sources/Public/Animation/AnimationPublic.swift
```diff
@@ -69,7 +69,7 @@ extension Animation {
       return animation
     } catch {
       /// Decoding error.
-      print(error)
+      LottieLogger.shared.assertionFailure("Error when decoding animation \"\(name)\": \(error)")
```
erichoracek, line 72: Wondering if you think this should be a `warn`? I could imagine a case where you want to recover from an invalid animation and not have it crash in debug e.g. if you're writing a lottie viewer app

erichoracek (approved):
LGTM! Just a few small questions

Sources/Public/Animation/AnimationPublic.swift
calda: Sure

calda:
```suggestion
      LottieLogger.shared.warn("Error when decoding animation \"\(name)\": \(error)")
```

PR #1475, opened 2022-01-11
Fix issues with layer ordering, support Keyframe.isHold

calda:
This PR fixes issues with layer ordering, and adds support for `Keyframe.isHold`

With these two fixes, several animations are looking a lot better. For example, here's the Watermelon:

| Before | After |
| ----- | ---- |
|  |  |


Sources/Private/Experimental/Animations/CAKeyPath.swift
```diff
@@ -0,0 +1,43 @@
+// Created by Cal Stephens on 1/11/22.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+import QuartzCore
+
+// MARK: - CAKeyPath
+
+/// A strongly typed value that can be used as the `keyPath` of a `CAAnimation`
+struct CAKeyPath<ValueRepresentation> {
```
calda, line 9: Moved this to a separate file with no changes

Sources/Private/Experimental/Layers/CALayer+setupLayerHierarchy.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    let layersByIndex = Dictionary(grouping: layersInZAxisOrder, by: \.index)
+      .compactMapValues(\.first)
+
+    /// Layers specify a `parent` layer. Child layers inherit the `transform` of their parent.
+    ///  - We can't add the child as a sublayer of the parent `CALayer`, since that would
+    ///    break the ordering specified in `layersInZAxisOrder`.
+    ///  - Instead, we create an invisible `TransformLayer` to handle the parent
+    ///    transform animations, and add the child layer to that `TransformLayer`.
+    func setupParentTransformLayers(
```
calda, line 27: The previous setup, where we would add children to their parent using `addSublayer`, was incorrect.

To render animations correctly, we need to exactly respect `layersInZAxisOrder`. It's possible to create orderings that are incompatible with an actual sublayer hierarchy. For example, if you have `[A, B, C]` where `C` is a child of `A`, then adding `C` as a sublayer of `A` would incorrectly change the layer ordering to `[A, C, B]`.

Instead, we now create invisible `TransformLayer`s to display / animate the `transform` of the parent. This makes sure the child layer correctly inherits the parent transform without affecting layer ordering.

This fixes a bunch of issues with the sample animations. Some examples:

| Before | After |
| ----- | ---- |
| Simulator Screen Shot - iPhone 8 - 2022-01-11 at 12 45 44 | Simulator Screen Shot - iPhone 8 - 2022-01-11 at 12 46 49 |
| Simulator Screen Shot - iPhone 8 - 2022-01-11 at 12 46 16 | Simulator Screen Shot - iPhone 8 - 2022-01-11 at 12 47 24 |

Sources/Private/Experimental/Layers/BaseCompositionLayer.swift
```diff
@@ -35,46 +35,23 @@ class BaseCompositionLayer: CALayer, AnimationLayer {
...
-  let transformLayer: CALayer = {
-    let layer = CALayer()
-    layer.name = "transform"
-    return layer
-  }()
-
-  /// The layer that should render visible elements displayed by this layer's `LayerModel`
-  ///  - This layer applies `LayerModel.transform.opacity`, which applies to this specific
-  ///    `LayerModel`'s content and not the content of any child `LayerModel`s.
-  let contentsLayer: CALayer = {
```
calda, line 50: The distinction between `transformLayer` and `contentsLayer` is no longer necessary, now that we don't add child `LayerModel`s as sublayers

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -142,69 +142,7 @@ final class ExperimentalAnimationLayer: CALayer {
 
 }
 
-// MARK: - CALayer + setupLayerHierarchy
-
-extension CALayer {
-  /// Sets up an `AnimationLayer` / `CALayer` hierarchy in this layer,
-  /// using the given list of layers.
-  func setupLayerHierarchy(
```
calda, line 150: Moved this to its own file (The new implementation is completely different from this old implementation though)

Sources/Private/Experimental/Layers/CALayer+setupLayerHierarchy.swift
erichoracek: Great solve!

erichoracek (approved):
Good stuff!

PR #1476, opened 2022-01-11
Fix artifacts from implicit sublayer animation in ShapeItemLayer

calda:
This PR fixes a visual artifact in `ShapeItemLayer` caused by an implicit `sublayer` animation. The fix was to setup the sublayers in `init` instead of in `setupAnimations`.

| Before (notice how the dots flash black for a moment as the animation starts) | After |
| ---- | ---- |
|  |  |


erichoracek (approved):
LGTM!

PR #1478, opened 2022-01-12
Improve performance when interactively scrubbing complex animations

calda:
The PR improves performance when interactively scrubbing the `currentFrame` of complex animations.

Previously, we discarded and rebuilt the entire layer hierarchy when setting `currentFrame`. This was pretty costly for large / complex animations, and could max-out the CPU.

Now, we just set `layer.timeOffset`, which doesn't have any overhead.

| Before | After |
| ----- | ----- |
|  |  |


Sources/Private/Experimental/Animations/CAAnimation+TimingConfiguration.swift
```diff
@@ -68,6 +68,24 @@ extension CAAnimation {
...
     clippingParent.isRemovedOnCompletion = false
 
+    // We can pause the animation on a specific frame by setting the root layer's
+    // `speed` to 0, and then setting the `timeOffset` for the given frame.
+    //  - For that setup to work properly, we have to set the `beginTime`
+    //    of this animation to a time slightly before the current time.
+    //  - It's not really clear why this is necessary, but `timeOffset`
+    //    is not applied correctly without this configuration.
+    let currentTime = layer.convertTime(CACurrentMediaTime(), from: nil)
+    clippingParent.beginTime = currentTime - .leastNonzeroMagnitude
```
calda, line 78: This is somewhat cursed, but it fixes an issue where the `timeOffset` would be applied incorrectly. Without the `- .leastNonzeroMagnitude`, the actual displayed frame is somehow dependent on the frame of the animation when switching from noninteractive to interactive playback:

| Before | After |
| ---- | ---- |
| ezgif-2-1fde89d872 |  |

calda: The fact that `clippingParent.beginTime = currentTime - .leastNonzeroMagnitude` works but `clippingParent.beginTime = currentTime` doesn't kinda blows my mind, considering `Double.leastNonzeroMagnitude` is `5e-324`.

erichoracek: Huh how strange, glad you found a path forward 😅

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -155,21 +157,24 @@ extension ExperimentalAnimationLayer: RootAnimationLayer {
...
-        context: .init(
-          playFrom: animation.startFrame,
-          playTo: animation.endFrame,
-          closure: nil),
-        timingConfiguration: .init(
-          speed: 0,
-          timeOffset: animation.time(forFrame: newValue)))
+      // We can display a specific frame of the animation by setting
+      // `timeOffset` of this layer. This requires setting up the layer hierarchy
+      // with a specific configuration (speed=0, etc) at least one. But if
```
erichoracek, line 162:
```suggestion
      // with a specific configuration (speed=0, etc) at least once. But if
```

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -155,21 +157,24 @@ extension ExperimentalAnimationLayer: RootAnimationLayer {
...
+      // the layer hierarchy is already set up correctly, we can update the
+      // `timeOffset` very cheaply.
+      let requiredTimingConfiguration = CAMediaTimingConfiguration(speed: 0)
+
+      if
+        currentAnimationConfiguration?.timingConfiguration != requiredTimingConfiguration
+        || currentAnimationConfiguration?.animationContext.playFrom != animation.startFrame
+        || currentAnimationConfiguration?.animationContext.playTo != animation.endFrame
```
erichoracek, line 170: nit: might be simpler to read if this logic is a method on `currentAnimationConfiguration`

erichoracek (approved):
LGTM!

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
calda: I'm tempted to make `AnimationConfiguration` equatable, but it contains a class that wraps a closure so I'm not entirely sure if it would be a good idea / idiomatic

calda: Ah, well that `AnimationCompletionDelegate` is already an `NSObject` subclass so it should already be `Equatable`. Maybe a non-issue then.

PR #1479, opened 2022-01-13
Fix keyframe timing regression when playing animations in real time

calda:
This PR fixes a regression introduced by https://github.com/airbnb/lottie-ios/pull/1478, where keyframe timing could be incorrect when playing animations in real time (e.g. when not just setting `currentFrame`).

We don't have automated tests for real time animations, which is why this wasn't caught by CI (I've tried to get these to work several times and haven't been able to figure out how).

| Before | After |
| ----- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-01-12 at 12 44 40 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-01-12 at 12 44 50 |

### DIFF
--- a/Example/iOS/ViewControllers/AnimationPreviewViewController.swift
+++ b/Example/iOS/ViewControllers/AnimationPreviewViewController.swift
@@ -22,6 +22,11 @@ class AnimationPreviewViewController: UIViewController {
 
   // MARK: Internal
 
+  override func viewDidDisappear(_ animated: Bool) {
+    super.viewDidDisappear(animated)
+    displayLink?.invalidate()
+  }
+
   override func viewDidLoad() {
     super.viewDidLoad()
     view.backgroundColor = .systemBackground
--- a/Sources/Private/Experimental/ExperimentalFeatureConfiguration.swift
+++ b/Sources/Private/Experimental/ExperimentalFeatureConfiguration.swift
@@ -7,6 +7,10 @@
 /// to change or be removed at any time.
 public struct ExperimentalFeatureConfiguration {
 
+  public init(useNewRenderingEngine: Bool = false) {
+    self.useNewRenderingEngine = useNewRenderingEngine
+  }
+
   /// The singleton configuration for experimental features,
   /// which applies to all `AnimationView`s by default.
   public static var shared = ExperimentalFeatureConfiguration()
@@ -15,9 +19,5 @@ public struct ExperimentalFeatureConfiguration {
   /// which leverages the Core Animation render server to
   /// animate without executing on the main thread every frame.
   public var useNewRenderingEngine: Bool
-
-  public init(useNewRenderingEngine: Bool = false) {
-    self.useNewRenderingEngine = useNewRenderingEngine
-  }
 
 }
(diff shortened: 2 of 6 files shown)


Example/iOS/ViewControllers/AnimationPreviewViewController.swift
```diff
@@ -22,6 +22,11 @@ class AnimationPreviewViewController: UIViewController {
 
   // MARK: Internal
 
+  override func viewDidDisappear(_ animated: Bool) {
+    super.viewDidDisappear(animated)
+    displayLink?.invalidate()
```
calda, line 27: I noticed this VC / display link was leaking (because `CADisplayLink` retains its `target` until invalidated, apparently)

Sources/Public/iOS/AnimatedSwitch.swift
```diff
@@ -160,11 +163,19 @@ open class AnimatedSwitch: AnimatedControl {
...
+    animationView.play(
+      fromProgress: startProgress,
+      toProgress: endProgress,
+      loopMode: LottieLoopMode.playOnce,
+      completion: { [weak self] finished in
+        guard let self = self else { return }
+
+        // For the legacy rendering engine, we freeze the animation at the expected final progress
+        // once the animation is complete. This isn't necessary on the new / experimental engine.
+        if finished, !(self.animationView.animationLayer is ExperimentalAnimationLayer) {
```
calda, line 175: #1478 also broke `AnimatedSwitch` animations somehow, not really sure why. This change fixes it.




We don't currently have automated tests for `AnimatedSwitch`. I spent some time this morning trying to set up an automated test that called this, waited for the animation to finish, and then validated that the final `currentProgress` was correct. I wasn't able to get the animation to play correctly in a headless environment (the completion handler was always called immediately with `finished: false`).

Example/iOS/ViewControllers/AnimationPreviewViewController.swift
erichoracek: Oh huh, I thought targets were supposed to be weak. Sounds like a CADisplayLink oddity

Sources/Private/Experimental/ExperimentalFeatureConfiguration.swift
```diff
@@ -16,4 +16,8 @@ public struct ExperimentalFeatureConfiguration {
   /// animate without executing on the main thread every frame.
   public var useNewRenderingEngine = false
```
erichoracek, line 17: nit: remove default value since it's specified in the initializer so there's a single source of truth
```suggestion
  public var useNewRenderingEngine: Bool
```

Sources/Private/Experimental/ExperimentalFeatureConfiguration.swift
```diff
@@ -16,4 +16,8 @@ public struct ExperimentalFeatureConfiguration {
...
 
+  public init(useNewRenderingEngine: Bool = false) {
```
erichoracek, line 19: nit: why not move this up top?

Sources/Public/iOS/AnimatedSwitch.swift
erichoracek: 😞 bummer, glad the new engine doesn't need this though

erichoracek (approved):
LGTM!

Sources/Private/Experimental/ExperimentalFeatureConfiguration.swift
calda: Good idea

PR #1483, opened 2022-01-19
Rebuild current animation after registering new value provider

calda:
This PR fixes an issue where new `ValueProvider`s wouldn't take effect until the next time an animation is set up.

| Before | After |
| ----- | ----- |
|  |  |

### DIFF
--- a/Sources/Private/Experimental/ExperimentalAnimationLayer.swift
+++ b/Sources/Private/Experimental/ExperimentalAnimationLayer.swift
@@ -123,6 +123,13 @@ final class ExperimentalAnimationLayer: CALayer {
   // Configuration for the animation that is currently setup in this layer
   private var currentAnimationConfiguration: AnimationConfiguration?
 
+  /// The current progress of the placeholder `CAAnimation`,
+  /// which is also the realtime animation progress of this layer's animation
+  @objc private var animationProgress: CGFloat = 0
+
+  private let animation: Animation
+  private let valueProviderStore = ValueProviderStore()
+
   // The current state of the animation that is playing in this layer
   private var playbackState: PlaybackState? {
     didSet {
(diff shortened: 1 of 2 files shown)


Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -110,7 +113,14 @@ final class ExperimentalAnimationLayer: CALayer {
     let timingConfiguration: CAMediaTimingConfiguration
   }
 
-  // Configuration for the animation that is currently playing in this layer
+  private enum PlaybackState: Equatable {
+    /// The animation is playing in real-time
+    case realtime
+    /// The animation is statically displaying a specific frame
+    case specificFrame(AnimationFrameTime)
```
erichoracek, line 120: nit: to me these case names are a bit confusing, I personally would find this more obvious if it was:
```suggestion
    /// The animation is playing in real-time
    case playing
    /// The animation is statically displaying a specific frame
    case paused(frame: AnimationFrameTime)
```

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -120,6 +130,18 @@ final class ExperimentalAnimationLayer: CALayer {
   private let animation: Animation
   private let valueProviderStore = ValueProviderStore()
 
+  // The current state of the animation that is playing in this layer
+  private var playbackState: PlaybackState? {
+    didSet {
+      switch playbackState {
```
erichoracek, line 136: nit: guard on equality with `oldValue` before setting `timeOffset`?

erichoracek (approved):
@calda Good stuff, just one naming nit!

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
calda: Nice, much better!

PR #1484, opened 2022-01-20
Add support for setting a custom AnimationImageProvider

calda:
This PR adds support for setting a custom `AnimationImageProvider` when using the experimental rendering engine


Tests/Utils/HardcodedImageProvider.swift
```diff
@@ -0,0 +1,21 @@
...
+/// An `AnimationImageProvider` that always returns a specific hardcoded image of a dog (hi Marley)
+struct HardcodedImageProvider: AnimationImageProvider {
+  func imageForAsset(asset _: ImageAsset) -> CGImage? {
+    #if os(iOS)
+    return UIImage(named: "Samples/Images/dog.png", in: .module, compatibleWith: nil)?.cgImage
+    #else
+    return nil
+    #endif
+  }
+}
```
erichoracek, line 21: nit: add a `name` property to this provider so we can use it in other places?
```suggestion
struct HardcodedImageProvider: AnimationImageProvider {
  var name: String
  
  func imageForAsset(asset _: ImageAsset) -> CGImage? {
    #if os(iOS)
    return UIImage(named: name, in: .module, compatibleWith: nil)?.cgImage
    #else
    return nil
    #endif
  }
}
```

erichoracek (approved):
LGTM! Just a few small nits

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -48,6 +53,8 @@ final class ExperimentalAnimationLayer: CALayer {
     var timeOffset: TimeInterval = 0
   }
 
+  var imageProvider: AnimationImageProvider
```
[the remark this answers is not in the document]
calda: This needs to be a `var` because it implements `var imageProvider: AnimationImageProvider { get set }` from `RootAnimationLayer`.

I added a `didSet` to make sure any existing `ImageLayer`s are reconfigured using the new `AnmationImageProvider` (and tweaked the tests so it fails without this).

Tests/Utils/HardcodedImageProvider.swift
calda: good idea

PR #1485, opened 2022-01-20
Rename LottieView to AnimationViewBase

erichoracek:
This makes it more clear how this API is meant to be used (as an implementation detail to enable cross platform compatibility), and differentiates it from `AnimationView` to make it easier for consumers to understand that the public Lottie API is available on `AnimationView`, not `LottieView`.

While we're here we also remove the `open` from `AnimationViewBase` to make it more clear that it shouldn't be subclassed by consumers of Lottie.


calda (approved):
Thanks!

PR #1486, opened 2022-01-21
Improve integration between AnimationView and ExperimentalAnimationLayer

calda:
This PR improves the integration between `AnimationView` and `ExperimentalAnimationLayer`.

I'm working on adopting the new rendering engine in the Airbnb app. There are some cases where it hits a `Size must be non-zero before an animation can be played` assertion, even though the `AnimationView` has a valid frame / layout. This was because there was a different setup codepath for `ExperimentalAnimationLayer`, with slightly different behavior. This PR consolidates the two setup paths so they have more similar behavior.

### DIFF
--- a/Tests/SnapshotTests.swift
+++ b/Tests/SnapshotTests.swift
@@ -29,12 +29,6 @@ class SnapshotTests: XCTestCase {
   /// Validates that all of the snapshots in __Snapshots__ correspond to
   /// a sample JSON file that is visible to this test target.
   func testAllSnapshotsHaveCorrespondingSampleFile() {
-    // We don't want assertions to crash the tests, so we stub out the shared logger singleton
-    LottieLogger.shared = LottieLogger(
-      assert: { _, _, _, _ in },
-      assertionFailure: { _, _, _ in },
-      warn: { _, _, _ in })
-
     for snapshotURL in snapshotURLs {
       // The snapshot files follow the format `testCaseName.animationName-percentage.png`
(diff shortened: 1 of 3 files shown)


Tests/SnapshotTests.swift
```diff
@@ -29,12 +29,6 @@ class SnapshotTests: XCTestCase {
   /// Validates that all of the snapshots in __Snapshots__ correspond to
   /// a sample JSON file that is visible to this test target.
   func testAllSnapshotsHaveCorrespondingSampleFile() {
-    // We don't want assertions to crash the tests, so we stub out the shared logger singleton
-    LottieLogger.shared = LottieLogger(
```
calda, line 33: I added this in #1484 but accidentally put it in the wrong place

Tests/SnapshotTests.swift
```diff
@@ -155,10 +162,9 @@ class SnapshotTests: XCTestCase {
           _experimentalFeatureConfiguration: ExperimentalFeatureConfiguration(
             useNewRenderingEngine: usingExperimentalRenderingEngine))
 
-        // Set up the animation view with a valid frame and layout
+        // Set up the animation view with a valid frame
         // so the geometry is correct when setting up the `CAAnimation`s
         animationView.frame.size = animation.snapshotSize
-        animationView.layoutIfNeeded()
```
calda, line 161: When I was testing things I noticed the tests pass without this, so I figured we may as well remove it

erichoracek (approved):
LGTM!

PR #1487, opened 2022-01-21
Rename ExperimentalFeatureConfiguration to LottieConfiguration

calda:
This PR renames `ExperimentalFeatureConfiguration` to `LottieConfiguration`. This is starting to be production ready, so I think we can drop the "Experimental" prefix.


Tests/SnapshotConfiguration.swift
```diff
@@ -14,11 +14,6 @@ struct SnapshotConfiguration {
   ///    but should be kept as high as possible (while still permitting the diff to succeed)
   var precision: Float = 1
 
-  /// Whether or not this snapshot should be tested using the experimental rendering engine
-  ///  - Defaults to false, since the experimental rendering engine
-  ///    currently supports a relatively small number of animations.
-  var testWithExperimentalRenderingEngine = false
```
calda, line 20: I removed this flag of havingI removed this flag in favor of having it enabled for all sample animations now. Most of the samples in the Example app work well now in the new engine.

erichoracek (approved):
LGTM!

PR #1490, opened 2022-01-21
Make AnimationView.configuration public

calda:
This PR makes `AnimationView.configuration` public. This is immutable, so there's no reason for it to be private. This is also necessary for any consumer to take an `AnimationView` and make an exact copy (if necessary).


erichoracek (approved):
LGTM

PR #1491, opened 2022-01-21
Add keyframe animation support to ValueProviders API

calda:
This PR adds support for adding `ValueProvider`s that perform animations, by specifying a set of keyframes.

The `ValueProvider` API currently supports custom animations by allowing consumers to pass in a `(AnimationFrameTime) -> Value` closure. The main thread rendering engine just called this closure every frame, which we can't do in the Core Animation rendering engine.

To provide an alternative option supported by the new rendering engine, we now permit `ValueProvider`s to provide a `KeyframeGroup`.

| Before | After |
| ---- | ---- |
|  |  |


Sources/Private/Experimental/ValueProviderStore.swift
```diff
@@ -12,8 +14,8 @@ final class ValueProviderStore {
   /// Registers an `AnyValueProvider` for the given `AnimationKeypath`
   func setValueProvider(_ valueProvider: AnyValueProvider, keypath: AnimationKeypath) {
     LottieLogger.shared.assert(
-      valueProvider.typeErasedStorage.isSingleValue,
-      "The new rendering engine only supports Value Providers with a single fixed value")
+      valueProvider.typeErasedStorage.isSupportedByCoreAnimationRenderingEngine,
+      "The Core Animation rendering engine doesn't support Value Providers that vend a closure.")
```
erichoracek, line 18: nit: might make sense to add a comment detailing that it is impossible to have an API like this with Core Animation, as-is this sounds like it may be added in the future

Sources/Private/Experimental/ValueProviderStore.swift
```diff
@@ -28,37 +30,67 @@ final class ValueProviderStore {
...
     switch anyValueProvider.typeErasedStorage {
     case .singleValue(let typeErasedValue):
+      typeErasedKeyframes = KeyframeGroup(typeErasedValue)
+
+    case .keyframes(let keyframes, _):
+      typeErasedKeyframes = keyframes
+
+    case .closure:
+      LottieLogger.shared.assertionFailure("""
+      The new rendering engine only supports Value Providers with a single fixed value
```
erichoracek, line 55: nit: same here

Sources/Private/Model/Keyframes/Keyframe.swift
```diff
@@ -14,12 +14,12 @@ import Foundation
  Keyframe represents a point in time and is the container for datatypes.
  Note: This is a parent class and should not be used directly.
```
erichoracek, line 15: Should this note be reworded now that this is part of the public API? And this type moved out of the private folder?

Sources/Private/Model/Keyframes/KeyframeGroup.swift
```diff
@@ -7,6 +7,8 @@
 
 import Foundation
 
+// MARK: - KeyframeGroup
+
 /**
  Used for coding/decoding a group of Keyframes by type.
```
erichoracek, line 13: Should this note be reworded now that this is part of the public API? And should this be moved out of the private folder?

Sources/Private/Model/Keyframes/KeyframeGroup.swift
```diff
@@ -111,10 +127,54 @@ final class KeyframeGroup<T>: Codable where T: Codable, T: Interpolatable {
...
+
+// MARK: - Collection helpers
+
+extension KeyframeGroup {
+  // Maps the value of each individual keyframe in this group using the given closure
+  func map<Result>(_ mapping: (T) -> Result) -> KeyframeGroup<Result> {
+    KeyframeGroup<Result>(keyframes: .init(keyframes.map { keyframe in
+      keyframe.withValue(mapping(keyframe.value))
+    }))
+  }
```
erichoracek, line 156: nit: the closure passed to map is typically named `transform`
```suggestion
  // Maps the value of each individual keyframe in this group using the given closure
  func map<Result>(_ transform: (T) -> Result) -> KeyframeGroup<Result> {
    KeyframeGroup<Result>(keyframes: .init(keyframes.map { keyframe in
      keyframe.withValue(transform(keyframe.value))
    }))
  }
```

Sources/Private/Model/Keyframes/KeyframeGroup.swift
```diff
@@ -111,10 +127,54 @@ final class KeyframeGroup<T>: Codable where T: Codable, T: Interpolatable {
...
+
+  // Maps the value of each individual keyframe in this group using the given closure,
+  // filtering out `nil`s.
+  func compactMap<Result>(_ mapping: (T) -> Result?) -> KeyframeGroup<Result> {
+    KeyframeGroup<Result>(keyframes: .init(keyframes.compactMap { keyframe in
+      guard let mappedValue = mapping(keyframe.value) else { return nil }
+      return keyframe.withValue(mappedValue)
+    }))
+  }
```
erichoracek, line 165: nit: the closure passed to map is typically named `transform`
```suggestion
  // Maps the value of each individual keyframe in this group using the given closure,
  // filtering out `nil`s.
  func compactMap<Result>(_ transform: (T) -> Result?) -> KeyframeGroup<Result> {
    KeyframeGroup<Result>(keyframes: .init(keyframes.compactMap { keyframe in
      guard let mappedValue = transform(keyframe.value) else { return nil }
      return keyframe.withValue(mappedValue)
    }))
  }
```

Sources/Private/Utility/Extensions/MathKit.swift
```diff
@@ -25,6 +25,20 @@ extension Double {
 
 extension CGFloat: Interpolatable {
 
+  // MARK: Public
+
+  public func interpolateTo(
```
erichoracek, line 30: If we're making these interpolation methods part of the public API should they get some documentation? I'm not really sure how these are intended to be used as-is

Sources/Private/Utility/Interpolatable/Interpolatable.swift
```diff
@@ -8,7 +8,7 @@
 import CoreGraphics
 import Foundation
 
-protocol Interpolatable {
+public protocol Interpolatable {
```
erichoracek, line 11: nit: same here re: public API documentation

Sources/Private/Utility/Interpolatable/InterpolatableExtensions.swift
```diff
@@ -108,6 +126,16 @@ extension Color: Interpolatable {
     self.a = a
   }
 
+  // MARK: Public
+
+  public func interpolateTo(_ to: Color, amount: CGFloat, spatialOutTangent _: CGPoint?, spatialInTangent _: CGPoint?) -> Color {
```
erichoracek, line 131: nit: wrap long method declarations?

Sources/Private/Utility/Interpolatable/InterpolatableExtensions.swift
```diff
@@ -205,11 +226,22 @@ extension TextDocument: Interpolatable {
...
+extension Array: Interpolatable where Element: Interpolatable {
+  public func interpolateTo(
+    _ to: [Element],
+    amount: CGFloat,
+    spatialOutTangent: CGPoint?,
+    spatialInTangent: CGPoint?)
+    -> [Element]
+  {
+    var returnArray = [Element]()
     for i in 0..<count {
```
erichoracek, line 238: nit:
```suggestion
    for i in indices {
```

Sources/Private/Utility/Interpolatable/InterpolatableExtensions.swift
```diff
@@ -205,11 +226,22 @@ extension TextDocument: Interpolatable {
...
-  func interpolateTo(_ to: [Element], amount: CGFloat, spatialOutTangent _: CGPoint?, spatialInTangent _: CGPoint?) -> [Element] {
-    var returnArray = [Double]()
+extension Array: Interpolatable where Element: Interpolatable {
+  public func interpolateTo(
+    _ to: [Element],
+    amount: CGFloat,
+    spatialOutTangent: CGPoint?,
+    spatialInTangent: CGPoint?)
+    -> [Element]
+  {
```
erichoracek, line 236: nit: assertion on count and to.count being equal?

Sources/Private/Utility/Primitives/VectorsExtensions.swift
```diff
@@ -54,7 +54,7 @@ extension Double {
 /**
  Needed for decoding json {x: y:} to a CGPoint
  */
-struct Vector2D: Codable {
+public struct Vector2D: Codable, Hashable {
```
erichoracek, line 57: Same question here re: moving this out of the private folder now that it's public API?

Sources/Public/DynamicProperties/AnyValueProvider.swift
```diff
@@ -42,51 +42,84 @@ extension AnyValueProvider {
...
+  public var typeErasedStorage: AnyValueProviderStorage {
     switch storage {
     case .closure(let typedClosure):
       return .closure(typedClosure)
+
     case .singleValue(let typedValue):
       return .singleValue(typedValue)
+
+    case .keyframes(let keyframes):
+      return .keyframes(keyframes.map { $0 }, interpolate: storage.value(frame:))
```
erichoracek, line 61: What is the purpose of the identity mapping here? If there is one can we add a comment?
```suggestion
      return .keyframes(keyframes, interpolate: storage.value(frame:))
```

erichoracek (approved):
@calda Good stuff!

Sources/Private/Model/Keyframes/Keyframe.swift
calda: aha good point

Sources/Private/Model/Keyframes/KeyframeGroup.swift
calda: I kept `KeyframeGroup` internal in favor of just using `[Keyframe]` in the public API. (`KeyframeGroup` is basically just a thin wrapper around an array that implements `Codable`, which we don't actually need in the public API).

Sources/Public/DynamicProperties/AnyValueProvider.swift
calda: This was because `KeyframeGroup<T>` isn't implicitly convertible to `KeyframeGroup<Any>`. (I removed the `KeyframeGroup.map` helper when I changed it back to being internal, so this isn't here anymore)

Issue #1492, opened 2022-01-22
Carthage build result can not be used on different machine


thangngoquocOura:
## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.3.0 - master

## What Platform are you on? 

- [x] iOS

## What Language are you in?
- [x] Swift

## Expected Behavior 
Pre-built xcframework of Lottie using Carthage (`carthage bootstrap --platform iOS --use-xcframeworks --cache-builds`) is portable and can be used on different dev machine.

Our setup includes checking out the `Carthage/Build` folder, the CI machine can uses the pre-built xcframeworks of all dependencies. In v3.2.3 it used to work.

## Actual Behavior
In v3.3.0 I changed the Carthfile to point to `master` according to https://github.com/airbnb/lottie-ios/issues/1448#issuecomment-998197310 It works on my local env after running `carthage update --platform iOS --use-xcframeworks --cache-builds`, but the CI can't compile the project with error message "Missing required module LottieCore".

After trying to run `carthage update lottie-ios --platform iOS --use-xcframeworks` on the CI machine, it can compile the project successfully.

The CI machine and my local env are using the same Xcode, tool chain and macOS version, they are both x86 intel Mac. Switching back to 3.2.3 works as expected.

## Animation JSON

calda:
Fixed in #1496 (which has been merged, so this should be fixed on `master` now). Thanks again for reporting this issue.

PR #1493, opened 2022-01-24
Test building with Carthage when running 'carthage update' on another machine

calda:
This draft PR tests whether or not the `CarthageTest` project build correctly when `carthage update` was run on a different machine (#1492).

I ran `carthage update` locally, committed the built framework, and then removed the `carthage update` step from `bundle exec rake test:carthage`.


calda:
The `Test Carthage support` step fails on CI even though it works on my machine, so there must be some sort of hidden dependency in the `Lottie.xcframework` that Carthage outputs

calda:
My local build fails if I delete `/Users/cal/Library/Caches/org.carthage.CarthageKit/DerivedData/13.2.1_13C100/lottie-ios/5bba5dcbb618ff392f916981bb730e2e68d3f57c/Build/Products/Release-iphonesimulator/LottieCore.swiftmodule`

PR #1494, opened 2022-01-24
Support wildcards in AnimationKeypaths

calda:
This PR adds support for wildcards (`**`) in `AnimationKeypath`s


Sources/Private/Experimental/ValueProviderStore.swift
```diff
@@ -99,3 +107,22 @@ extension AnyValueProviderStorage {
...
+    //  - "**.Color" matches both "Layer 1.Color" and "Layer 1.Layer 2.Color"
+    regex = regex.replacingOccurrences(of: "**", with: ".+")
+
+    // * wildcards match any individual path component
+    //  - "*.Color" matches "Layer 1.Color" but not "Layer 1.Layer 2.Color"
+    regex = regex.replacingOccurrences(of: "*", with: "[^.]+")
+
+    return keypath.fullPath.range(of: regex, options: .regularExpression) != nil
+  }
+}
```
erichoracek, line 128: Do you think we could write some small unit tests on this logic? This method is super well documented but it feels like it might be helpful to test this too as it doesn't seem like the snapshot tests validate single * wildcards

Sources/Private/Experimental/ValueProviderStore.swift
```diff
@@ -83,7 +78,20 @@ final class ValueProviderStore {
...
+  private func valueProvider(for keypath: AnimationKeypath) -> AnyValueProvider? {
+    // Iterate backwards though the list of value providers,
+    // so the first match we find is the one that was registered most recently
+    for (registeredKeypath, valueProvider) in valueProviders.reversed() {
+      if registeredKeypath.matches(keypath) {
+        return valueProvider
+      }
+    }
+
+    return nil
```
erichoracek, line 93: nit: if you want to be concise!
```suggestion
    valueProviders.reversed().first(where: { $0.matches($1) }.valueProvider
```

erichoracek (approved):
@calda LGTM! Just one suggestion/question

Sources/Private/Experimental/ValueProviderStore.swift
calda: That seems like a good idea 👍🏻

PR #1495, opened 2022-01-25
Remove Carthage derived data folder in test:carthage CI step

calda:
This PR removes the Carthage derived data folder in the `test:carthage` CI step. This makes the test fail for me locally. Using this draft to confirm that it makes CI fail as well. (This demonstrates the issue in #1492.)


calda:
The tests fail now, as expected 👍🏻

PR #1496, opened 2022-01-25
Fix issues with Carthage build by reintroducing manually-managed Xcode project

calda:
It looks like Carthage has issues (#1492, #1495) with the setup we're using, where the `Lottie` Carthage framework re-exports an underlying `LottieCore` SPM dependency. The best way to fix this seems to be to go back to using a manually-managed Xcode project (like before #1427).


Rakefile
```diff
@@ -68,8 +68,23 @@ namespace :test do
...
+      sh 'rm -rf ~/Library/Caches/org.carthage.CarthageKit/DerivedData'
+
       # Build a test app that imports and uses the LottieCarthage framework
-      sh 'xcodebuild build -scheme CarthageTest -destination "platform=iOS Simulator,name=iPhone 8"'
+      xcodebuild('build -scheme CarthageTest -destination "platform=iOS Simulator,name=iPhone 8"')
+    end
+  end
+
+  desc 'Tests Swift Package Manager support'
+  task :spm do
```
calda, line 82: The `build:package` commands no longer build the targets generated by `Package.swift`, so this adds a new separate `test:spm` flow that validates the targets in `Package.swift`.

Lottie.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,2502 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 55;
+	objects = {
+
+/* Begin PBXBuildFile section */
```
calda: 🥲

calda: 🥲

Rakefile
```diff
@@ -68,8 +68,23 @@ namespace :test do
       # Build the LottieCarthage framework scheme
       sh 'carthage build --use-xcframeworks'
 
+      # Delete Carthage's derived data to verify that the built .xcframework doesn't depend on any
+      # side effects from building on this specific machine.
+      # https://github.com/airbnb/lottie-ios/issues/1492
+      sh 'rm -rf ~/Library/Caches/org.carthage.CarthageKit/DerivedData'
```
calda, line 74: This changes causes the test to fail with the previous project setup (#1495)

erichoracek (approved):
@calda LGTM!

PR #1504, opened 2022-01-28
Fix support for negative `transform.scale.x` values

calda:
This PR fixes support for negative `transform.scale.x` values.

For some reason, negative `transform.scale.x` values aren't applied by Core Animation. This PR fixes support for animations using this by adding a `transform.rotation.y` animation that flips the view horizontally when necessary.

| Before | After |
| ---- | ---- |
|  |  |
| bad | better |


calda:
pretty weird -- this works correctly when I test the Example app in the simulator but the snapshot test cases look broken. will see if I can fix it, but I haven't run into an issue like this before (where the test snapshots look different from what I see at runtime in the example app)

calda:
I wasn't able to find alternative implementation for this that rendered correctly in the snapshot tests, and have no clue why this wouldn't work in a headless snapshot test.

Sources/Private/Experimental/Animations/TransformAnimations.swift
```diff
@@ -151,17 +151,41 @@ extension CALayer {
...
+        //    to flip the view horizontally, which gives us the desired effect.
         CGFloat(scale.x) / 100
       },
       context: context)
 
+    // When `scale.x` is negative, we have to rotate the view
+    // half way around the y axis to flip it horizontally.
+    //  - We don't do this in snapshot tests because it breaks the tests
+    //    in surprising ways that don't happen at runtime. Definitely not ideal.
+    if !SnapshotTests.snapshotTestsAreRunning {
```
calda, line 165: This commit shows what happens when you run the snapshot tests with this behavior enabled: https://github.com/airbnb/lottie-ios/pull/1504/commits/a4a598ab0cdbbf002efcbd2d569ed19c356682d9

This works correctly at runtime in the Example app so it's unclear why this wouldn't work in the snapshot tests. I think the tests appear more broken with this code enabled than with it disabled.

erichoracek: Maybe we could print something in the `else` to the console indicating that this feature isn't going to function correctly in the snapshot tests (not respecting negative scale.x)? May help prevent someone from having to dig through the rendering implementation to figure out what's going on

Sources/Private/Experimental/Animations/TransformAnimations.swift
```diff
@@ -151,17 +151,41 @@ extension CALayer {
         // Lottie animation files express scale as a numerical percentage value
         // (e.g. 50%, 100%, 200%) so we divide by 100 to get the decimal values
         // expected by Core Animation (e.g. 0.5, 1.0, 2.0).
-        CGFloat(scale.x) / 100
+        //  - For some reason, negative `scaleX` values aren't applied by
+        //    Core Animation. Instead, we set up a `rotationY` animation below
+        //    to flip the view horizontally, which gives us the desired effect.
+        abs(CGFloat(scale.x) / 100)
```
erichoracek, line 157: Do you think it makes sense to file a radar reporting this? Perhaps it could get fixed in the future by the CA team

erichoracek (approved):
LGTM!

Sources/Private/Experimental/Animations/TransformAnimations.swift
calda: Good idea, this seems easy to create a repro case for

calda: Experimenting in a sample project, it looks like this is a consequence of how we set up a different `CAAnimation` for each component of the transform, instead of having a single `CAAnimation` for the entire transform.

For example, this animation correctly flips the layer on both vertically and horizontally:

```swift
let transform = CABasicAnimation(keyPath: "transform")
transform.duration = 1.0
transform.fromValue = CATransform3DMakeScale(-0.5, -0.5, 1)
transform.toValue = CATransform3DMakeScale(-1, -1, 1)
transform.repeatCount = .greatestFiniteMagnitude
transform.autoreverses = true
square.add(transform, forKey: "transform")
```

Screen Shot 2022-01-28 at 2 33 17 PM

but this seeming equivalent code causes the layer to only be flipped on one axis:

```swift
let scaleX = CABasicAnimation(keyPath: "transform.scale.x")
scaleX.duration = 1.0
scaleX.fromValue = -0.5
scaleX.toValue = -1
scaleX.repeatCount = .greatestFiniteMagnitude
scaleX.autoreverses = true
square.add(scaleX, forKey: "transform.scale.x")

let scaleY = CABasicAnimation(keyPath: "transform.scale.y")
scaleY.duration = 1.0
scaleY.fromValue = -0.5
scaleY.toValue = -1
scaleY.repeatCount = .greatestFiniteMagnitude
scaleY.autoreverses = true
square.add(scaleY, forKey: "transform.scale.y")
```

Screen Shot 2022-01-28 at 2 31 15 PM

calda: Got an interesting response on the radar:

> Thanks for submitting this report.  This behaves as expected.
>
> There is inherent ambiguity in a transform matrix. For example, Scale(-0.5, -1, -1) * Rotation(PI, 1, 0, 0) is the same transform matrix as Scale(-0.5, 1, 1). 
>
> If scale.y is then set to -0.5, you would not get a y-flip in the first decomposition, and that’s the “issue” you observed. There’s no “fix” for this because whichever way we decide to decompose a transform matrix for merging multiple animations, one can always find a counterexample due to the ambiguity.

erichoracek: Interesting! So I guess there's no way to specify this in a way that Core Animation will render—bummer.

PR #1506, opened 2022-01-31
Improve support for Groups with multiple Shapes

calda:
This PR improves support for `Group`s with multiple `Shape`s

Previously, we created a separate `ShapeItemLayer` for each `Shape`, which would animate independently. This causes intersecting shapes to render incorrectly compared to the main thread rendering engine, where the shapes are combined into a single `CGPath` (so `CAShapeLayerFillRule.evenOdd` applies to the intersecting areas).

Now, we combine the `Shape`s into a single `CGPath` when possible. Specifically, this is possible when all of the `Shape`s have the same keyframe timing information (e.g. when `[KeyframeGroup<Shape>]` and `KeyframeGroup<[Shape]>` are equivalent).

| Before | After |
| ----- | ----- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-01-31 at 10 25 31 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-01-31 at 10 24 10 |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-01-31 at 10 25 15 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-01-31 at 10 24 25 |


Sources/Private/Experimental/Animations/CombinedShapeAnimation.swift
```diff
@@ -0,0 +1,47 @@
+// Created by Cal Stephens on 1/28/22.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+import QuartzCore
+
+extension CAShapeLayer {
+  /// Adds animations for the given `[BezierPath]` keyframes to this `CALayer`
```
erichoracek, line 7: Does this comment need to be updated? Seems like it takes a `CombinedShapeItem` now

Sources/Private/Experimental/Layers/ShapeLayer.swift
```diff
@@ -30,15 +30,45 @@ final class ShapeLayer: BaseCompositionLayer {
...
       let otherItemsInGroup = itemsInGroup.filter { !$0.item.drawsCGPath }
 
-      for pathDrawingItem in pathDrawingItemsInGroup {
-        let sublayer = ShapeItemLayer(shape: pathDrawingItem, otherItems: otherItemsInGroup)
+      // If all of the path-drawing `ShapeItem`s have keyframes with the same timing information,
+      // we can combine the `[KeyframeGroup<BezierPath>]` (which have to animate in separate layers)
+      // into a single `KeyframeGroup<[BezierPath]>`, which can be combined into a single CGPath animation.
+      //
+      // This is how Groups with multiple path-drawing items are supposed to be rendered,
+      // because combing multiple paths into a single `CGPath` (instead of render them in separate layers)
```
erichoracek, line 41:
```suggestion
      // because combing multiple paths into a single `CGPath` (instead of rendering them in separate layers)
```

erichoracek (approved):
Good stuff!

Sources/Private/Experimental/Animations/CombinedShapeAnimation.swift
calda: this is somewhat right since a `CombinedShapeItem` is a thin wrapper around `[BezierPath]`, but I'll update this to be more clear (you're right that it looks wrong at a glance)

Sources/Private/Experimental/Layers/ShapeLayer.swift
calda: fixing this in #1507 to prevent merge conflicts, ty for the close eye

PR #1507, opened 2022-02-01
Add support for nested Group items

calda:
This PR adds support for nested `Group` items. Before, we would silently ignore (and not render) any `Group` that wasn't a top-level child of a `ShapeItemLayer`.

| Before | After |
| ----- | ---- |
|  |  |


Sources/Private/Experimental/Layers/ShapeLayer.swift
```diff
@@ -83,14 +62,92 @@ final class ShapeLayer: BaseCompositionLayer {
...
+    // recursively creating more `GroupLayer`s
+    setupGroups(from: group.items, parentGroup: group)
 
+    let nonGroupItems = group.items
+      .filter { !($0 is Group) }
+      .map { ShapeItemLayer.Item(item: $0, parentGroup: group) }
+      + inheritedItems
+
+    let pathDrawingItemsInGroup = nonGroupItems.filter { $0.item.drawsCGPath }
+    let otherItemsInGroup = nonGroupItems.filter { !$0.item.drawsCGPath }
```
erichoracek, line 89: Wondering if it might make sense to use:
```swift
Dictionary(grouping: nonGroupItems, by: { $0.item.drawsCGPath })
```
to group them items so we only do one iteration of `nonGroupItems` here. Then we could reference them as:
```swift
let pathDrawingItemsInGroup = grouped[true, default: []] 
let otherItemsInGroup = grouped[false, default: []]
```
Not sure how big `nonGroupItems` could be so it might not be worth it but it's just a thought

Sources/Private/Experimental/Layers/ShapeLayer.swift
```diff
@@ -83,14 +62,92 @@ final class ShapeLayer: BaseCompositionLayer {
...
+
+    // If all of the path-drawing `ShapeItem`s have keyframes with the same timing information,
+    // we can combine the `[KeyframeGroup<BezierPath>]` (which have to animate in separate layers)
+    // into a single `KeyframeGroup<[BezierPath]>`, which can be combined into a single CGPath animation.
+    //
+    // This is how Groups with multiple path-drawing items are supposed to be rendered,
+    // because combing multiple paths into a single `CGPath` (instead of rendering them in separate layers)
```
erichoracek, line 96:
```suggestion
    // because combining multiple paths into a single `CGPath` (instead of rendering them in separate layers)
```

erichoracek (approved):
LGTM!

Sources/Private/Experimental/Layers/ShapeLayer.swift
calda: ohoh good idea! thats a nice simplification

PR #1509, opened 2022-02-01
Fix keypaths for positionX and positionY properties

calda:
This PR fixes an issue where the keypath for the `positionX` and `positionY` layer properties were backwards, which caused some layers to be displayed in the wrong position.

We apparently don't have any animations using these properties in this repo's regression suite, which explains why it took me so long to notice this (lol).

| Before | After |
| ----- | ----- |
|  |  |


erichoracek (approved):
Oops!

PR #1372, opened 2021-07-02
Fixed GroupNode empty transform issue (#1)

sroik:
Default empty transform should match `ShapeTransform.swift`default empty state

PR #1510, opened 2022-02-01
Fixed GroupNode empty transform issue (#1372)

calda:
This PR integrates @sroik's changes from  #1372 (I tried to rebase that PR manually but messed up, classic mistake)

`1` was incorrect here -- these values are actually percentages, so `1` meant "1%"


erichoracek (approved):
LGTM!

PR #1514, opened 2022-02-09
Fix stroke Trim support in cases where strokeStart > strokeEnd

calda:
This PR fixes stroke `Trim` support in cases where `strokeStart` > `strokeEnd`.

`CAShapeLayer` requires that `strokeStart` be less than `strokeEnd`. If we detect that the value for `trim.start` is always greater than `trim.end`, we swap the values to match what `CAShapeLayer` expects.

| Before | After |
| ----- | ----- |
|  |  |


Sources/Private/Experimental/Animations/ShapeAnimation.swift
```diff
@@ -77,9 +77,24 @@ extension CAShapeLayer {
...
+    //    we swap the start/end keyframes to match what `CAShapeLayer` expects.
+    let strokeStartKeyframes: KeyframeGroup<Vector1D>
+    let strokeEndKeyframes: KeyframeGroup<Vector1D>
+    if trim.startValueIsAlwaysGreaterThanEndValue() {
+      strokeStartKeyframes = trim.end
+      strokeEndKeyframes = trim.start
+    } else {
+      strokeStartKeyframes = trim.start
+      strokeEndKeyframes = trim.end
+    }
```
erichoracek, line 93: nit: might be simpler to reason about if this method returns a tuple of the start/end keyframes:
```suggestion
    let (strokeStartKeyframes, strokeEndKeyframes) = trim.caShapeLayerKeyframes()
```

Sources/Private/Experimental/Animations/ShapeAnimation.swift
```diff
@@ -98,3 +113,28 @@ extension CAShapeLayer {
...
+  ///    `strokeStart` be less than `strokeEnd`.
+  fileprivate func startValueIsAlwaysGreaterThanEndValue() -> Bool {
+    let keyframeTimes = Set(start.keyframes.map { $0.time } + end.keyframes.map { $0.time })
+
+    let startInterpolator = KeyframeInterpolator(keyframes: start.keyframes)
+    let endInterpolator = KeyframeInterpolator(keyframes: end.keyframes)
+
+    for keyframeTime in keyframeTimes {
+      let startAtTime = startInterpolator.value(frame: keyframeTime) as! Vector1D
+      let endAtTime = endInterpolator.value(frame: keyframeTime) as! Vector1D
```
erichoracek, line 131: nit: include a comment on why this force cast is safe?

erichoracek (approved):
LGTM!

Sources/Private/Experimental/Animations/ShapeAnimation.swift
calda: oh whoops I meant to clean this up, ty for the reminder

Sources/Private/Experimental/Animations/ShapeAnimation.swift
calda: ooohh good idea

PR #1518, opened 2022-02-16
Add performance tests, improve performance of Core Animation rendering engine

calda:
This PR adds performance tests that compare the performance of the Core Animation rendering engine and the Main Thread rendering engine. These tests show that the Core Animation engine is more expensive to setup, but significantly more performant once setup.

As a part of this PR I also made some optimizations to the `CAAnimation` setup, which led to some pretty big performance wins. For example, these optimizations improved the performance of setting up a complex animation in the new engine to be ~1.6x more expensive than the old engine, compared to a baseline of ~3x more expensive than the old engine prior to the optimizations.


Tests/AnimationKeypathTests.swift
```diff
@@ -3,7 +3,7 @@
 
 import XCTest
 
-@testable import LottieCore
+@testable import Lottie
```
calda, line 6: This file accidentally wasn't included in the Xcode project, so wasn't being built / run on master

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  func testAnimationViewScrubbing_simpleAnimation() {
+    let ratio = compareEngineScrubbingPerformance(for: simpleAnimation, iterations: 1000)
+    XCTAssertEqual(ratio, 0.05, accuracy: 0.02)
+  }
+
+  func testAnimationViewScrubbing_complexAnimation() {
+    let ratio = compareEngineScrubbingPerformance(for: complexAnimation, iterations: 1000)
+    XCTAssertEqual(ratio, 0.002, accuracy: 0.001)
```
calda, line 37: I was pretty impressed with this number -- interactively scrubbing the current frame of a complex animation in the new rendering engine is up to 500x faster than in the main thread engine

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+
+    // This is basically a snapshot test for the performance of the Core Animation engine
+    // compared to the Main Thread engine. Currently, the Core Animation engine is
+    // about 1.6x slower than the Main Thread engine in this example.
+    XCTAssertEqual(ratio, 1.6, accuracy: 0.2)
+  }
+
+  func testAnimationViewSetup_complexAnimation() {
+    let ratio = compareEngineSetupPerformance(for: complexAnimation, iterations: 500)
+    XCTAssertEqual(ratio, 1.75, accuracy: 0.3)
```
calda, line 27: Before the optimizations elsewhere in this PR, this number was `3.0` -- so these optimizations made quite the difference for complex animations!

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+  // MARK: Internal
+
+  func testAnimationViewSetup_simpleAnimation() {
+    // Compare the performance of displaying this simple animation in the two animation engines
+    let ratio = compareEngineSetupPerformance(for: simpleAnimation, iterations: 2000)
+
+    // This is basically a snapshot test for the performance of the Core Animation engine
+    // compared to the Main Thread engine. Currently, the Core Animation engine is
+    // about 1.6x slower than the Main Thread engine in this example.
+    XCTAssertEqual(ratio, 1.6, accuracy: 0.2)
```
calda, line 22: Before the optimizations elsewhere in this PR, this number was ~`2.2`

Tests/PerformanceTests.swift
calda: (not too surprising, since to scrub the Core Animation engine you just have to update the `timeOffset` of the root layer)

Tests/PerformanceTests.swift
calda, line 22: bumping this to give more leeway in CI

```suggestion
    XCTAssertEqual(ratio, 1.6, accuracy: 0.3)
```

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  func testAnimationViewScrubbing_simpleAnimation() {
+    let ratio = compareEngineScrubbingPerformance(for: simpleAnimation, iterations: 1000)
+    XCTAssertEqual(ratio, 0.05, accuracy: 0.02)
+  }
+
+  func testAnimationViewScrubbing_complexAnimation() {
+    let ratio = compareEngineScrubbingPerformance(for: complexAnimation, iterations: 1000)
+    XCTAssertEqual(ratio, 0.002, accuracy: 0.001)
```
calda, line 37: have to bump the threshold for this since I got `0.148` at least once when running this in CI (I was getting more like `0.002` when running this locally, and got that at least once in CI, but we can't have flaky tests)

```suggestion
    XCTAssertEqual(ratio, 0.1, accuracy: 0.1)
```

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  func testAnimationViewSetup_complexAnimation() {
+    let ratio = compareEngineSetupPerformance(for: complexAnimation, iterations: 500)
+    XCTAssertEqual(ratio, 1.75, accuracy: 0.3)
+  }
+
+  func testAnimationViewScrubbing_simpleAnimation() {
+    let ratio = compareEngineScrubbingPerformance(for: simpleAnimation, iterations: 1000)
+    XCTAssertEqual(ratio, 0.05, accuracy: 0.02)
```
calda, line 32: https://github.com/airbnb/lottie-ios/pull/1518/files#r808573829

```suggestion
    XCTAssertEqual(ratio, 0.1, accuracy: 0.1)
```

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+  // MARK: Internal
+
+  func testAnimationViewSetup_simpleAnimation() {
+    // Compare the performance of displaying this simple animation in the two animation engines
+    let ratio = compareEngineSetupPerformance(for: simpleAnimation, iterations: 2000)
+
+    // This is basically a snapshot test for the performance of the Core Animation engine
+    // compared to the Main Thread engine. Currently, the Core Animation engine is
+    // about 1.6x slower than the Main Thread engine in this example.
+    XCTAssertEqual(ratio, 1.6, accuracy: 0.3)
```
calda, line 22: Before the optimizations elsewhere in this PR, this number was ~`2.2`

Sources/Private/Experimental/Animations/CALayer+addAnimation.swift
```diff
@@ -39,11 +39,48 @@ extension CALayer {
...
+      if keyframeValue == property.defaultValue {
+        return nil
+      }
+
+      // If the property on the CALayer being animated hasn't been modified from the default yet,
+      // then we can apply the keyframe value directly to the layer using KVC instead
+      // of creating a `CAAnimation`. (For example, this check prevents us from using
+      // KVC to change the `anchorPoint`, which conflicts with where we set the `anchorPoint`
+      // for our layers to a non-default value in `BaseAnimationLayer` / `fillBoundsOfSuperlayer()`)
+      let currentValue = value(forKey: property.caLayerKeypath) as? ValueRepresentation
```
calda, line 68: dropping a note to remember to check this before merging -- since KVC lookups are somewhat expensive, we may want to check if `property.defaultValue` was defined (e.g. is not `nil`) before doing the KVC lookup. I think that shouldn't affect any runtime behaviors, but would just lets us avoid KVC lookups for conditions that are going to fail anyway.

calda: cool, I made this change in https://github.com/airbnb/lottie-ios/pull/1518/commits/e1f7a758d7fee38fd8b8e014a114b30688559ab4 and it improved performance a bit more

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  func testAnimationViewSetup_complexAnimation() {
+    let ratio = compareEngineSetupPerformance(for: complexAnimation, iterations: 500)
+    XCTAssertEqual(ratio, 1.6, accuracy: 0.3)
```
calda, line 27: Before the optimizations elsewhere in this PR, this number was `3.0` -- so these optimizations made quite the difference for complex animations!

I've seen this be as low as `1.4` locally, but I kept the tolerance a bit high to give leeway in CI.

calda: haha oh wow I just got 1.2 in CI? I probably need to increase the number of iterations here to make this more consistent

Tests/PerformanceTests.swift
calda: fascinating -- I noticed that memory usage went up continuously while running these tests because the views / layers weren't being deallocated. I fixed that issue so memory stays flat the entire time, and now I'm seeing that the Core Animation engine actually runs _faster_ than the Main Thread engine now (getting `0.7x`). That feels almost too good to be true, so I'll need to look at this in more detail tomorrow.

Sources/Private/Experimental/Animations/LayerProperty.swift
```diff
@@ -10,10 +10,16 @@ import QuartzCore
...
 /// and https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreAnimation_guide/Key-ValueCodingExtensions/Key-ValueCodingExtensions.html
-struct LayerProperty<ValueRepresentation> {
+struct LayerProperty<ValueRepresentation: Equatable> {
   /// The `CALayer` KVC key path that this value should be assigned to
   let caLayerKeypath: String
 
+  /// The default value of this property on a `CALayer`
+  ///  - If the keyframe values are just equal to the default value,
+  ///    then we can improve performance a bit by just not creating
+  ///    a CAAnimation (since it would be redundant).
```
erichoracek, line 20:
```suggestion
  ///    a `CAAnimation` (since it would be redundant).
```

Tests/PerformanceTests.swift
erichoracek: Let me know if you'd like for me to try to run these tests on my M1X, should hopefully help get another example to fine tune the accuracy

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+  private let complexAnimation = Animation.named(
+    "LottieLogo2",
+    bundle: .module,
+    subdirectory: "Samples")!
+
+  /// Compares initializing the given animation with both the Main Thread and Core Animation engine,
+  /// and returns the ratio of how much slower the Core Animation is than the Main Thread engine
+  private func compareEngineSetupPerformance(for animation: Animation, iterations: Int) -> Double {
+    let mainThreadEnginePerformance = measurePerformance {
+      for _ in 0 ..< iterations {
```
erichoracek, line 56: I believe our style guide recommends not having a space with range operators
```suggestion
      for _ in 0..<iterations {
```

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+    let coreAnimationEnginePerformance = measurePerformance {
+      for i in 0 ..< iterations {
+        coreAnimationView.currentProgress = Double(i) / Double(iterations)
+      }
+    }
+
+    print("Core Animation engine took \(coreAnimationEnginePerformance) seconds")
+
+    let ratio = coreAnimationEnginePerformance / mainThreadEnginePerformance
+    print("Core Animation engine is \(ratio)x slower than the Main Thread engine")
```
erichoracek, line 103: These logs presupposes a ratio difference, perhaps we could use something like
```suggestion
    print("Core Animation engine took \(ratio)x as long as the Main Thread engine")
```

Sources/Private/Experimental/Animations/CALayer+addAnimation.swift
```diff
@@ -39,11 +39,49 @@ extension CALayer {
...
+      }
+
+      // Otherwise, we still need to create a `CAAnimation`, but we can
+      // create a simple `CABasicAnimation` that is still less expensive
+      // than computing a `CAKeyframeAnimation`.
+      let animation = CABasicAnimation(keyPath: property.caLayerKeypath)
+      animation.fromValue = keyframeValue
+      animation.toValue = keyframeValue
+      return animation
+    }
```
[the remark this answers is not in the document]
calda: sure, good idea

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+    let coreAnimationEnginePerformance = measurePerformance {
+      for _ in 0 ..< iterations {
+        setupAnimationView(with: animation, configuration: .init(renderingEngine: .coreAnimation))
+      }
+    }
+
+    print("Core Animation engine took \(coreAnimationEnginePerformance) seconds")
+
+    let ratio = coreAnimationEnginePerformance / mainThreadEnginePerformance
+    print("Core Animation engine is \(ratio)x slower than the Main Thread engine")
```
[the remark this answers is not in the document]
calda: good idea

Sources/Private/Experimental/Animations/LayerProperty.swift
```diff
@@ -51,60 +57,70 @@ extension LayerProperty {
   static var position: LayerProperty<CGPoint> {
     .init(
       caLayerKeypath: "transform.translation",
+      defaultValue: CGPoint(x: 0, y: 0),
```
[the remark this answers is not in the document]
calda: Interesting idea. Some immediate thoughts:
 - we can't create a collection of these since they all have different generic types (unless we create a type erased wrapped), so a test would probably have to test each one manually
  - the snapshot tests fail if you change these to an incorrect value, so we do currently have a layer of tests validating that they work as expected

Tests/PerformanceTests.swift
```diff
@@ -0,0 +1,124 @@
...
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+import Foundation
+import XCTest
+
+@testable import Lottie
+
+// MARK: - PerformanceTests
+
+final class PerformanceTests: XCTestCase {
```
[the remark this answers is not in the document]
calda: Good idea, I never tried the `measure` API because it doesn't return values that you could compare. It seems like a good idea to have both so you can use the `measure` ones locally, will take a look.

PR #1520, opened 2022-02-18
Improve performance by batching animation setup to next display() call

calda:
This PR improves the performance of the Core Animation engine by batching animation setup calls to be deferred until the next `display()` call. This prevents us from setting up multiple animations in the same run loop cycle (e.g. during view setup), which causes wasted work.

### DIFF
--- a/Sources/Private/Experimental/ExperimentalAnimationLayer.swift
+++ b/Sources/Private/Experimental/ExperimentalAnimationLayer.swift
@@ -56,6 +56,13 @@ final class ExperimentalAnimationLayer: BaseAnimationLayer {
     var timeOffset: TimeInterval = 0
   }
 
+  enum PlaybackState: Equatable {
+    /// The animation is playing in real-time
+    case playing
+    /// The animation is statically displaying a specific frame
+    case paused(frame: AnimationFrameTime)
+  }
+
   /// The `AnimationImageProvider` that `ImageLayer`s use to retrieve images,
   /// referenced by name in the animation json.
   var imageProvider: AnimationImageProvider {
@@ -68,43 +75,20 @@ final class ExperimentalAnimationLayer: BaseAnimationLayer {
     didSet { reloadFonts() }
   }
 
-  /// Sets up `CAAnimation`s for each `AnimationLayer` in the layer hierarchy,
-  /// playing from `currentFrame`.
-  func setupAnimation(
+  /// Queues the animation with the given timing configuration
+  /// to begin playing at the next `display()` call.
+  ///   - This batches together animations so that even if `playAnimation`
+  ///     is called multiple times in the same run loop cycle, the animation will only be
+  func playAnimation(
     context: AnimationContext,
-    timingConfiguration: CAMediaTimingConfiguration)
+    timingConfiguration: CAMediaTimingConfiguration,
+    playbackState: PlaybackState = .playing)
   {
-    // Remove any existing animations from the layer hierarchy
-    removeAnimations()
+    pendingAnimationConfiguration = (
+      animationConfiguration: .init(animationContext: context, timingConfiguration: timingConfiguration),
+      playbackState: playbackState)
 
-    playbackState = .playing
-
-    currentAnimationConfiguration = AnimationConfiguration(
-      animationContext: context,
-      timingConfiguration: timingConfiguration)
-
-    let layerContext = LayerAnimationContext(
-      animation: animation,
-      timingConfiguration: timingConfiguration,
-      startFrame: context.playFrom,
-      endFrame: context.playTo,
-      valueProviderStore: valueProviderStore,
-      currentKeypath: AnimationKeypath(keys: []))
-
-    // Perform a layout pass if necessary so all of the sublayers
-    // have the most up-to-date sizing information
-    layoutIfNeeded()
-
-    // Set the speed of this layer, which will be inherited
-    // by all sublayers and their animations.
-    //  - This is required to support scrubbing with a speed of 0
-    speed = timingConfiguration.speed
-
-    // Setup a placeholder animation to let us track the realtime animation progress
-    setupPlaceholderAnimation(context: layerContext)
-
-    // Set up the new animations with the current `TimingConfiguration`
-    setupAnimations(context: layerContext)
+    setNeedsDisplay()
   }
 
   override func layoutSublayers() {
@@ -113,10 +97,21 @@ final class ExperimentalAnimationLayer: BaseAnimationLayer {
     // If no animation has been set up yet, display the first frame
     // now that the layer hierarchy has been setup and laid out
     if
+      pendingAnimationConfiguration == nil,
       currentAnimationConfiguration == nil,
       bounds.size != .zero
     {
       currentFrame = animation.frameTime(forProgress: animationProgress)
+    }
+  }
+
+  override func display() {
+    super.display()
+
+    if let pendingAnimationConfiguration = pendingAnimationConfiguration {
+      self.pendingAnimationConfiguration = nil
+      setupAnimation(for: pendingAnimationConfiguration.animationConfiguration)
+      currentPlaybackState = pendingAnimationConfiguration.playbackState
     }
   }
 
@@ -127,14 +122,13 @@ final class ExperimentalAnimationLayer: BaseAnimationLayer {
     let timingConfiguration: CAMediaTimingConfiguration
   }
 
-  private enum PlaybackState: Equatable {
-    /// The animation is playing in real-time
-    case playing
-    /// The animation is statically displaying a specific frame
-    case paused(frame: AnimationFrameTime)
-  }
+  /// The configuration for the most recent animation which has been
+  /// queued by calling `playAnimation` but not yet actually set up
+  private var pendingAnimationConfiguration: (
+    animationConfiguration: AnimationConfiguration,
+    playbackState: PlaybackState)?
 
-  // Configuration for the animation that is currently setup in this layer
+  /// Configuration for the animation that is currently setup in this layer
   private var currentAnimationConfiguration: AnimationConfiguration?
 
   /// The current progress of the placeholder `CAAnimation`,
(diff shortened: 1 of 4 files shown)


Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -56,6 +56,13 @@ final class ExperimentalAnimationLayer: BaseAnimationLayer {
     var timeOffset: TimeInterval = 0
   }
 
+  enum PlaybackState: Equatable {
```
calda, line 59: made this internal so I could add it as an argument of `playAnimation`

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -176,6 +175,39 @@ final class ExperimentalAnimationLayer: BaseAnimationLayer {
       context: layerContext)
   }
 
+  /// Immediately builds and begins playing `CAAnimation`s for each sublayer
+  private func setupAnimation(for configuration: AnimationConfiguration) {
```
calda, line 179: This is now only called a maximum of once per run loop cycle, from `display()`

Sources/Public/Animation/AnimationView.swift
```diff
@@ -123,10 +125,24 @@ final public class AnimationView: AnimationViewBase {
...
+   the Core Animation engine does not have any CPU overhead.
    */
-  public var backgroundBehavior: LottieBackgroundBehavior = .pause
+  public lazy var backgroundBehavior: LottieBackgroundBehavior = {
+    switch configuration.renderingEngine {
+    case .mainThread:
+      return .pause
+
+    case .coreAnimation:
+      return .doNothing
```
calda, line 143: We don't need to pause the Core Animation engine when it moves into the background, since it doesn't introduce any CPU overhead. This prevents us from having to `stop()` and then `start()` the animation, which avoids unnecessary animation setup passes.

Sources/Public/Animation/AnimationView.swift
```diff
@@ -259,7 +275,7 @@ final public class AnimationView: AnimationViewBase {
    */
   public var currentFrame: AnimationFrameTime {
     set {
-      removeCurrentAnimation()
+      removeCurrentAnimationIfNecessary()
```
calda, line 278: Calling `removeCurrentAnimation()` before `updateAnimationFrame(...)` is unnecessary for the Core Animation engine, but causes unnecessary animation setup passes (since it sets the `currentFrame` to `0`)

Sources/Public/Animation/AnimationView.swift
```diff
@@ -803,12 +819,22 @@ final public class AnimationView: AnimationViewBase {
...
-      CATransaction.setAnimationTimingFunction(CAMediaTimingFunction(name: .linear))
-      animationLayer.position = position
-      animationLayer.transform = xform
-      CATransaction.commit()
+      // In performance tests, we have to wrap the animation view setup
+      // in a `CATransaction` in order for the layers to be deallocated at
+      // the correct time. The `CATransaction`s in this method interfere
+      // with the ones managed by the performance test, and aren't actually
+      // necessary in a headless environment, so we disable them.
+      if TestHelpers.performanceTestsAreRunning {
```
calda, line 827: I also had to do this in https://github.com/airbnb/lottie-ios/pull/1518/files#diff-b8031f1d50b5687c758122fa8aa487e09cb9068eee0db0d2ecff606ba944e9d3R834. I must have missed this on the last PR, or changes the code path just enough to start hitting this in the performance tests.

Tests/PerformanceTests.swift
```diff
@@ -19,25 +19,25 @@ final class PerformanceTests: XCTestCase {
     // This is basically a snapshot test for the performance of the Core Animation engine
     // compared to the Main Thread engine. Currently, the Core Animation engine is
     // about the same speed as the Main Thread engine in this example.
-    XCTAssertEqual(ratio, 1.0, accuracy: 0.15)
+    XCTAssertEqual(ratio, 0.95, accuracy: 0.15)
```
calda, line 22: Now the new engine actually performs BETTER than the old engine, amazing

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -68,43 +75,20 @@ final class ExperimentalAnimationLayer: BaseAnimationLayer {
     didSet { reloadFonts() }
   }
 
-  /// Sets up `CAAnimation`s for each `AnimationLayer` in the layer hierarchy,
-  /// playing from `currentFrame`.
-  func setupAnimation(
+  /// Queues the animation with the given timing configuration
+  /// to begin playing at the next `display()` call.
+  ///   - This batches together animations so that even if `playAnimation`
+  ///     is called multiple times in the same run loop cycle, the animation will only be
```
erichoracek, line 81:
```suggestion
  ///     is called multiple times in the same run loop cycle, the animation will only be
  ///     setup to play once.
```

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
erichoracek: Nice!

Sources/Private/Experimental/ExperimentalAnimationLayer.swift
```diff
@@ -242,13 +274,22 @@ extension ExperimentalAnimationLayer: RootAnimationLayer {
           closure: nil),
         timingConfiguration: CAMediaTimingConfiguration(speed: 0))
 
-      if currentAnimationConfiguration != requiredAnimationConfiguration {
-        setupAnimation(
+      if pendingAnimationConfiguration != nil {
+        pendingAnimationConfiguration = (
+          animationConfiguration: requiredAnimationConfiguration,
+          playbackState: .paused(frame: newValue))
```
erichoracek, line 280: Do we not need a `setNeedsDisplay` here to flush this `pendingAnimationConfiguration`? Might be helpful to have an explanatory comment

Sources/Private/LayerContainers/AnimationContainer.swift
```diff
@@ -134,7 +134,9 @@ final class AnimationContainer: CALayer, RootAnimationLayer {
     if respectAnimationFrameRate {
       newFrame = floor(newFrame)
     }
-    animationLayers.forEach( { $0.displayWithFrame(frame: newFrame, forceUpdates: false) })
+    animationLayers.forEach {
+      $0.displayWithFrame(frame: newFrame, forceUpdates: false)
+    }
```
erichoracek, line 139: nit: `forEach` feels non-idiomatic here, especially when used with `$0`
```suggestion
    for layer in animationLayers {
      layer.displayWithFrame(frame: newFrame, forceUpdates: false)
    }
```

Sources/Public/Animation/AnimationView.swift
```diff
@@ -20,6 +20,8 @@ public enum LottieBackgroundBehavior {
   case pauseAndRestore
   /// Stops the animation and sets it to the end of its current play time. The completion block is called.
   case forceFinish
+  /// Does nothing when the animation moves to the background
+  case doNothing
```
erichoracek, line 24: Can you clarify that this is the default for the core animation engine? And then can you clarify that in the `pause` case with the core animation rendering engine that it would introduce unnecessary overhead? I think it would be helpful to have more discussion in these to help users make sensible choices from a performance perspective

Tests/PerformanceTests.swift
```diff
@@ -70,7 +70,8 @@ final class PerformanceTests: XCTestCase {
         // the layers aren't deallocated until the end of the test run,
         // which causes memory usage to grow unbounded.
         CATransaction.begin()
-        setupAnimationView(with: animation, configuration: .init(renderingEngine: .mainThread))
+        let animationView = setupAnimationView(with: animation, configuration: .init(renderingEngine: .mainThread))
+        animationView.animationLayer!.display()
```
erichoracek, line 74: Maybe add a comment here and below about how this line is required to trigger the layer to perform its pending setup?

Sources/Public/Animation/AnimationView.swift
```diff
@@ -123,10 +125,24 @@ final public class AnimationView: AnimationViewBase {
...
-   The default is `pause` which pauses the animation when the application moves to
-   the background. The completion block is called with `false` for completed.
+   The default for the Main Thread animation engine is `pause`,
+   which pauses the animation when the application moves to
+   the background. This prevents the animation from consuming CPU
+   resources when not on-screen. The completion block is called with
+   `false` for completed.
+
+   The default for the Core Animation engine is `doNothing`, since
+   the Core Animation engine does not have any CPU overhead.
```
erichoracek, line 135: Does this mean that the animation will start up again automatically? Or is the implicit behavior of core animation to pause when offscreen (from what I remember that's the case?), and maybe we shouldn't have the `doNothing` case and instead just handle the `pause` state like:
```swift
case .pause:
  switch configuration.renderingEngine {
  case .main:
    removeCurrentAnimation()
  case .coreAnimation:
    break
  }
```

erichoracek (approved):
@calda LGTM! Just a few suggestions/questions

Sources/Public/Animation/AnimationView.swift
calda: The default behavior of Core Animation is to remove the animation when the app is backgrounded. That doesn't work for the new engine, since the view is _only_ rendered via the animations (including when paused), so we set `isRemovedOnCompletion = false` on all of the animations. By default for animations playing in realtime, this causes the animation to appear as it it was playing continuously while the app was in the background (e.g. the layer's "current time" continues to advance).

This is semantically different from all of the other existing background modes, so I think we need a new option for it.

erichoracek: Got it, makes a lot of sense, thanks! Some of this discussion would be helpful to have in the documentation comments I think

calda: Updated and added more extensive docs in https://github.com/airbnb/lottie-ios/pull/1520/commits/0dec30ff550f9dbae219f4b5c6c488fed9f9fb7c

PR #1521, opened 2022-02-18
Update airbnb.swiftformat to latest version from airbnb/swift

calda:
This PR updates `airbnb.swiftformat` to the [latest version](https://github.com/airbnb/swift/blob/master/resources/airbnb.swiftformat) from the [airbnb/swift](https://github.com/airbnb/swift) repo, and re-runs the linter

### DIFF
--- a/Sources/Private/Model/Keyframes/KeyframeGroup.swift
+++ b/Sources/Private/Model/Keyframes/KeyframeGroup.swift
@@ -9,13 +9,11 @@ import Foundation
 
 // MARK: - KeyframeGroup
 
-/**
- Used for coding/decoding a group of Keyframes by type.
-
- Keyframe data is wrapped in a dictionary { "k" : KeyframeData }.
- The keyframe data can either be an array of keyframes or, if no animation is present, the raw value.
- This helper object is needed to properly decode the json.
- */
+/// Used for coding/decoding a group of Keyframes by type.
+///
+/// Keyframe data is wrapped in a dictionary { "k" : KeyframeData }.
+/// The keyframe data can either be an array of keyframes or, if no animation is present, the raw value.
+/// This helper object is needed to properly decode the json.
 
 final class KeyframeGroup<T> {
 
@@ -50,20 +48,18 @@ extension KeyframeGroup: Decodable where T: Decodable {
       /// Try to decode raw value; No keyframe data.
       self.init(keyframes: [Keyframe<T>(keyframeData)])
     } else {
-      /**
-       Decode and array of keyframes.
-
-       Body Movin and Lottie deal with keyframes in different ways.
-
-       A keyframe object in Body movin defines a span of time with a START
-       and an END, from the current keyframe time to the next keyframe time.
-
-       A keyframe object in Lottie defines a singular point in time/space.
-       This point has an in-tangent and an out-tangent.
-
-       To properly decode this we must iterate through keyframes while holding
-       reference to the previous keyframe.
-       */
+      /// Decode and array of keyframes.
+      ///
+      /// Body Movin and Lottie deal with keyframes in different ways.
+      ///
+      /// A keyframe object in Body movin defines a span of time with a START
+      /// and an END, from the current keyframe time to the next keyframe time.
+      ///
+      /// A keyframe object in Lottie defines a singular point in time/space.
+      /// This point has an in-tangent and an out-tangent.
+      ///
+      /// To properly decode this we must iterate through keyframes while holding
+      /// reference to the previous keyframe.
 
       var keyframesContainer = try container.nestedUnkeyedContainer(forKey: .keyframeData)
       var keyframes = ContiguousArray<Keyframe<T>>()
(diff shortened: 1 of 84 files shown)


Sources/Private/Model/Keyframes/KeyframeGroup.swift
```diff
@@ -50,20 +48,18 @@ extension KeyframeGroup: Decodable where T: Decodable {
...
+      /// Body Movin and Lottie deal with keyframes in different ways.
+      ///
+      /// A keyframe object in Body movin defines a span of time with a START
+      /// and an END, from the current keyframe time to the next keyframe time.
+      ///
+      /// A keyframe object in Lottie defines a singular point in time/space.
+      /// This point has an in-tangent and an out-tangent.
+      ///
+      /// To properly decode this we must iterate through keyframes while holding
+      /// reference to the previous keyframe.
```
erichoracek, line 62: nit: this shouldn't be a doc comment right?
```suggestion
      // Decode and array of keyframes.
      //
      // Body Movin and Lottie deal with keyframes in different ways.
      //
      // A keyframe object in Body movin defines a span of time with a START
      // and an END, from the current keyframe time to the next keyframe time.
      //
      // A keyframe object in Lottie defines a singular point in time/space.
      // This point has an in-tangent and an out-tangent.
      //
      // To properly decode this we must iterate through keyframes while holding
      // reference to the previous keyframe.
```

Sources/Private/Model/Objects/Transform.swift
```diff
@@ -13,10 +13,8 @@ final class Transform: Codable {
   // MARK: Lifecycle
 
   required init(from decoder: Decoder) throws {
-    /**
-     This manual override of decode is required because we want to throw an error
-     in the case that there is not position data.
-     */
+    /// This manual override of decode is required because we want to throw an error
+    /// in the case that there is not position data.
```
erichoracek, line 17:
```suggestion
    // This manual override of decode is required because we want to throw an error
    // in the case that there is not position data.
```

Sources/Private/NodeRenderSystem/NodeProperties/ValueProviders/KeyframeInterpolator.swift
```diff
@@ -128,19 +126,17 @@ final class KeyframeInterpolator<ValueType>: ValueProvider where ValueType: AnyI
...
+    /// that contain the current time.
+    ///
+    /// We could use Array.first(where:) but that would search through the entire array
+    /// each frame.
+    /// Instead we track the last used index and search either forwards or
+    /// backwards from there. This reduces the iterations and complexity from
+    ///
+    /// O(n), where n is the length of the sequence to
+    /// O(n), where n is the number of items after or before the last used index.
+    ///
```
erichoracek, line 139:
```suggestion
    // This function searches through the array to find the span of two keyframes
    // that contain the current time.
    //
    // We could use Array.first(where:) but that would search through the entire array
    // each frame.
    // Instead we track the last used index and search either forwards or
    // backwards from there. This reduces the iterations and complexity from
    //
    // O(n), where n is the length of the sequence to
    // O(n), where n is the number of items after or before the last used index.
    //
```

Sources/Private/NodeRenderSystem/Nodes/RenderNodes/StrokeNode.swift
```diff
@@ -136,17 +136,16 @@ extension Array where Element == DashElement {
...
-  {
-    var dashPatterns = ContiguousArray<ContiguousArray<Keyframe<Vector1D>>>()
-    var dashPhase = ContiguousArray<Keyframe<Vector1D>>()
-    for dash in self {
-      if dash.type == .offset {
-        dashPhase = dash.value.keyframes
-      } else {
-        dashPatterns.append(dash.value.keyframes)
+    dashPhase: ContiguousArray<Keyframe<Vector1D>>) {
+      var dashPatterns = ContiguousArray<ContiguousArray<Keyframe<Vector1D>>>()
```
erichoracek, line 140: This is tough to read, could we maybe make a typealias for this type?

erichoracek (approved):
@calda Good stuff! Just a few small nits/suggestions

Sources/Private/Model/Keyframes/KeyframeGroup.swift
calda: Yep, good catch (SwiftFormat did this because it saw the leading `/**`)

Sources/Private/NodeRenderSystem/Nodes/RenderNodes/StrokeNode.swift
calda: sure

calda: I wonder why SwiftFormat indented this property, must be an obscure implementation bug that I haven't hit before

PR #1522, opened 2022-02-18
Reduce number of CI jobs

calda:
CI for PRs is currently taking around 40 minutes. Almost all of this is just queueing for individual jobs to start. To alleviate this a bit, I think we can reduce the number of individual jobs.


.github/workflows/main.yml
```diff
@@ -27,17 +27,9 @@ jobs:
   build-example:
     name: "Build Example App"
     runs-on: macos-latest
-    strategy:
-      matrix:
-        xcode:
-        - '12.4' # Swift 5.3
-        - '12.5.1' # Swift 5.4
-        - '13.2' # Swift 5.5
```
calda, line 35: I don't think we need to verify that the Example app compiles on multiple Swift versions, since it's not an artifact that we ship

.github/workflows/main.yml
```diff
@@ -100,13 +92,4 @@ jobs:
...
-
-  swiftformat:
-    name: "Lint Swift (SwiftFormat)"
-    runs-on: macos-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/actions/setup
-      - name: Lint Swift using SwiftFormat
-        run: bundle exec rake lint:swift:swiftformat
+        run: bundle exec rake lint:swift
```
calda, line 95: I combined the two linters into a single job again, since now the test job is that one that takes the longest to run once started

calda:
Technically I think we could also remove the `Build Package (13.2)` job since that job failing would case the `Test Package` and `Build Example App` jobs (and probably also the Cocoapods / Carthage / SPM jobs) to fail.

erichoracek (approved):
LGTM!

PR #1354, opened 2021-05-18
Fix repeated nil host ... allowsSpecificHTTPSCertificateForHost warnings

MichaelTransloc:
This is in response to this issue:
https://github.com/airbnb/lottie-ios/issues/1256

Many people seem to be complaining about this. It's a quick and easy fix provided by a comment on that ticket.

PR #1524, opened 2022-02-24
Fix for nil host warning when data url is base64 encoded.

### LINKED ISSUE #1256: Repeated "nil host used in call to allowsSpecificHTTPSCertificateForHost" in Log

## This issue is a:
- [] Crashing Bug
- [] Feature Request

Lottie 3.0

## What Platform are you on? 

- [] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [] Objective-C

## Expected Behavior 

Loads animation without issues.

## Actual Behavior

Loads animation without issues, but in the debug logs, this is what it spams me with 18 times:
```
2020-09-13 04:45:20.335076-0400 AppName[8898:1527281] nil host used in call to allowsSpecificHTTPSCertificateForHost
2020-09-13 04:45:20.335111-0400 AppName[8898:1527281] nil host used in call to allowsAnyHTTPSCertificateForHost:
```

## Code Example
In Collection View Cell:
```
var animationView = AnimationView()
```

In cellForItemAt:
```
cell.animationView.animation = Animation.named("myMandela")
cell.animationView.layer.masksToBounds = true
cell.animationView.contentMode = .scaleAspectFill
cell.animationView.play()
cell.animationView.loopMode = .loop
cell.animationView.backgroundBehavior = .pauseAndRestore
```

When I remove `cell.animationView.animation = Animation.named("myMandela")`, everything works fine—no errors in log.

This happens on iOS 14 beta physical device & simulator using Xcode 12 Beta.

## Animation JSON

The JSON file is enormous. I will email it.

NicholasMata:
This commit fixes the following warnings
```
nil host used in call to allowsSpecificHTTPSCertificateForHost
nil host used in call to allowsAnyHTTPSCertificateForHost:
```

__This only fixes the issue when data url uses Base64 encoded data which is the most common__ *(As without Base64 encoding the string is much larger.)*

This PR also moves converting from `ImageAsset` to `Data` to be non platform specific.

However, I did write a test case for a data url that doesn't use Base64 encoding (`testValidDataURLWithoutBase64`) this will still print the nil host warnings. This test can be used in the future to try and avoid nil host warnings for non base64 encoded data.

The fix and implementation is based on the following [Data URL Definition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) 

This fixes #1256 and applies comments specified in #1354 by @calda

### DIFF
--- a/Tests/DataURLTests.swift
+++ b/Tests/DataURLTests.swift
@@ -12,11 +12,11 @@ import XCTest
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
 final class DataURLTests: XCTestCase {
 
-  let red5x5 = "%89%50%4e%47%0d%0a%1a%0a%00%00%00%0d%49%48%44%52%00%00%00%05%00%00%00%05%08%06%00%00%00%8d%6f%26%e5%00%00%00%12%49%44%41%54%78%da%63%fc%cf%c0%00%44%a8%80%91%06%82%00%5c%65%09%fc%86%fe%00%b0%00%00%00%00%49%45%4e%44%ae%42%60%82"
+  let red5x5 =
+    "%89%50%4e%47%0d%0a%1a%0a%00%00%00%0d%49%48%44%52%00%00%00%05%00%00%00%05%08%06%00%00%00%8d%6f%26%e5%00%00%00%12%49%44%41%54%78%da%63%fc%cf%c0%00%44%a8%80%91%06%82%00%5c%65%09%fc%86%fe%00%b0%00%00%00%00%49%45%4e%44%ae%42%60%82"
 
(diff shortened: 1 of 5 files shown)


Tests/DataURLTests.swift
```diff
@@ -0,0 +1,64 @@
...
+import Foundation
+import XCTest
+
+@testable import Lottie
+
+// MARK: - DataURLTests
+
+// Tests are based on implementation found here
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
+final class DataURLTests: XCTestCase {
```
calda, line 13: Thank you for these tests!

calda (approved):
Thank you!

PR #1525, opened 2022-03-08
Fix issue where animations with a startTime greater than 0 would animate with incorrect duration

calda:
This PR fixes an issue where animations with a `startTime` greater than 0 would animate with an incorrect duration, causing it to appear broken in some circumstances:

This specific animation has a `startTime` of ~200:

| Before | After |
| ---- | ---- |
|  |  |


erichoracek (approved):
LGTM!

PR #1553, opened 2022-05-02
Add supporting infrastructure for `RenderingEngineOption.automatic`

calda:
This PR adds supporting infrastructure for a new `RenderingEngineOption.automatic`, which will default to using the Core Animation engine and fall-back to using the Main Thread engine for animations that use features which are known to be unsupported in the CA engine. Hopefully, we'll be able to make `RenderingEngineOption.automatic` the default for all consumers at some point in the future.

This PR doesn't actually implement any logic in the `Animation.supportedByCoreAnimationEngine` check. Will work on building this out in follow-up PRs.

### DIFF
--- a/Sources/Public/Animation/AutomaticRenderingEngine.swift
+++ b/Sources/Public/Animation/AutomaticRenderingEngine.swift
@@ -1,0 +1,32 @@
+// Created by Cal Stephens on 5/2/22.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+import Foundation
+
+extension RenderingEngineOption {
+  public func renderingEngine(for animation: Animation) -> RenderingEngine {
+    switch self {
+    case .specific(let engine):
+      return engine
+    case .automatic:
+      return animation.supportedByCoreAnimationEngine ? .coreAnimation : .mainThread
+    }
+  }
+
+  public func renderingEngine(for animation: Animation?) -> RenderingEngine? {
+    switch self {
+    case .specific(let engine):
+      return engine
+    case .automatic:
+      guard let animation = animation else { return nil }
+      return animation.supportedByCoreAnimationEngine ? .coreAnimation : .mainThread
+    }
+  }
+}
+
+extension Animation {
+  /// Whether or not this animation can be rendered by the Core Animation engine
+  public var supportedByCoreAnimationEngine: Bool {
+    false // TODO: Implement
+  }
+}
--- a/Sources/Public/Animation/AutomaticRenderingEngine.swift
+++ b/Sources/Public/Animation/AutomaticRenderingEngine.swift
@@ -27,6 +27,6 @@ extension RenderingEngineOption {
 extension Animation {
   /// Whether or not this animation can be rendered by the Core Animation engine
   public var supportedByCoreAnimationEngine: Bool {
-    false // TODO: Implement
+    true // TODO: Implement checks
   }
 }
--- a/Example/Example.xcodeproj/xcshareddata/xcschemes/Example (iOS).xcscheme
+++ b/Example/Example.xcodeproj/xcshareddata/xcschemes/Example (iOS).xcscheme
@@ -20,6 +20,20 @@
                ReferencedContainer = "container:Example.xcodeproj">
             </BuildableReference>
          </BuildActionEntry>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "2E80409927A0725D006E74CB"
+               BuildableName = "Lottie.framework"
+               BlueprintName = "Lottie-iOS"
+               ReferencedContainer = "container:../Lottie.xcodeproj">
+            </BuildableReference>
+         </BuildActionEntry>
       </BuildActionEntries>
    </BuildAction>
    <TestAction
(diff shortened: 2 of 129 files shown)


Sources/Public/Animation/AutomaticRenderingEngine.swift
```diff
@@ -0,0 +1,32 @@
...
+      return animation.supportedByCoreAnimationEngine ? .coreAnimation : .mainThread
+    }
+  }
+}
+
+extension Animation {
+  /// Whether or not this animation can be rendered by the Core Animation engine
+  public var supportedByCoreAnimationEngine: Bool {
+    true // TODO: Implement checks
+  }
```
erichoracek, line 31: Do you think it will be feasible to keep these checks in sync with the failures that occur when you're instantiating the layer hierarchy, especially in more dynamic scenarios? Wondering if it may be simpler to make the layer hierarchy setup code propagate a thrown error and then set up an alternative main thread layer hierarchy in that case, rather than having two ways to set up different hierarchies—although I guess that method will have a lot of extra overhead to every animation setup in that scenario. 

So I guess with your approach with the unit test validation and you're able to create a robust enough suite of example animations we'd probably be fine 👍 Just want to make sure you considered an alternate approach.

erichoracek (approved):
@calda LGTM! Just one question

Sources/Public/Animation/AutomaticRenderingEngine.swift
calda: I agree that a setup like you're describing would almost certainly be easier to maintain.

>  I guess that method will have a lot of extra overhead to every animation setup in that scenario.

Yeah, this is my main concern with that approach. Setting up the layer hierarchy and all of the `CAAnimation`s is the most expensive part of using the CA engine, so I worry that doing this unnecessarily would be too much overhead.

On the other hand, manual checks also come with overhead since it would have to basically iterate through every keyframe in the entire animation. This would always be wasted work in cases where the animation does support the CA engine.  Attempting to set up the Core Animation engine, and then falling back to the main thread engine when an assertion is emitted, would have no performance penalty in cases where the animation supports the CA engine. This would only give us a performance penalty when falling back to the Main Thread engine. 

I think it's worth evaluating the relative performance of these two approaches. I'll whip up a prototype of each of these and instrument the difference.

calda: As a rough comparison, I compared the performance of two implementations:
 1. one that iterated through the majority of the animation keyframes in the animation, and did some placeholder checks
 2. one that set up the animation using the CA engine, and then fell back to setting up the main thread engine

(1) was 5-10% slower than the existing Core Animation engine. On the other hand, (2) was double as slow as the existing Main Thread engine (not a big surprise). 

Having a single source of truth for this is definitely appealing, but it feels somewhat hard to justify making the Main Thread engine double as expensive.

erichoracek: @calda Got it, thanks for doing the test—do you have any idea of what percentage of animations are going to fall into the buckets of:
- Work with the Core Animation renderer
- Work with the main thread renderer

The reason I ask is because my naïve guess would be the breakdown is probably 80% CA / 20% MT, and one of the benefits of doing the check at layer hierarchy setup time is that you'll only pay the cost of double-setup in the case of the fallback from CA to MT—so in the case of the animation being CA-compliant, there's no overhead. 

With the animation keyframe-based solution, it sounds like we'll be adding a 5-10% overhead by default to every animation, regardless of CA/MT renderer, and it will be up to the consumer author to specify explicitly if they want to remove that overhead.

With either option, as a further optimization, you could even log when the fallback occurs, and specifically call out that because this animation is non-compliant that setup has some extra overhead, and tell them they can explicitly mark this animation with its renderer only to speed up initialization (in the case of the layer setup time fallback approach).

Let me know what you think!

calda: > do you have any idea of what percentage of animations are going to fall into the buckets of:
>
> - Work with the Core Animation renderer
> - Work with the main thread renderer

Good question. 80/20 is a reasonable place to start from. If you amortize the total cost over an 80/20 split, you get:

`(2.0 * 0.2) + (1.0 * 0.8) = 1.2x` (20% of the animations cost 2x baseline, but 80% are unchanged)
`(1.15 * 1.0) = 1.15x` (all animations cost 1.15x baseline, as an over-estimate)

So with an 80/20 split the total performance cost would be slightly in favor of a manual check, but not by much.

> You could even log when the fallback occurs, and specifically call out that because this animation is non-compliant that setup has some extra overhead, and tell them they can explicitly mark this animation with its renderer only to speed up initialization (in the case of the layer setup time fallback approach).

This is probably the best argument in favor of using the CA-setup-then-fallback approach. To get back to the existing performance baseline with that approach, you'd only have to manually annotate 20% of the animations as using the main thread engine. On the other hand, to get back to the existing performance baseline with the other approach, you'd have to annotate _all_ animations.

Given that the amortized total performance cost of the two approaches are roughly equivalent, and there are many practical benefits to the CA-setup-then-fallback approach, I think I'll change gears and pursue that approach instead. Thanks for the thorough and thoughtful feedback!

calda: Another data point -- right now 13% of the animations in the regression suite emit assertions when rendered with the Core Animation engine. 

That number of definitely an under-estimate, because:
 - There are some animations that I know render incorrectly but haven't figured out why yet (so don't have an assertion)
 - I only know about the Lottie features used by the animations in the regression suite, so there are definitely Lottie features that I still don't know about (so are unsupported by the CA engine, and don't have an assertion).

but I can believe that 15-25% is the right ballpark.

PR #1555, opened 2022-05-04
Rename ExperimentalAnimationLayer to CoreAnimationLayer

calda:
Now that the Core Animation rendering engine is fairly mature, I think it's time to drop the "experimental" nomenclature. This PR renames:
 - `ExperimentalAnimationLayer` to `CoreAnimationLayer`
 - `RootAnimationContainer` to `MainThreadAnimationLayer`


erichoracek (approved):
@calda LGTM!

erichoracek:
@calda on thought I had after this was merged is that `CALayer` already stands for "Core Animation Layer", which may be confusing—probably not worth another rename but may be worth disambiguating in a comment

calda:
Haha, I had the same thought. Although I think this is ok as-is because:
 1. this is an internal type
 2. the name feels clear in-context (especially as it relates to the `MainThreadAnimationLayer`)
 3. the doc comment mentions that this is for Lottie's Core Animation engine.

 Thoughts? Can also add more discussion in a comment somewhere (and not opposed to renaming that specific type if you have a better idea, it would be an easy change)

PR #1556, opened 2022-05-04
Implement fallback behavior for RenderingEngineOption.automatic

calda:
This PR implements the fallback behavior for `RenderingEngineOption.automatic`. If a compatibility issue is encountered, `AnimationView` will fall-back to using the main thread engine.

This PR includes some snapshot tests that exercise various code paths, and a performance test demonstrating that using `RenderingEngineOption.automatic` has the same performance characteristics as `RenderingEngineOption.coreAnimation` (for supported animations).


Sources/Private/CoreAnimation/CoreAnimationLayer.swift
```diff
@@ -15,28 +15,17 @@ final class CoreAnimationLayer: BaseAnimationLayer {
...
     setup()
-
-    do {
-      try setupChildLayers()
-    } catch {
-      if case CompatibilityTracker.Error.encounteredCompatibilityIssue(let compatibilityIssue) = error {
-        // If the initial layer setup was successful, then the animation will continue
-        // being setup later in `CALayer.display()` below. If this initial setup failed,
-        // then we report the error now.
-        didSetUpAnimation?([compatibilityIssue])
```
calda, line 37: I reworked this initializer to `throw` after I encountered a silly issue:

```swift
self.animationLayer = CoreAnimationLayer(
  animation: animation,
  compatibilityTrackerMode: .abort,
  didSetUpAnimation: {
     // ...
     makeAnimationLayer(usingEngine: .mainThread)
     // This could run from inside the initializer, meaning the code that takes the result of the initializer
     // and applies it to `self.animationLayer` would run _after_ we finished setting up the main thread layer,
     // putting the `AnimationView` in an inconsistent state.
  })
```

Sources/Public/Animation/AnimationView.swift
```diff
@@ -939,6 +934,87 @@ final public class AnimationView: AnimationViewBase {
...
+    }
+
+    LottieLogger.shared.warn(
+      compatibilityIssues.compatibilityMessage
+        + "\nFalling back to Main Thread rendering engine.\n")
+
+    let animationContext = self.animationContext
+    let currentFrame = self.currentFrame
+
+    makeAnimationLayer(usingEngine: .mainThread)
```
calda, line 1007: This was easier than I expected!

Sources/Public/Animation/AnimationView.swift
```diff
@@ -939,6 +955,87 @@ final public class AnimationView: AnimationViewBase {
...
+
+      coreAnimationLayer.didSetUpAnimation = { compatibilityIssues in
+        LottieLogger.shared.assert(
+          compatibilityIssues.isEmpty,
+          compatibilityIssues.compatibilityMessage)
+      }
+
+      return coreAnimationLayer
+    } catch {
+      LottieLogger.shared.assertionFailure("Encountered unexpected error \(error)")
```
erichoracek, line 982: Could you include a comment detailing how we don't expect this case to occur since we're using the `track` mode, and this is programmer error if we hit this (I assume)?

Sources/Private/CoreAnimation/CompatibilityTracker.swift
```diff
@@ -126,3 +126,12 @@ extension LayerAnimationContext: CompatibilityTrackerProviding {
     currentKeypath.fullPath
   }
 }
+
+extension Array where Element == CompatibilityIssue {
+  var compatibilityMessage: String {
+    let lines = ["Encountered Core Animation compatibility issues while setting up animation:"]
```
erichoracek, line 132: Probably be worth including details on how to recover from this error in both types of animation configuration

Sources/Public/Animation/AnimationView.swift
```diff
@@ -939,6 +955,87 @@ final public class AnimationView: AnimationViewBase {
...
+  // Handles any compatibility issues with the Core Animation engine
+  // by falling back to the Main Thread engine
+  fileprivate func automaticEngineLayerDidSetUpAnimation(_ compatibilityIssues: [CompatibilityIssue]) {
+    // If there weren't any compatibility issues, then there's nothing else to do
+    if compatibilityIssues.isEmpty {
+      return
+    }
+
+    LottieLogger.shared.warn(
+      compatibilityIssues.compatibilityMessage
```
erichoracek, line 1022: Maybe we could include information that this is the first error, but there may be more (since we're using the `abort` mode)? 

And maybe details on the fact that this fallback has performance overhead, and you should specify the main thread rendering engine directly if you're seeing degraded performance

Sources/Public/Animation/AnimationView.swift
erichoracek: Nice!

erichoracek (approved):
@calda Nice, LGTM! Just some suggestions on user-facing error messages

Sources/Public/Animation/AnimationView.swift
calda: good idea

Issue #1363, opened 2021-06-01
Gradient rendering regression in Lottie 3+ compared to ObjC


absolute-heike:
@buba447 I am currently debugging the gradient regression in 3.2.2 and found out, that this regression was already introduced when switching from Obj-C to Swift.
The problem seems to be related to layers scaled inside of After Effects. The sample animation contains 3 layers:
1. A shape with the size of 23x23
2. A linear gradient
3. A transform that scales this shape up

When debugging the different versions I am seeing that the context size is 46x46 in `Obj-C` as well as Swift `3.2.2`. Context size in Swift `3.2.1` is 3 times as high (138x138) which has a drastic impact on performance and still looks way worse than Obj-C.

I couldn't find any significant difference in the gradient rendering code, so I assume that the issue is related to shape rendering or transform handling in Lottie-Swift.

Any hints, where I should have a deeper look to track down the bug?

## Comparison
https://user-images.githubusercontent.com/522881/120313618-88e06b00-c2da-11eb-809e-51c45227d7c4.mov

## Animation JSON
[gradient-animation.json.zip](https://github.com/airbnb/lottie-ios/files/6575772/gradient-animation.json.zip)

## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [x] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [] Feature Request
- [x] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?
Lottie 3.0

## What Platform are you on?

- [] MacOS
- [x] iOS

absolute-heike:
This seems to be related to https://github.com/airbnb/lottie-ios/issues/1022#issuecomment-556124547

Now I just need to figure out the right place to make this change^^

absolute-heike:
@buba447 I finally figured it out! 🎉

The issue is with the way the path is rendered. In ObjC the path was solved with a `CAShapeLayer` as a mask whereas Swift adds the path to the `CGContext` directly.

When the `CAShapeLayer` is scaled up with a transform, it's scaled razer-sharp. But `CGContext.addPath()` is rendered at low resolution and only transformed afterwards

The solution I'm now aiming for now is:
[...]

PR #1558, opened 2022-05-05
Improve gradient fill performance

marcelofabri:
This builds off the great work by @absolute-heike in https://github.com/airbnb/lottie-ios/compare/master...absolute-heike:create_temporary_pod?expand=1. 

https://github.com/airbnb/lottie-ios/issues/1363 also has more details, but in short, Lottie 3 switched from using a `CAShapeLayer` (see [LOTRadialGradientLayer](https://github.com/airbnb/lottie-ios/blob/2.5.3/lottie-ios/Classes/Extensions/LOTRadialGradientLayer.m) and [LOTGradientFillRender](https://github.com/airbnb/lottie-ios/blob/2.5.3/lottie-ios/Classes/RenderSystem/RenderNodes/LOTGradientFillRender.m) to drawing the gradient every time (see [GradientFillRenderer](https://github.com/airbnb/lottie-ios/blob/master/Sources/Private/MainThread/NodeRenderSystem/Nodes/OutputNodes/Renderables/GradientFillRenderer.swift). This is orders of magnitude slower (almost 3x slower in my testing) and is blocking us from moving to Lottie 3. It's specially bad when using a `CADisplayLink` to drive the Lottie animation (so we can use a custom animation curve).

Gradient stokes are still rendered using the Lottie 3.x way and probably should be updated at some point, but at least it's not a regression since gradient strokes were introduced in Lottie 3 (https://github.com/airbnb/lottie/commit/5182094d074f5f532945eb9a9c07bee973a37246#).

Note that this introduces some visual differences in the gradients, but it's hard to tell which is the correct one. For example, I compared `LottieFiles/gradient_1.json` before and after this PR with the rendering from the preview tool in https://lottiefiles.com and each one rendered it slightly different. If this is really a problem and we can't figure out why (I double checked the rendering params and they look the same), we could introduce this behind a flag in `Configuration`.


calda (approved):
Thank you! I don't see any meaningful visual changes in any of the sample animations, so this looks safe to merge.

PR #1559, opened 2022-05-05
Expose API to get original value from animation

### LINKED ISSUE #1547: Value Provider API doesn't provide current value

## This issue is a:
- [ ] Non-Crashing Bug (Visual or otherwise)
- [ ] Crashing Bug
- [x] Feature Request
- [x] Regression (Something that once worked, but doesn't work anymore)

`master`

## What Platform are you on? 
- [ ] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [ ] Objective-C

## Expected Behavior 

There should be a way to get the current value for the keypath in a `ValueProvider`, so it can use it in its calculations.

This is something that was supported in `LOTValueDelegate`. One use case is to move parts of the animation dynamically as 
the device moves.

## Actual Behavior

There's no way to get the current value for a keypath. I've tried using `animatedView.getValue(for:keypath:)` but it ends up calling the value provider, causing an infinite recursion.

## Code Example

```swift
import CoreMotion
import Lottie

final class AnimationPositionDelegate: NSObject, LOTPointValueDelegate {
    private var referenceAttitude: CMAttitude?
    private weak var motionManager: CMMotionManager?
    private let takeoverFrameRange: ClosedRange<Int>? // Nil implies full-range takeover
    private let pitchMultiplier: CGFloat
    private let rollMultiplier: CGFloat

    init(motionManager: CMMotionManager?,
         takeoverFrameRange: ClosedRange<Int>?,
         pitchMultiplier: CGFloat,
         rollMultiplier: CGFloat) {

        self.motionManager = motionManager
        self.takeoverFrameRange = takeoverFrameRange
        self.pitchMultiplier = pitchMultiplier
        self.rollMultiplier = rollMultiplier

        super.init()
    }

    func reCalibrateReferenceAttitude() {
        guard let motionManager = motionManager,
              motionManager.isDeviceMotionAvailable,
              let deviceMotion = motionManager.deviceMotion else {
            return
        }

        referenceAttitude = deviceMotion.attitude
    }

    func clearReferenceAttitude() {
        referenceAttitude = nil
    }

    // MARK: - LOTPointValueDelegate

    public func point(forFrame currentFrame: CGFloat,
                      startKeyframe: CGFloat,
                      endKeyframe: CGFloat,
                      interpolatedProgress: CGFloat,
                      start startPoint: CGPoint,
                      end endPoint: CGPoint,
                      currentPoint interpolatedPoint: CGPoint) -> CGPoint {

        if let takeoverFrameRange = takeoverFrameRange {
            guard takeoverFrameRange.contains(Int(currentFrame)) else {
                return interpolatedPoint
            }
        }

        guard let motionManager = motionManager,
              motionManager.isDeviceMotionAvailable,
              let deviceMotion = motionManager.deviceMotion else {
            return interpolatedPoint
        }

        if let referenceAttitude = referenceAttitude {
            deviceMotion.attitude.multiply(byInverseOf: referenceAttitude)
        }

        let roll = clampedRoll(forDeviceMotion: deviceMotion)
        let pitch = clampedPitch(forDeviceMotion: deviceMotion)

        let adjustedPoint = CGPoint(
            x: interpolatedPoint.x + CGFloat(roll) * rollMultiplier,
            y: interpolatedPoint.y + CGFloat(pitch) * pitchMultiplier
        )

        return adjustedPoint
    }

    private func clampedRoll(forDeviceMotion deviceMotion: CMDeviceMotion) -> Double {
        let roll = deviceMotion.attitude.roll
        let lowerBoundedRoll = max(roll, Constants.rollLowerBound)
        return min(lowerBoundedRoll, Constants.rollUpperBound)
    }

    private func clampedPitch(forDeviceMotion deviceMotion: CMDeviceMotion) -> Double {
        let pitch = deviceMotion.attitude.pitch
        let lowerBoundedPitch = max(pitch, Constants.pitchLowerBound)
        return min(lowerBoundedPitch, Constants.pitchUpperBound)
    }

    private struct Constants {
        static let rollLowerBound: Double = -1.1
        static let rollUpperBound: Double = 1.1
        static let pitchLowerBound: Double = -1
        static let pitchUpperBound: Double = 1
    }
}
```

marcelofabri:
Fixes #1547

### DIFF
--- a/Tests/ValueProvidersTests.swift
+++ b/Tests/ValueProvidersTests.swift
@@ -1,0 +1,27 @@
+//
+//  ValueProvidersTests.swift
+//  LottieTests
+//
+//  Created by Marcelo Fabri on 5/5/22.
+//
+
+import Lottie
+import XCTest
+
+final class ValueProvidersTests: XCTestCase {
+
+  func testGetValue() throws {
+    let animationView = try XCTUnwrap(SnapshotConfiguration.makeAnimationView(
+      for: "HamburgerArrow",
+      configuration: .init(renderingEngine: .mainThread)))
+
+    let keypath = AnimationKeypath(keypath: "A1.Shape 1.Stroke 1.Color")
+    animationView.setValueProvider(ColorValueProvider(.red), keypath: keypath)
+    let updatedColor = try XCTUnwrap(animationView.getValue(for: keypath, atFrame: 0) as? Color)
+    XCTAssertEqual(updatedColor, .red)
+
+    let originalColor = try XCTUnwrap(animationView.getOriginalValue(for: keypath, atFrame: 0) as? Color)
+    XCTAssertEqual(originalColor, Color(r: 0.4, g: 0.16, b: 0.7, a: 1))
+  }
+
+}
--- a/Lottie.xcodeproj/project.pbxproj
+++ b/Lottie.xcodeproj/project.pbxproj
@@ -545,6 +545,7 @@
 		2EAF5B0427A0798700E00531 /* AnimationFontProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2EAF59F227A0798700E00531 /* AnimationFontProvider.swift */; };
 		2EAF5B0527A0798700E00531 /* AnimationFontProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2EAF59F227A0798700E00531 /* AnimationFontProvider.swift */; };
 		2EAF5B0627A0798700E00531 /* AnimationFontProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2EAF59F227A0798700E00531 /* AnimationFontProvider.swift */; };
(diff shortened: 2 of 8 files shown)


Tests/ValueProvidersTests.swift
```diff
@@ -0,0 +1,27 @@
...
+      for: "HamburgerArrow",
+      configuration: .init(renderingEngine: .mainThread)))
+
+    let keypath = AnimationKeypath(keypath: "A1.Shape 1.Stroke 1.Color")
+    animationView.setValueProvider(ColorValueProvider(.red), keypath: keypath)
+    let updatedColor = try XCTUnwrap(animationView.getValue(for: keypath, atFrame: 0) as? Color)
+    XCTAssertEqual(updatedColor, .red)
+
+    let originalColor = try XCTUnwrap(animationView.getOriginalValue(for: keypath, atFrame: 0) as? Color)
+    XCTAssertEqual(originalColor, Color(r: 0.4, g: 0.16, b: 0.7, a: 1))
```
calda, line 24: Thanks for the test case!

calda (approved):

PR #1373, opened 2021-07-02
Add support for Animation init with a dictionary (2-10x faster decoding)

zakforrest:
Provides 2-10x faster decoding time and also uses 3-4x less memory while decoding, compared to using `JSONDecoder`. This is documented under the Performance Regression Refactor project card [here](https://github.com/airbnb/lottie-ios/projects/2#card-29711649) and was discussed in the past as a solution [here](https://github.com/airbnb/lottie-ios/issues/872#issuecomment-498733043)  (#872).

- Adds `DictionaryInitializable` and `AnyInitializable` protocols
- The `Animation` and child objects adhere to `DictionaryInitializable`
- Added a public `Animation` initializer using `Data`
- No core logic changed - only added protocols and applied them to existing models

Tested using all provided animations, as well as ~20 animations of my own, on an iPhone XS.

PR #1561, opened 2022-05-05
Add alternative dictionary-based decoding

marcelofabri:
This is mostly a rebase from https://github.com/airbnb/lottie-ios/pull/1373, with some fixes and additional tests

When running the tests, decoding with Codable is ~2x as slow as this new approach. However, I think it makes sense to keep `Codable` as the default strategy for now, until we're fully confident this doesn't introduce any regressions.

I'll test this with the ~150 animations I have from a private app to double check.

### DIFF
--- a/Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -2,6 +2,15 @@
   "object": {
     "pins": [
       {
+        "package": "Difference",
+        "repositoryURL": "https://github.com/krzysztofzablocki/Difference",
+        "state": {
+          "branch": null,
+          "revision": "02fe1111edc8318c4f8a0da96336fcbcc201f38b",
+          "version": "1.0.1"
+        }
+      },
+      {
         "package": "Epoxy",
         "repositoryURL": "https://github.com/airbnb/epoxy-ios.git",
         "state": {
--- a/Sources/Private/Model/DictionaryInitializable.swift
+++ b/Sources/Private/Model/DictionaryInitializable.swift
@@ -32,19 +32,19 @@ protocol AnyInitializable {
 extension Dictionary {
 
   @_disfavoredOverload
-  func value<T>(for key: Key) throws -> T {
-    guard let value = self[key] as? T else {
+  func value<T, KeyType: RawRepresentable>(for key: KeyType) throws -> T where KeyType.RawValue == Key {
+    guard let value = self[key.rawValue] as? T else {
       throw InitializableError.invalidInput
     }
     return value
   }
 
-  func value<T: AnyInitializable>(for key: Key) throws -> T {
-    if let value = self[key] as? T {
+  func value<T: AnyInitializable, KeyType: RawRepresentable>(for key: KeyType) throws -> T where KeyType.RawValue == Key {
+    if let value = self[key.rawValue] as? T {
       return value
     }
 
-    if let value = self[key] {
+    if let value = self[key.rawValue] {
       return try T(value: value)
     }
 
--- a/Sources/Private/Model/DictionaryInitializable.swift
+++ b/Sources/Private/Model/DictionaryInitializable.swift
@@ -1,0 +1,67 @@
+//
+//  DictionaryInitializable.swift
+//  Lottie
+//
+//  Created by Marcelo Fabri on 5/5/22.
+//
+
+import Foundation
+
+// MARK: - InitializableError
+
+enum InitializableError: Error {
+  case invalidInput
+}
+
+// MARK: - DictionaryInitializable
+
+protocol DictionaryInitializable {
+
+  init(dictionary: [String: Any]) throws
+
+}
+
+// MARK: - AnyInitializable
+
+protocol AnyInitializable {
+
+  init(value: Any) throws
+
+}
+
+extension Dictionary {
+
+  @_disfavoredOverload
+  func valueFor<T>(key: Key) throws -> T {
+    guard let value = self[key] as? T else {
+      throw InitializableError.invalidInput
+    }
+    return value
+  }
+
+  func valueFor<T: AnyInitializable>(key: Key) throws -> T {
+    if let value = self[key] as? T {
+      return value
+    }
+
+    if let value = self[key] {
+      return try T(value: value)
+    }
+
+    throw InitializableError.invalidInput
+  }
+
+}
+
+// MARK: - Array + AnyInitializable
+
+extension Array: AnyInitializable where Element == Double {
+
+  init(value: Any) throws {
+    guard let array = value as? [Double] else {
+      throw InitializableError.invalidInput
+    }
+    self = array
+  }
+
+}
--- a/Sources/Private/Model/DictionaryInitializable.swift
+++ b/Sources/Private/Model/DictionaryInitializable.swift
@@ -32,14 +32,14 @@ protocol AnyInitializable {
 extension Dictionary {
 
   @_disfavoredOverload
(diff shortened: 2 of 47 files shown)


Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,6 +1,15 @@
 {
   "object": {
     "pins": [
+      {
```
marcelofabri, line 4: this is just because I didn't want to add `Equatable` conformance to the whole model tree

calda:
Could you share more details about the performance gains of this change? How many ms does it take to parse typical animations, and how many ms does this new parsing implementation save us?

Tests/ParsingTests.swift
```diff
@@ -0,0 +1,42 @@
...
+final class ParsingTests: XCTestCase {
+
+  func testParsingIsTheSameForBothImplementations() throws {
+    for url in Samples.sampleAnimationURLs {
+      do {
+        let data = try Data(contentsOf: url)
+        let codableAnimation = try Animation.from(data: data, strategy: .codable)
+        let dictAnimation = try Animation.from(data: data, strategy: .dictionaryBased)
+
+        XCTAssertNoDiff(codableAnimation, dictAnimation)
```
calda, line 24: Thanks for this test, good idea!

marcelofabri:
> Could you share more details about the performance gains of this change? How many ms does it take to parse typical animations, and how many ms does this new parsing implementation save us?

I'm running some more tests and I'll report back! 

> It would be interesting, for example, to create a test in `PerformanceTests.swift` that compares the performance of parsing all of the sample animations with `Codable` vs parsing all of the sample animations with the new dictionary implementation.

I did that ad-hoc and decoding all sample animations with `Codable` was taking ~2s and ~1s with the new implementation. I can add a more formalized test ✨

Sources/Public/LottieConfiguration.swift
```diff
@@ -16,6 +20,7 @@ public struct LottieConfiguration: Hashable {
 
   /// The rendering engine implementation to use when displaying an animation
   public var renderingEngine: RenderingEngineOption
+  public var decodingStrategy: DecodingStrategy
```
calda, line 23:
```suggestion

  /// The decoding implementation to use when parsing an animation JSON file
  public var decodingStrategy: DecodingStrategy
```

marcelofabri:
On a quick test in a real app:

```
codableTime: 6.436601638793945, avg: 0.041796114537623025
dictTime: 2.2157540321350098, avg: 0.014388013195681882
largest diff: 0.4925088882446289
``` 

So almost 3x improvement and saving up to 500ms.

calda:
Great! That's a pretty impressive improvement.

> `largest diff: 0.4925088882446289`

Does this mean you have an animation that parses HALF A SECOND FASTER with the new implementation? How big is this thing 😝

Sources/Private/Model/DictionaryInitializable.swift
```diff
@@ -0,0 +1,67 @@
...
+protocol AnyInitializable {
+
+  init(value: Any) throws
+
+}
+
+extension Dictionary {
+
+  @_disfavoredOverload
+  func valueFor<T>(key: Key) throws -> T {
```
calda, line 35: Why are both of these methods necessary?

Sources/Private/Model/DictionaryInitializable.swift
```diff
@@ -0,0 +1,67 @@
...
+    guard let value = self[key] as? T else {
+      throw InitializableError.invalidInput
+    }
+    return value
+  }
+
+  func valueFor<T: AnyInitializable>(key: Key) throws -> T {
```
calda, line 42: I think something like `value(for key: Key)` would be a bit more idiomatic.

Also, I see most of the callsites have `valueFor(key: CodingKeys.myKey.rawValue)`. It could be nice to simplify this to just `value(for: CodingKeys.myKey)` by supporting a generic `RawRepresentable` key:

```suggestion
  func value<T: AnyInitializable, KeyType>(key: KeyType) throws -> T were KeyType: RawRepresentable, KeyType.RawValue == Key {
```

marcelofabri:
> > `largest diff: 0.4925088882446289`
> 
> Does this mean you have an animation that parses HALF A SECOND FASTER with the new implementation? How big is this thing 😝

Yes! It's huge: 3.2MB 🥵

Sources/Private/Model/DictionaryInitializable.swift
marcelofabri: > I think something like value(for key: Key) would be a bit more idiomatic.

Agreed, I mostly copied from the other PR and didn't think too much about it. I'll add a way to improve usage with `CodingKeys`, good idea!

Sources/Public/LottieConfiguration.swift
```diff
@@ -6,8 +6,12 @@
 /// Global configuration options for Lottie animations
 public struct LottieConfiguration: Hashable {
 
-  public init(renderingEngine: RenderingEngineOption = .mainThread) {
+  public init(
+    renderingEngine: RenderingEngineOption = .mainThread,
+    decodingStrategy: DecodingStrategy = .codable)
```
calda, line 11: I think I'm comfortable with making this the default, considering the tests pass (and the performance wins are pretty substantial!)

```suggestion
    decodingStrategy: DecodingStrategy = .dictionaryBased)
```

Sources/Public/LottieConfiguration.swift
```diff
@@ -91,5 +98,16 @@ extension RenderingEngineOption: RawRepresentable, CustomStringConvertible {
...
+
+// MARK: - DecodingStrategy
+
+/// How animation files should be decoded
+public enum DecodingStrategy: Hashable {
+  /// Use Codable. This is the default strategy introduced on Lottie 3.
+  case codable
 
+  /// Manually deserialize a dictionary into an Animation. This should be faster,
+  /// but since it's manually implemented, there might be issues while it's experimental.
```
calda, line 111: It would be nice to include discussion here about how much faster the `dictionaryBased` implementation is than the `codable` implementation. Also, if we do [make it the default](https://github.com/airbnb/lottie-ios/pull/1561/files#r866326477), we probably don't want to call it experimental here.

calda (approved):
Thanks for working on this!

Sources/Private/Model/DictionaryInitializable.swift
marcelofabri: There are cases where we don't implement `AnyInitializable` (eg primitives like `Double`) so we can't simply add the constraint.

I couldn't find a good way to dynamically check whether `T` conforms to `AnyInitializable`, so had to create a separate function

calda: Makes sense. Is this `@_disfavoredOverload` annotation necessary? My understanding is that Swift calls the most-specific method that applies to a given type, so types that conform to `AnyInitializable` would automatically use the other method (without needing `@_disfavoredOverload`).

Sources/Public/LottieConfiguration.swift
marcelofabri: I think we should let this mature a little bit just in case before making it the default. It makes me nervous to potentially break animations 🥵

Sources/Private/Model/DictionaryInitializable.swift
marcelofabri: Yeah, without it the test was failing since the version with the constraint was never called 😅

Sources/Public/LottieConfiguration.swift
calda: Sounds good

PR #1562, opened 2022-05-06
Allow loading animations with “.json” extension in name

marcelofabri:
This is another small (surprising) compatibility issue I've found when updating a large codebase to Lottie 3.

We had a few places loading local animations using "animationName.json" and it started falling after the update. Turns out Lottie 2.x had [a weird logic](https://github.com/airbnb/lottie-ios/blob/2.5.3/lottie-ios/Classes/Private/LOTComposition.m#L27-L28) dropping everything after a `.` in the filename. I don't think we need to go that far because it's not a great behavior, but supporting `".json"` would cover the most common case.

### DIFF
--- a/Sources/Private/Model/Extensions/Bundle.swift
+++ b/Sources/Private/Model/Extensions/Bundle.swift
@@ -6,6 +6,7 @@ import UIKit
 extension Bundle {
   func getAnimationData(_ name: String, subdirectory: String? = nil) throws -> Data? {
     // Check for files in the bundle at the given path
+    let name = name.removingJSONSuffix()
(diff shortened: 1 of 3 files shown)


Sources/Private/Model/Extensions/Bundle.swift
```diff
@@ -19,3 +20,13 @@ extension Bundle {
     #endif
   }
 }
+
+extension String {
+  fileprivate func removingJSONSuffix() -> String {
+    guard hasSuffix(".json") else {
```
calda, line 26: A comment explaining why we specifically only do this for `.json` files would be nice

calda (approved):
Sure, seems reasonable enough considering Lottie 2.x supported this

PR #1563, opened 2022-05-06
Add support for spatial keyframe animations

calda:
This PR adds support for spatial keyframes.

Position keyframes can provide `spatialInTangent` and `spatialOutTangent` values that define bezier path control points. We can apply those to the `CAKeyframeAnimation` by creating a `GGPath` for the layer to animate along, instead of simply animating linearly between the keyframe points.

For example:

| Before | After |
| ----- | ---- |
|  |  |


Example/iOS/ViewControllers/AnimationPreviewViewController.swift
```diff
@@ -46,11 +46,19 @@ class AnimationPreviewViewController: UIViewController {
     animationView.backgroundBehavior = .pauseAndRestore
     animationView.translatesAutoresizingMaskIntoConstraints = false
 
+    engineLabel.font = .preferredFont(forTextStyle: .footnote)
```
calda, line 49: While I was here I also added a label to the example app's preview VC that tells you which rendering engine is being used

Sources/Private/CoreAnimation/Animations/CALayer+addAnimation.swift
```diff
@@ -205,6 +222,44 @@ extension CALayer {
     return timingFunctions
   }
 
+  /// Creates a `CGPath` for the given `position` keyframes,
+  /// which accounts for `spatialInTangent`s and `spatialOutTangents`
+  private func path<KeyframeValue>(
+    keyframes positionKeyframes: ContiguousArray<Keyframe<KeyframeValue>>,
+    value keyframeValueMapping: (KeyframeValue) throws -> CGPoint) rethrows
+    -> CGPath {
```
calda, line 230: SwiftFormat doesn't like this open brace for some reason, it won't let me put it on the next line. Never seen this issue before.

PR #1566, opened 2022-05-06
Actually fix issue where animation could fail to start playing when using RenderingEngineOption.automatic

calda:
This PR _actually_ fixes the issue where an animation could fail to start when using `RenderingEngineOption.automatic`. 

The fix in #1560 worked in some cases, but not all. After digging in, I found out that this was because the `AnimationContext`'s `AnimationCompletionDelegate` (the animation layer's `CAAnimationDelegate`) was being put in a bad state: 
 - the `CoreAnimationLayer` would call `delegate.animationDidStop(_:finished:)` when it was removed from the view hierarchy
 - at that point the delegate object had been reconfigured so that `self.animationLayer` is a `MainThreadAnimationLayer`
 - the delegate called `animationLayer.removeAnimation`, cancelling the animation on the new `MainThreadAnimationLayer`

To fix this, we can just initialize a new `AnimationContext` with the same values instead of using the previous instance.

### DIFF
--- a/Sources/Public/Animation/AnimationView.swift
+++ b/Sources/Public/Animation/AnimationView.swift
@@ -1062,11 +1062,14 @@ final public class AnimationView: AnimationViewBase {
     self.currentFrame = currentFrame
 
     if let animationContext = animationContext {
-      // We have to wait until the next run loop cycle to start playing the animation,
-      // or it may not actually start as expected.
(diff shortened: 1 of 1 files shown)


erichoracek (approved):
Nice! Glad we can remove the DispatchQueue.main.async

PR #1586, opened 2022-05-10
Fix issue where layer matte would fail to be applied

calda:
This PR fixes an issue where the layer matte would fail the be applied. In this specific case, the `mask` layer itself also had a `mask`, which Core Animation seems to silently ignore. Adding an additional intermediate layer fixes the issue.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-05-10 at 16 35 20 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-05-10 at 16 34 46 |

There are some other issues with this animation still, which I'm still looking in to.


Tests/SnapshotTests.swift
```diff
@@ -35,6 +35,11 @@ class SnapshotTests: XCTestCase {
   /// a sample JSON file that is visible to this test target.
   func testAllSnapshotsHaveCorrespondingSampleFile() {
     for snapshotURL in Samples.snapshotURLs {
+      // Exclude snapshots of private samples, since those aren't checked in to the repo
+      if snapshotURL.lastPathComponent.contains("Private") {
+        continue
```
calda, line 40: This fails locally for me any time I move or rename samples in the Private folder, which seems unnecessary considering those aren't checked in to the repo

erichoracek (approved):
LGTM!

PR #1589, opened 2022-05-11
Fix issue where Fill opacity was ignored

calda:
This PR fixes an issue where the `opacity` animation / configuration from `Fill` `ShapeItem`s was ignored.

Previously, we animated the `opacity` of `ShapeItemLayers` from the `opacity` value of any `ShapeItemTransform` items. I checked how this works in the Main Thread engine, and found that we only apply `ShapeItemTransform`s to `Group`s. Moving the `ShapeItemTransform` animation to the parent `GroupLayer` makes it so the `ShapeItemLayer.opacity` property can instead be animated by the `Fill` `opacity` is present.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-05-10 at 16 34 46 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-05-11 at 10 47 00 |


erichoracek (approved):
@calda LGTM!

PR #1590, opened 2022-05-12
Use image cached provider to reuse asset when creating RootAnimationLayer

jianjunwoo:
## Change summary:

Before this change, the Lottie animation layer creates images for each layer from asset which costs memory.

In this PR, the `BundleImageProvider ` is enhanced with image cache to reduce memory usage when creating layers with same asset.

The lifecycle of `CachedImageProvider` is the same as each `animation` in `AnimationView`.

## Please review:

|| before | after |
|-|-|-|
| Sample Lottie |before | after|

### DIFF
--- a/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
+++ b/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
@@ -45,4 +45,3 @@ extension AnimationImageProvider {
     CachedImageProvider(imageProvider: self)
   }
 }
-
--- a/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
+++ b/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
@@ -3,8 +3,6 @@
 
 import CoreGraphics
 import Foundation
-#if os(iOS) || os(tvOS) || os(watchOS) || targetEnvironment(macCatalyst)
-import UIKit
 
 // MARK: - CachedImageProvider
 
@@ -48,4 +46,3 @@ extension AnimationImageProvider {
   }
 }
 
-#endif
--- a/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
+++ b/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
@@ -1,0 +1,51 @@
+// Created by Jianjun Wu on 2022/5/12.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+import CoreGraphics
+import Foundation
+#if os(iOS) || os(tvOS) || os(watchOS) || targetEnvironment(macCatalyst)
+import UIKit
+
+// MARK: - CachedImageProvider
+
+private final class CachedImageProvider: AnimationImageProvider {
+
+  // MARK: Lifecycle
+
+  /// Initializes an image provider with an image provider
+  ///
+  /// - Parameter imageProvider: The provider to load image from asset
+  ///
+  public init(imageProvider: AnimationImageProvider) {
+    self.imageProvider = imageProvider
+  }
+
+  // MARK: Public
+
+  public func imageForAsset(asset: ImageAsset) -> CGImage? {
+    if let image = imageCache.object(forKey: asset.id as NSString) {
+      return image
+    }
+    if let image = imageProvider.imageForAsset(asset: asset) {
+      imageCache.setObject(image, forKey: asset.id as NSString)
+      return image
+    }
+    return nil
+  }
+
+  // MARK: Internal
+
+  let imageCache: NSCache<NSString, CGImage> = .init()
+  let imageProvider: AnimationImageProvider
+}
+
+extension AnimationImageProvider {
+  /// Create a cache enabled image provider which will reuse the asset image with the same asset id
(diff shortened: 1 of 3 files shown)


calda (approved):
Thanks!

calda:
@jianjunwoo, to make the linter check pass you can run `bundle exec rake format:swift` in your local checkout

Sources/Public/iOS/BundleImageProvider.swift
```diff
@@ -23,15 +23,30 @@ public class BundleImageProvider: AnimationImageProvider {
...
+  public init(bundle: Bundle, searchPath: String?, cacheEnabled: Bool = true) {
     self.bundle = bundle
     self.searchPath = searchPath
+    imageCache = cacheEnabled ? .init() : nil
   }
 
   // MARK: Public
 
   public func imageForAsset(asset: ImageAsset) -> CGImage? {
+    if let imageCache = imageCache, let cachedImage = imageCache.object(forKey: asset.id as NSString) {
```
marcelofabri, line 37: Is it safe to cache assets between different animations? Couldn’t two animations use the same id for different assets?

calda: Good point! We use `asset.name` and `asset.directory` below when retrieving the image (and _not_ the `id`), so we should use those as the cache key.

jianjunwoo: The provider is created for each AnimationView https://github.com/airbnb/lottie-ios/blob/master/Sources/Public/Animation/AnimationView.swift#L97
Thus, the asset id should be unique in an lottie json I think.
The `asset.name` may represent for base64 image data, its size is too big. @marcelofabri @calda

jianjunwoo:
> bundle exec rake format:swift

thanks! this is helpful

Sources/Public/iOS/BundleImageProvider.swift
jianjunwoo: > Is it safe to cache assets between different animations? Couldn’t two animations use the same id for different assets?

This inspire me that if it is possible we could share `BundleImageProvider` between `AnimationView`s. If the provider is shared, the memory usage of image would be reduced if an animation is used in more than on `AnimationView`. Then, the cache key may need to have the animation json hash as key. @marcelofabri

calda: An individual `AnimationView` can display multiple different Lottie animations over time

calda: I think we need a solution for using `name` and `directory` in the cache key. Otherwise it will be possible for the cache to return an incorrect result.

> The asset.name may represent for base64 image data, its size is too big.

Too big for what? Does `NSCache` not allow you to store these?

jianjunwoo: It may cost cpu for calculate hash value of big string (` name`) as NSCache key @calda

jianjunwoo: > An individual `AnimationView` can display multiple different Lottie animations over time

I would like to see if we could reset cache for when animation changed.

jianjunwoo: I updated the PR with different solution, the cached provider is recreated for each time `Animation` updated.
This solves the cache key conflict problem among different Lottie animations. @calda

calda: Sounds good, it seems safe to assume that an individual animation can only have a single image with a given `id`.

Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
```diff
@@ -0,0 +1,51 @@
...
+
+  let imageCache: NSCache<NSString, CGImage> = .init()
+  let imageProvider: AnimationImageProvider
+}
+
+extension AnimationImageProvider {
+  /// Create a cache enabled image provider which will reuse the asset image with the same asset id
+  /// It wraps the current provider as image loader, and uses `NSCache` to cache the images for resue.
+  /// The cache will be reset when the `animation` is reset.
+  var cachedImageProvider: AnimationImageProvider {
```
calda, line 46: We'll also need a definition of this for `macOS` and `tvOS` for the CI to pass.

Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
```diff
@@ -0,0 +1,51 @@
+// Created by Jianjun Wu on 2022/5/12.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+import CoreGraphics
+import Foundation
+#if os(iOS) || os(tvOS) || os(watchOS) || targetEnvironment(macCatalyst)
+import UIKit
```
calda, line 7: What is the UIKit import for?

calda (approved):
LGTM once CI passes

Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
jianjunwoo: Thanks for point out!
I once used cache UIImage to cache, and forgot to remove this after I changed to use CGImage.

PR #1592, opened 2022-05-12
Update precomp layers to mask sublayers to bounds

calda:
This PR updates `PreCompLayer` to mask its sublayers to its bounds (using the size specified in the animation json, which we were previously ignoring). This is the behavior used by the main thread engine.

| Before | After |
| ----- | ----- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-05-11 at 16 34 55 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-05-11 at 16 36 42 |


erichoracek (approved):
LGTM!

PR #1594, opened 2022-05-12
Handle Groups with multiple Fill items

calda:
To continue validating the `.automatic` rendering engine option and the Core Animation engine, I downloaded and tested ~a dozen complex animations from LottieFiles. I found once that displayed incorrectly, and added it to the regression suite.

This specific issue here was that `Group`s had multiple different `Fill` items, which we didn't handle correctly. For example, for a group with `[Rectangle, Fill (Red), Circle, Fill (Blue)]`:
 - previously, both of the two shapes would be red
 - now, the rectangle is red and the circle is blue (matching the behavior of the main thread engine)

| Before | After |
| ----- | ---- |
| testCoreAnimationRenderingEngine Private-BrokenLottieFiles-growth-50 | testCoreAnimationRenderingEngine LottieFiles-growth-50 |


Sources/Private/CoreAnimation/Animations/GradientAnimations.swift
```diff
@@ -53,6 +53,8 @@ extension GradientRenderLayer {
       },
       context: context)
 
+    try addOpacityAnimation(for: gradient.opacity, context: context)
```
calda, line 56: While I was working on this I also noticed we were ignoring the opacity configuration on `GradientFill`s, so I added that here like we did for `Fill`s in #1589

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
```diff
@@ -97,44 +97,49 @@ final class GroupLayer: BaseAnimationLayer {
...
-        shape: ShapeItemLayer.Item(item: combinedShape, parentGroup: group),
-        otherItems: otherItemsInGroup,
-        context: context)
-
-      addSublayer(sublayer)
-    }
+    // Create `ShapeItemLayer`s for each subgroup of shapes that should be rendered as a single unit
+    //  - These groups are listed from front-to-back, so we have to add the sublayers in reverse order
+    for shapeRenderGroup in nonGroupItems.shapeRenderGroups.reversed() {
+      // If all of the path-drawing `ShapeItem`s have keyframes with the same timing information,
```
calda, line 103: Everything in this method was just re-indented to account for this new `for` loop

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
```diff
@@ -197,3 +214,47 @@ extension Collection {
     return (trueElements, falseElements)
   }
 }
+
+// MARK: - ShapeRenderGroup
+
+/// A group of `ShapeItem`s that should be rendered together as a single unit
```
calda, line 220: This is the key new functionality:

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
```diff
@@ -197,3 +214,47 @@ extension Collection {
...
+
+    for item in self {
+      // `renderGroups` is non-empty, so is guaranteed to have a valid end index
+      let lastIndex = renderGroups.indices.last!
+
+      if item.item.drawsCGPath {
+        renderGroups[lastIndex].pathItems.append(item)
+      }
+
+      // `Fill` items are unique, because they specifically only apply to _previous_ shapes in a `Group`
```
erichoracek, line 241: What a fun gotcha!

erichoracek (approved):
Nice, LGTM!

erichoracek:
image
😱

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
calda: I never cease to be surprised by how complicated / nuanced Lottie's `ShapeLayer` actually is

PR #1599, opened 2022-05-17
Add support for Stroke opacity animations

calda:
This PR adds support for Stroke opacity animations, which were previously ignored.

Both `Fill` and `Stroke` items can provide an `opacity` animation. Since an individual `CALayer` can only play a single `alpha` animation, we have to render `Fill` and `Stroke`s as separate layers. This is how the main thread engine works as well.

This causes a test failure in one animation, but I think this was because #1589 was only partially correct. I'll work on a fix for that new issue in a follow-up.

| Before | After |
| ---- | ---- |
|  |  |


erichoracek (approved):
LGTM!

calda:
The fix for the regression I mentioned turned out to be a pretty small change, so I included it in this PR (https://github.com/airbnb/lottie-ios/pull/1599/commits/515666be6840fa572dc442a666b9e40d18246030):

image

muhuaxin (commented):
Thanks!

PR #1600, opened 2022-05-18
Fix issue where some animations would unexpectedly allocate a very large amounts of memory

calda:
This PR fixes an issue where some animations would unexpectedly allocate a very large amount of memory (400mb+). I'm not exactly sure what was causing this, but the Allocations instrument says the memory is being allocated by the `super.display()` call in `CoreAnimationLayer.display`. 

Since that layer doesn't render any content directly, we can just remove the `super.display()` call without experiencing any behavior changes. Alternatively we could have `CoreAnimationLayer` subclass `CATransformLayer`, which also works but causes some unwanted log output:

> <Lottie.CoreAnimationLayer: 0x7fce30d14c10> - changing property contentsScale in transform-only layer, will have no effect

### DIFF
--- a/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
+++ b/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
@@ -118,7 +118,12 @@ final class CoreAnimationLayer: BaseAnimationLayer {
   }
 
   override func display() {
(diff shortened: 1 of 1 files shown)


erichoracek (approved):
if it works! 😅

PR #1601, opened 2022-05-18
Update Core Animation engine to explicitly not support time remapping keyframes with `isHold = true`

calda:
This PR updates the Core Animation engine to explicitly not support time remapping keyframes with `isHold = true`. This is not supported right now, but previously failed silently. We may be able to support this in the future, but for now we should fall back to using the main thread engine in this case.

| Before | After |
| ----- | ----- |
|  |  |


thedrick (approved):

erichoracek (approved):
👍

PR #1604, opened 2022-05-24
Fall back to main thread engine for animations using time remapping

calda:
This PR updates `RenderingEngineOption.automatic` to fall back to using the main thread rendering engine for animations that use time remapping. 

I've found a handful of cases that animate incorrectly with the current implementation of time remapping, which I haven't been able to fix without causing regressions elsewhere. I haven't been able to figure out why these animations are failing when other animations seem to work.

Since we want 100% of animations to play correctly when using `RenderingEngineOption.automatic`, the only option right now is to fall back to the main thread engine for animations that use time remapping. Fortunately this is a relatively uncommon feature so this change doesn't affect that many animations.


erichoracek (approved):
Bummer, but makes sense, LGTM!

PR #1605, opened 2022-05-25
Fix mysterious crash due to reentrant PreCompLayer.init call

calda:
As far as I can tell, this PR fixes a mysterious crash that appears to have been caused by reentrant calls to `PreCompLayer.init`.

I haven't been able to reproduce this crash a single time locally. @thedrick was able to repro this crash 100% of the time on a certain page, and he confirmed that this change fixed the issue for him.


erichoracek (approved):
Glad this fixed it 😅

PR #1608, opened 2022-05-25
Fix AnimationView.isAnimationPlaying for Core Animation engine

calda:
This PR fixes `AnimationView.isAnimationPlaying` for the Core Animation engine. Previously is returned `true` even when the animation was paused on a specific frame.

| Before | After |
| ----- | ---- |
|  |  |


erichoracek (approved):
LGTM!

PR #1612, opened 2022-06-03
Add support for trim offsets to Core Animation rendering engine

jtbergman:
**Overview**
Currently, the Core Animation rendering engine does not support trim offsets. This PR adds partial support for trim offsets by adjusting the `strokeStart` and `strokeEnd` keyframes accordingly. Trim offsets can be supported when the adjusted stroke values are all in the range `[0, 100]`.

**Verification**
Lottie tests pass locally and a significant number of animations in our app are fixed. Examples:

| Master | Pull Request |
| ------- | ------------- |
| <video src="https://user-images.githubusercontent.com/61445278/171775541-23332afa-b6c9-44be-9929-dd7ace81a8ef.mp4"> | <video src="https://user-images.githubusercontent.com/61445278/171775522-3fea5941-3f9e-455b-afe4-8c7370d5f96d.mp4"> |


Sources/Private/CoreAnimation/Animations/ShapeAnimation.swift
```diff
@@ -82,19 +76,48 @@ extension Trim {
...
+      offsetKeyframes: offset.keyframes,
+      context: context)
+
+    // Validate the adjusted keyframes and fallback on original if invalid
+    if
+      (adjustedStrokeStart + adjustedStrokeEnd).contains(where: {
+        $0.value.cgFloatValue < 0 || $0.value.cgFloatValue > 100
+      }) {
+      try context.logCompatibilityIssue("""
+        The CoreAnimation rendering engine doesn't support Trim offsets
```
calda, line 113: Can we make this error more specific, so it describes the specific Trim configuration that's unsupported? For example:

```suggestion
        The Core Animation rendering engine doesn't support Trim offsets with values outside of the range [0, 1].
```

Sources/Private/CoreAnimation/Animations/ShapeAnimation.swift
```diff
@@ -120,4 +143,82 @@ extension Trim {
...
   }
+
+  /// Adjusts a `KeyframeGroup` associated with a `strokeStart` or `strokeEnd` to account
+  /// for trim offsets. The CoreAnimation rendering engine does not fully support trim offsets. If applying
+  /// the offsets results in a `KeyframeGroup` with all values in the range `[0, 1]` then the offsets
+  /// can be supported.
+  private func adjustKeyframesForTrimOffsets(
+    strokeKeyframes: ContiguousArray<Keyframe<Vector1D>>,
+    offsetKeyframes: ContiguousArray<Keyframe<Vector1D>>,
+    context _: LayerAnimationContext) throws -> ContiguousArray<Keyframe<Vector1D>> {
```
calda, line 154: Interesting, so this is basically combining the `trim.strokeStart` / `trim.strokeEnd` keyframes with the `offset` keyframes? Good to know. Could you add some discussion about this as a doc comment (discussion about why we have to combine these two values into a single set of keyframes would be super valuable!)

jtbergman: I updated the comment but let me know if it still isn't clear. 

It should be possible to add full support for trim offsets. The main problem now is that an adjusted value `(strokeStart: 0.8, strokeEnd: 1.2)` cannot but supported because it crosses the "zero-point" of the animation. It should be possible to support this by appending the path to itself (creating two overlapping copies), and then adjusting that value again to be `(strokeStart: 0.4, strokeEnd: 0.6)` which is now valid and draws what was originally intended. 

This works in plain Core Animation code as implemented in [this Gist](https://gist.github.com/jtbergman/c138ddf3cf4082b261770b4d877c73b0), but when I tried to implement it [for Lottie](https://github.com/airbnb/lottie-ios/compare/master...jtbergman:jt-trim-offsets-full-support) it didn't work as expected. It seems the `CAShapeLayer` the stroke values are applied to is not the same `CAShapeLayer` the path is defined on? I'm also not sure the approach to duplicating the path works correctly in Lottie. Let me know if you have any ideas!

jtbergman:
@calda Thanks for the speedy review! I don't think I can add our animations to the test suite, but I will try to create a simple animation that tests this functionality. I'll try to add that to this PR by Monday 😄

calda:
That would be super helpful, thanks! No rush, thanks again for contributing.

Sources/Private/CoreAnimation/Animations/ShapeAnimation.swift
calda: I have an implementation of that path-duplicating approach here https://github.com/airbnb/lottie-ios/commit/416856c2f784025e85bf61d740eeb6815a174b44. This seems to sort-of-work, but isn't quite working with this `LottieFiles/loading_gradient_strokes` animation. I can't quite tell why, maybe `strokeStart` is growing to be larger than `strokeEnd` for a portion of the animation, causing the stroke to disappear entirely?

 | Default | Swapping `strokeStart` and `strokeEnd` keyframes (you can see the opposite portion of the animation works) |
| --- | ---- |
|  |  |

calda: It's clearly wrapping around the zero-point of the circle, though, which is progress! Would be amazing if we can figure out how to get this animation working.

jtbergman: Nice work and thanks for doing this! 

> I can't quite tell why, maybe `strokeStart` is growing to be larger than `strokeEnd` for a portion of the animation, causing the stroke to disappear entirely?

I think this is the problem and it's caused by a bug that affects the original PR as well. In the original PR I just used the `inTangent` and `outTangent` of the most recent stroke or offset keyframe when creating the adjusted keyframes ([source](https://github.com/airbnb/lottie-ios/blob/1ce5ac6e9e032800cfb414ca7f687447698e745f/Sources/Private/CoreAnimation/Animations/ShapeAnimation.swift#L216-L219)). But the stroke and offset values can have independent `inTangent`/`outTangent` values which means the current approach to combining the keyframes is too simple, and I doubt there's an easy solution. As a next step, I'm going to try to manually interpolate the keyframes so that the `inTangent` and `outTangent` can be set to `nil`. My guess is this will be too expensive to use as-is, but it'll help verify the problem and approach.

jtbergman: I implemented this and the performance seems to be ok! I need to touch up the code before pushing a new commit, but here's the result: 

| Main Thread | Core Animation |
| ----- | ------------- |
| <video src="https://user-images.githubusercontent.com/61445278/174523861-bed1810a-fced-4c9a-b47f-14850e093a82.mp4"> | <video src="https://user-images.githubusercontent.com/61445278/174523899-47aa4ac8-c703-4698-8a0f-d38e1fa27a50.mp4"> |

The animation for type-face G is also much better. 

| Master | Pull Request |
| ----- | ------------- |
| <video src="https://user-images.githubusercontent.com/61445278/174524198-7509969d-a1e3-4fff-9190-97b59f0bf358.mp4"> | <video src="https://user-images.githubusercontent.com/61445278/174523994-5e35da8f-aeb9-442e-b031-8d48672285ec.mp4"> |

calda: Amazing work!!!! Great stuff, thanks for the help

Sources/Private/Utility/Interpolatable/KeyframeGroup+Extensions.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import Foundation
+
+extension KeyframeGroup where T == Vector1D {
+  /// Manually interpolates the keyframes so that they are defined linearly
+  ///
+  /// This method uses `UnitBezier` to perform the interpolation. It will create one keyframe
+  /// for each frame of the animation. For instance, if it is given a keyframe at time 0 and a keyframe
+  /// at time 10, it will create 10 interpolated keyframes. It is currently not optimized.
```
jtbergman, line 15: > It is currently not optimized

The gradient strokes animation seems to use 5% CPU with Core Animation vs 15% CPU with main thread. The Duolingo animations CPU usage isn't changed much since there is now one keyframe. There's room for optimization here, but this PR is already quite long. The simplest idea is that a new keyframe doesn't need to created if the following is true: 
```
// Linear change can be collapsed into two keyframes instead of 3
curr.newValue - prev.newValue == prev.newValue - prev.prev.newValue
```

Considering the CPU usage, it may be worth still throwing an error to warn the user?

Sources/Private/Utility/Interpolatable/KeyframeGroup+Extensions.swift
```diff
@@ -0,0 +1,58 @@
...
+      let startValue = prev.value.value
+      let startTime = prev.time
+      let duration = max(Int(curr.time - prev.time), 0)
+
+      // Create one interpolated keyframe for each time in the duration
+      for t in 0 ... duration {
+        let progress = timingFunction.value(
+          for: CGFloat(t) / CGFloat(duration),
+          epsilon: 0.005)
+        let value = startValue + progress * difference
```
jtbergman, line 43: There exists a similar method `Keyframe.interpolate(to:progress:)` but is either performing a different operation or giving less accurate results.

Sources/Private/Utility/Interpolatable/KeyframeGroup+Extensions.swift
calda: This just happens once when setting up the animation, right? As long as we don't incur a performance hit on every frame (like with the main thread engine) then this seems perfectly fine to me!

calda (approved):
Thanks! Will review a bit more closely tomorrow and get this merged.

Sources/Private/Utility/Interpolatable/KeyframeGroup+Extensions.swift
calda: That test is a bit flaky, I'd just suggest increasing the threshold from `0.5` to `0.55`

Tests/__Snapshots__/AutomaticEngineTests/testAutomaticEngineDetection.LottieFiles-loading_gradient_strokes.txt
```diff
@@ -1,9 +1 @@
...
-[Shape Layer 1.Trim Paths 1] The CoreAnimation rendering engine doesn't support Trim offsets
-[Shape Layer 1.Trim Paths 1] The CoreAnimation rendering engine doesn't support Trim offsets
-[Shape Layer 2.Trim Paths 1] The CoreAnimation rendering engine doesn't support Trim offsets
-[Shape Layer 2.Trim Paths 1] The CoreAnimation rendering engine doesn't support Trim offsets
-[Shape Layer 4.Trim Paths 1] The CoreAnimation rendering engine doesn't support Trim offsets
-[Shape Layer 4.Trim Paths 1] The CoreAnimation rendering engine doesn't support Trim offsets
-[Shape Layer 3.Trim Paths 1] The CoreAnimation rendering engine doesn't support Trim offsets
-[Shape Layer 3.Trim Paths 1] The CoreAnimation rendering engine doesn't support Trim offsets
\ No newline at end of file
+Supports Core Animation engine
```
calda, line 1: wooooooo :)

calda:
@jtbergman, here's a related change that you may find interesting: https://github.com/airbnb/lottie-ios/pull/1698. Seems like Core Animation doesn't support _negative_ offset values, so we have to shift all of them to be positive.

PR #1613, opened 2022-06-03
Add a LottieLogger param to AnimationView

calda:
This PR adds a `logger: LottieLogger = .shared` param to `AnimationView`, so consumers can customize the logger both globally (via `LottieLogger.shared`) and on a per-instance basis (like you can already do with `LottieConfiguration`).

As an example, this is useful for cases where using `renderingEngine: .coreAnimation` emits an assertion but you want to use the Core Animation engine anyway. An option there is to pass in `logger: .printToConsole`, which suppresses any animation validation assertions.


erichoracek (approved):
Nice, LGTM!

PR #1619, opened 2022-06-27
`NSImage.CGImage` ⇒ `.lottie_CGImage` to avoid category warning.

jasdev:
Unfortunately https://github.com/airbnb/lottie-ios/pull/1501 doesn’t fix the category warning when using Lottie on macOS. Would y’all be open to namespacing the extension?

Screen Shot 2022-06-27 at 5 11 19 PM


calda (approved):
LGTM, thanks!

PR #1634, opened 2022-07-07
Add support for gradients with alpha values

### LINKED ISSUE #1628: `automatic` `renderingEngine` in `AnimationView` not rendering gradients properly
Gradients in a couple of my Lottie files are not rendering properly with the `.automatic` `renderingEngine` enabled

| Main Thread  | Automatic |
| ------------- | ------------- |
|   |    |

It’s a normal linear gradient, but the colour at one end is rgba(75,177,97,0.2) and the colour at the other is rgba(75,177,97,0), so the gradient is only changing the alpha

Lottie 3.4.0, installed via SPM. iOS 15.5 simulator running on macOS 12.4 (21F79) (M1 Macbook Pro)

If you would like a copy of the Lottie file I would prefer to send it privately.

Note this particular animation view is also embedded in a SwiftUI view, in case that's relevant.

calda:
This PR adds support for gradients with alpha values. Previously we silently ignored these alpha values. Fixes #1628.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-07-07 at 08 55 57 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-07-07 at 09 20 07 |


thedrick (approved):
LGTM @calda

Sources/Private/CoreAnimation/Animations/GradientAnimations.swift
```diff
@@ -131,14 +135,31 @@ extension GradientShapeItem {
     for colorIndex in 0..<numberOfColors {
       let colorStartIndex = colorIndex * 4
 
-      let location = CGFloat(colorComponents[colorStartIndex])
-
-      let color = CGColor.rgb(
+      let colorLocation = CGFloat(colorComponents[colorStartIndex])
+
+      // Alpha values aren't stored contiguously with the RGB color components,
+      // but are instead optionally listed at the end of the array
```
erichoracek, line 141: haha that's fun one 😅

PR #1637, opened 2022-07-07
Fixed Animated Control Accessibility Configuration

brynbodayle:
- Fixes an issue where `AnimatedButton` and `AnimatedSwitch` weren't showing up as accessibly elements
- Fixes an issues where `AnimatedButton` and `AnimatedSwitch` weren't providing the default traits for UIControl provided states like when `isEnabled = false` or `isSelected = true`

Screen Shot 2022-07-07 at 4 20 07 PM

### DIFF
--- a/Sources/Public/iOS/AnimatedControl.swift
+++ b/Sources/Public/iOS/AnimatedControl.swift
@@ -126,15 +126,15 @@ open class AnimatedControl: UIControl {
     get { animationView.animationSpeed }
   }
 
+  public override var accessibilityTraits: UIAccessibilityTraits {
+    set { super.accessibilityTraits = newValue }
+    get { super.accessibilityTraits.union(.button) }
+  }
+
   /// Sets which Animation Layer should be visible for the given state.
   public func setLayer(named: String, forState: UIControl.State) {
     stateMap[forState.rawValue] = named
     updateForState()
-  }
-
-  public override var accessibilityTraits: UIAccessibilityTraits {
-    set { customAccessibilityTraits = newValue }
-    get { customAccessibilityTraits.union(.button).union(super.accessibilityTraits) }
   }
 
   /// Sets a ValueProvider for the specified keypath
@@ -164,8 +164,6 @@ open class AnimatedControl: UIControl {
   }
 
   // MARK: Private
-
-  private var customAccessibilityTraits: UIAccessibilityTraits = []
 
   private func commonInit() {
     isAccessibilityElement = true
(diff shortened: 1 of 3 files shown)


Sources/Public/iOS/AnimatedControl.swift
```diff
@@ -132,6 +132,11 @@ open class AnimatedControl: UIControl {
     updateForState()
   }
 
+  public override var accessibilityTraits: UIAccessibilityTraits {
+    set { customAccessibilityTraits = newValue }
+    get { customAccessibilityTraits.union(.button).union(super.accessibilityTraits) }
+  }
```
calda, line 138: I wonder if we could avoid the stored `customAccessibilityTraits` property with an implementation like this:

```suggestion
  public override var accessibilityTraits: UIAccessibilityTraits {
    set { super.accessibilityTraits = newValue }
    get { super.accessibilityTraits.union(.button) }
  }
```

calda (approved):

Sources/Public/iOS/AnimatedControl.swift
calda: or do we specifically need to not modify `super.accessibilityTraits` to get the UIControl provided states like when `isEnabled = false` or `isSelected = true`?

brynbodayle: Good question, confirming now.

brynbodayle: Your simplification works and maintains the UIControl provided states! Will update.

calda: Nice! Thanks

Sources/Public/iOS/AnimatedControl.swift
```diff
@@ -126,6 +126,11 @@ open class AnimatedControl: UIControl {
     get { animationView.animationSpeed }
   }
 
+  public override var accessibilityTraits: UIAccessibilityTraits {
+    set { super.accessibilityTraits = newValue }
+    get { super.accessibilityTraits.union(.button) }
```
erichoracek, line 131: @brynbodayle Are all controls buttons? I'm worried that with things like `UITextField` subclassing `UIControl` this may not be a valid assumption at the root `AnimatedControl` and instead should be something we add on a per-subclass basis

brynbodayle: Hmm, I think it in this case all our controls would have user interaction, but in the case of a slider it wouldn't be a button. Let me update this to only be button for `AnimatedButton` and `AnimatedSwitch`

PR #1641, opened 2022-07-08
Fix Carthage support on macOS

### LINKED ISSUE #1638: Carthage macOS use the wrong Swift Module Name (Lottie_macOS) but not (Lottie)

Lottie 3.4.0

## Expected Behavior 

The `PRODUCT_MODULE_NAME` should be `Lottie`, not `Lottie_macOS`, which is the target name.

This will cause downstream have to use 

```swift
import Lottie_macOS
```

instead, which is a break chanage from Lottie v3.1.0

截屏2022-07-09 01 32 44

calda:
This PR fixes #1638. The Carthage library for macOS was unintentionally called `Lottie_macOS` instead of `Lottie`, so consumers had to use `import Lottie_macOS` instead of `import Lottie`.

44dad4c updates the `bundle exec rake test:carthage` script to test Carthage support on macOS.

5ea1d57 fixes the library name, which lets the new test case pass.


dreampiggy:
LGTM

erichoracek (approved):

PR #1646, opened 2022-07-11
Fix issue where a `ShapeLayer` with no `Group`s would fail to render

### LINKED ISSUE #1645: `Samples/TypeFace/Y` renders incorrectly using automatic engine
I noticed `Samples/TypeFace/Y` rendered incorrectly using `.automatic` -- the Y shape at the end of the animation never appears.

| `.automatic` | `.mainThread` |
| ---- | ---- |
| testCoreAnimationRenderingEngine TypeFace-Y-100 | testMainThreadRenderingEngine TypeFace-Y-100 |

calda:
This PR fixes an issue where a `ShapeLayer` with no `Group`s would fail to render. Fixes #1645.

| Before | After |
| ---- | ---- |
| testCoreAnimationRenderingEngine TypeFace-Y-100 | testMainThreadRenderingEngine TypeFace-Y-100 |


Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
```diff
@@ -150,7 +150,14 @@ extension CALayer {
   ///  - Each `Group` item becomes its own `GroupLayer` sublayer.
   ///  - Other `ShapeItem` are applied to all sublayers
   fileprivate func setupGroups(from items: [ShapeItem], parentGroup: Group?, context: LayerContext) throws {
-    let (groupItems, otherItems) = items.grouped(by: { $0 is Group })
+    var (groupItems, otherItems) = items.grouped(by: { $0 is Group })
+
+    // If this shape doesn't have any groups but just has top-level shape items,
+    // we can create a placeholder group with those items.
```
erichoracek, line 156: It might be helpful to discuss what happens if you don't do this, since it's not entirely clear from the context

erichoracek (approved):
LGTM!

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
calda: Good idea

calda:
```suggestion
    // If this shape doesn't have any groups but just has top-level shape items,
    // we can create a placeholder group with those items. (Otherwise the shape items
    // would be silently ignored, since we expect all shape layers to have a top-level group).
```

PR #1653, opened 2022-07-18
Fix issue where `AnimationView.animationContext` could be put in an inconsistent state

calda:
This PR fixes an issue where `AnimationView.animationContext` could be put in an inconsistent state.

`AnimationView.animationContext` should be present when an animation is actively playing, and `nil` when the animation is paused. Previously the Core Animation engine wasn't clearing out this value when pausing an animation, which meant `animationContext` would be non-nil even though the animation was paused.

This caused `updateAnimationForBackgroundState()` and `updateAnimationForForegroundState()` to unexpectedly run even when the animation was paused. In this issue below the `AnimationView` is using `.forceFinish`, so the animation progress is unexpectedly changed when using the Core Animation engine.

| Core Animation engine (before) | Main thread engine, Core Animaiton engine (after) |
| ----- | ---- |
|  |  |

### DIFF
--- a/Sources/Public/Animation/AnimationView.swift
+++ b/Sources/Public/Animation/AnimationView.swift
@@ -927,7 +927,11 @@ final public class AnimationView: AnimationViewBase {
 
   // MARK: Fileprivate
 
+  /// Context describing the animation that is currently playing in this `AnimationView`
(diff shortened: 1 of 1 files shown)


erichoracek (approved):
Good catch!

PR #1654, opened 2022-07-19
Add configuration parameter for initializers of animation view

jhen0409:
Provide the `configuration` parameter for initializers, so we can use it to initialize a new animation view by source name / file path or URL.


calda (approved):
Thanks!

PR #1656, opened 2022-07-20
Adopt AirbnbSwiftFormatPlugin for code formatting and linting

calda:
This PR adopts the new `AirbnbSwiftFormatPlugin` from https://github.com/airbnb/swift/pull/168. This lets us remove our dependency on `mint`, and remove our direct dependencies on SwiftFormat / SwiftLint (now managed by the format plugin).


Package@swift-5.6.swift
```diff
@@ -0,0 +1,15 @@
+// swift-tools-version:5.6
```
calda, line 1: The style tool requires `swift-tools-version:5.6`. We can't modify the existing `Package.swift` file since that will break the build for old toolchains, so instead we can add a separate `Package@swift-5.6.swift` file that gets used when compiling with Xcode 5.6+.

.swift-version
```diff
@@ -0,0 +1 @@
+5.3
```
erichoracek, line 1: If we have the `#if swift(>=5.6)` in the package.swift do we still need this? Also where is this file documented?

erichoracek (approved):
@calda LGTM! Just one question

Package@swift-5.6.swift
```diff
@@ -0,0 +1,15 @@
...
+
+let package = Package(
+  name: "Lottie",
+  platforms: [.iOS("11.0"), .macOS("10.10"), .tvOS("11.0")],
+  products: [
+    .library(name: "Lottie", targets: ["Lottie"]),
+  ],
+  dependencies: [
+    .package(url: "https://github.com/airbnb/swift", branch: "master"),
+  ],
```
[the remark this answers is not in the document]
calda: Ohhh smart, good idea! I always forget that this file is actual Swift source and not like yml

.swift-version
calda: This controls the Swift version that SwiftFormat uses -- otherwise it will make changes that are invalid in 5.3. I'll see if I can put a comment or something in this file.

calda: Yep this file supports comments as long as the version number is on the first line

Package@swift-5.6.swift
```diff
@@ -0,0 +1,15 @@
...
+import PackageDescription
+
+let package = Package(
+  name: "Lottie",
+  platforms: [.iOS("11.0"), .macOS("10.10"), .tvOS("11.0")],
+  products: [
+    .library(name: "Lottie", targets: ["Lottie"]),
+  ],
+  dependencies: [
+    .package(url: "https://github.com/airbnb/swift", branch: "master"),
```
calda, line 11: We probably either need to check in the `Package.resolved` file, or reference a specific commit here. Otherwise builds on Lottie master could start failing because we updated the Style repo. If I can't get the `Package.resolved` to work properly I'll update this to reference a specific commit.

.swift-version
calda: Unfortunately adding the comment caused `pod spec lint` to fail so I had to remove it. On the bright side that means this file is the source-of-truth for Lottie's support Swift version across multiple different tools.

PR #1660, opened 2022-07-24
Add support set negative speed to do reverse transition on Core Animation

jhen0409:
It looks like the Core Animation rendering engine is not supported to set `animationSpeed` less than 0 for doing reverse animation, which the main thread engine does.

My solution is switch `playFrom` and `playTo` if `animationSpeed` less than 0, not sure if this is best way. I also add speed menu into example.


calda (approved):
Thank you!

PR #1665, opened 2022-07-26
Fix issue where `LottieLogger.shared.assertionFailure` would unexpectedly crash in release builds

### LINKED ISSUE #1663: `LottieLogger.assertFailure` should not crash in Release build.

Lottie 3.3.0

## Expected Behavior 

`LottieLogger.shared().assertFailure` should not trigger in Release build. Assertion should be filtered.

## Actual Behavior

It crashes when `assertFailure` happened. Did not work as expected.

## Animation JSON

[latest.zip](https://github.com/airbnb/lottie-ios/files/9188910/latest.zip)

calda:
This PR fixes issue where `LottieLogger.shared.assertionFailure` would unexpectedly crash in release builds. 

Naively I would expect both of these implementations to behave the same. I filed a compiler bug for this: https://github.com/apple/swift/issues/60249

Fixes #1663.

### DIFF
--- a/Sources/Public/Logging/LottieLogger.swift
+++ b/Sources/Public/Logging/LottieLogger.swift
@@ -10,8 +10,19 @@ public final class LottieLogger {
   // MARK: Lifecycle
 
   public init(
-    assert: @escaping Assert = Swift.assert,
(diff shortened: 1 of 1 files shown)


erichoracek (approved):
LGTM, thanks!

Issue #1672, opened 2022-07-27
Installing Lottie 3.4.1 is not possible with gn generated xcodeProjects


levinda:
## Which Version of Lottie are you using?

Lottie 3.4.1

works just fine with Lottie 3.4.0

## Expected Behavior 

Installs lottie pod
## Actual Behavior

Does not install and provokes an error:

`Unable to determine Swift version for the following pods:
- `lottie-ios` does not specify a Swift version and none of the targets  integrating it have the `SWIFT_VERSION` attribute set. Please contact the author or set the `SWIFT_VERSION` attribute in at least one of the targets that integrate this pod.`

I have looked through the repository and found that swift_version in podspec has been removed [here](https://github.com/airbnb/lottie-ios/commit/97c05f8272fc8e3a2cefb3d41c6bcb90c6d28739#diff-ca71f5c8020e3c192f455b0eb9f9f8df32f43b28f11c0089f6e1f03078ec3d74) in favor of .swift-version file that has been deprecated with CocoaPods 1.4.0, [link to the change log](https://blog.cocoapods.org/CocoaPods-1.4.0/)
 
<img width="805" alt="Screen Shot 2022-07-27 at 23 46 13" src="https://user-images.githubusercontent.com/44981990/181368984-0e2c6949-b33d-4234-bc8d-64d242097f83.png">

calda:
Adopting https://github.com/airbnb/swift/pull/180 once merged will let us remove the `.swift-version` file and add back the `swift_version` attribute to the podspec.

calda:
@levinda, this should be fixed on master by https://github.com/airbnb/lottie-ios/pull/1674. Could you check if that change resolves your issue?

PR #1674, opened 2022-07-28
Fix CocoaPods installation issue by removing `.swift-version` file

calda:
According to #1672 it sounds like the configuration change in #1656 is causing issues. I added a `.swift-version` file to configure the Swift version used by SwiftFormat. This conflicted with the `swift_version` configuration in the podspec, so I removed that in favor of just using the `.swift-version` file. That seems to cause issues in some versions and/or use cases of CocoaPods, so we probably need to revert that change.

Instead, we can use the functionality from https://github.com/airbnb/swift/pull/180 to use our `// swift-tools-version:5.3` configuration in `Package.swift` to configure the Swift version used by SwiftFormat. This lets us revert the changes to the podspec.


Package.swift
```diff
@@ -9,5 +9,5 @@ let package = Package(
 
 #if swift(>=5.6)
 // Add the Airbnb Swift formatting plugin if possible
-package.dependencies.append(.package(url: "https://github.com/airbnb/swift", .branch("master")))
+package.dependencies.append(.package(url: "https://github.com/airbnb/swift", .branch("cal--swiftversion")))
```
calda, line 12: I'll change this to `.upToNextMajor(from: 1.0.1)` after https://github.com/airbnb/swift/pull/180 is merged / released

erichoracek (approved):
LGTM! Do we need to update the Package.resolved for the Example xcode project?

calda:
The Xcode project/workspace is separate from the `Package.swift` since Lottie also supports Carthage. That Xcode workspace doesn't include the package plugin as a dependency.

PR #1679, opened 2022-07-29
Fix performance test flakiness by reducing memory usage

calda:
The performance tests have been flaky for the past few days. This is probably because they were using an extremely large amount of memory, because the views / layers weren't being deallocated correctly.

| Before | After |
| ---- | ---- |
| Screen Shot 2022-07-29 at 9 07 02 AM | Screen Shot 2022-07-29 at 9 19 18 AM |

### DIFF
--- a/Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -20,21 +20,12 @@
         }
       },
       {
-        "package": "SourceKitten",
-        "repositoryURL": "https://github.com/jpsim/SourceKitten.git",
-        "state": {
-          "branch": null,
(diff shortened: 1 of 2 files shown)


Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -19,22 +19,13 @@
           "version": "0.7.0"
         }
       },
-      {
```
calda, line 22: Looks like I was wrong about https://github.com/airbnb/lottie-ios/pull/1674#issuecomment-1198430505

calda: image

erichoracek (approved):
LGTM!

PR #1688, opened 2022-08-02
Add support for Repeater shape items

calda:
This PR adds support for Repeater shape items to the Core Animation rendering engine. This seems like a highly requested feature: Fixes #575. Fixes #1316. Fixes #846. Fixes #1209. Fixes #1529. Fixes #769.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-02 at 10 01 48 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-02 at 09 56 16 |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-02 at 10 01 26 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-02 at 10 00 52 |

PR #1690, opened 2022-08-02
Support animating multiple properties of shape bezier paths

calda:
I noticed the animations added in #1688 were marked as unsupported by the Core Animation rendering engine, even though they rendered correctly.

This PR updates the Core Animation rendering engine to support animating multiple properties of shape bezier paths. Now those animations are fully supported by the new engine.

To render shapes, we basically have to combine multiple `KeyframeGroup<Vector3D>` / `KeyframeGroup<Vector1D>`s into a single `KeyframeGroup<BezierPath>`. Originally we implemented this by just taking the keyframe timing information from one component and discarding the rest. Instead we can use `Keyframes.combinedIfPossible` to merge together all of the keyframes groups if they have the same timing information.


Sources/Private/CoreAnimation/Animations/RectangleAnimation.swift
```diff
@@ -14,18 +14,48 @@ extension CAShapeLayer {
...
+    let cornerRadius: Vector1D
+  }
+
+  /// Creates a single array of animatable keyframes from the separate arrays of keyframes in this Rectangle
+  func combinedKeyframes(context: LayerAnimationContext) throws-> KeyframeGroup<Rectangle.Keyframe> {
+    let combinedKeyframes = Keyframes.combinedIfPossible(size, position, cornerRadius) { untypedValues in
+      Keyframe(
+        size: untypedValues[0] as! Vector3D,
+        position: untypedValues[1] as! Vector3D,
+        cornerRadius: untypedValues[2] as! Vector1D)
```
calda, line 45: If we had variadic generics we could have the input parameters be `(KeyframeGroup<Vector3D>, KeyframeGroup<Vector3D>, KeyframeGroup<Vector1D>)` and have the output parameters be `(Vector3D, Vector3D, Vector1D). Since we can't represent that in the type system today, we have to just output an `[Any]` and cast the parameters manually.

This implementation is either correct and will always work, and is incorrect and would always crash. Since our test suite tests these methods very extensively, imo it's safe to assume that this will always succeed in production as opposed to doing an `as?` optional unwrap and handing a hypothetical failure case.

Sources/Private/CoreAnimation/Animations/StarAnimation.swift
```diff
@@ -71,25 +62,72 @@ extension CAShapeLayer {
   {
     try addAnimation(
       for: .path,
-      keyframes: star.position.keyframes,
-      value: { position in
-        // We can only use one set of keyframes to animate a given CALayer keypath,
-        // so we currently animate `position` and ignore any other keyframes.
-        // TODO: Is there a way to support this properly?
```
calda, line 78: > // TODO: Is there a way to support this properly?

Good news, yes! 😝

Sources/Private/CoreAnimation/Animations/EllipseAnimation.swift
```diff
@@ -14,15 +14,44 @@ extension CAShapeLayer {
...
+    let position: Vector3D
+  }
+
+  /// Creates a single array of animatable keyframes from the separate arrays of keyframes in this Ellipse
+  func combinedKeyframes(context: LayerAnimationContext) throws-> KeyframeGroup<Ellipse.Keyframe> {
+    let combinedKeyframes = Keyframes.combinedIfPossible(size, position) { untypedValues in
+      Keyframe(
+        size: untypedValues[0] as! Vector3D,
+        position: untypedValues[1] as! Vector3D)
+    }
```
erichoracek, line 43: I'm a bit confused what's going on here, why are we force casting these to `Vector3D` if both params are of type `Vector3D`? Is this due to limitations around variadic parameters and generics? Wondering if we could instead just make multiple arities of this function with tuples, similar to how Combine works with operators, since this feels a bit brittle/unsafe as-is

erichoracek (approved):
Looks good, just one question about whether we can make it safer using generics

Sources/Private/CoreAnimation/Animations/EllipseAnimation.swift
erichoracek: Looks like you answered this before I refreshed the page, feel free to merge as-is if you'd like 👍

calda: Having multiple arities is an interesting idea, I'll try it out and see how it goes. It would definitely be easier to understand the callsite without this casting.

Issue #1696, opened 2022-08-04
`CoreTextRenderLayer` incorrectly renders some letters partially off the frame


A30008910-wei:
Since Lottie 3.1.9, `CoreTextRenderLayer` is rendering some single line text out of the drawing frame if the letter strokes to descent (e.g. `j`, `g` and `y`).

Note: the images below were captured with `CoreTextRenderLayer.drawDebug(_:)`

Tested on 

* Xcode 14 Beta 4, iOS 16 simulator
* Xcode 13.4.1, iOS 15.5 simulator
* Xcode 13.4.1, macOS 12.4 & macOS 12.5

## Which Version of Lottie are you using?

Lottie 3.4.1 (but issue exists since 3.1.9)

## Expected Behavior 

<img width="270" src="https://user-images.githubusercontent.com/89169304/182768903-853d9931-ca05-44ec-8286-2b655edfd8f0.png" />

It looks like a previously removed workaround in the commit 766f555fc4a5e5b1271f9dbd53d2500d47abb20a can be applied to `CoreTextRenderLayer` for a fix.

```Swift
    guard let font = font else { return }

    // ...

    // This fixes a vertical padding issue that arises when drawing some fonts.
    // For some reason some fonts, such as Helvetica draw with and ascender that is greater than the one reported by CTFontGetAscender.
    // I suspect this is actually an issue with the Attributed string, but cannot reproduce.

    if let fillFrame = fillFrame {
      var o = [CGPoint] (repeating: .zero, count: 1)
      let count = CFArrayGetCount(CTFrameGetLines(fillFrame))
      if count > 0 {
        CTFrameGetLineOrigins (fillFrame, CFRange (location: count-1,length: 1), &o)
        let diff = CTFontGetDescent(font) - o[0].y
        ctx.translateBy(x: 0, y: diff)
      }
    } else if let strokeFrame = strokeFrame {
      var o = [CGPoint] (repeating: .zero, count: 1)
      let count = CFArrayGetCount(CTFrameGetLines(strokeFrame))
      if count > 0 {
        CTFrameGetLineOrigins (strokeFrame, CFRange (location: count-1,length: 1), &o)
        let diff = CTFontGetDescent(font) - o[0].y
        ctx.translateBy(x: 0, y: diff)
      }
    }
```

## Actual Behavior

Note that the letter `g` is partially out of the frame.

[...]

PR #1697, opened 2022-08-04
Re-apply a workaround to `CoreTextRenderLayer`

weisunOW:
Re-apply the previously removed workaround around line origin adjustment from #1250.

This PR addresses https://github.com/airbnb/lottie-ios/issues/1696

It looks like the workaround is still required, see the comparison below,

The issue can be observed with [Tests/Samples/Nonanimating/verifyLineHeight.json](https://github.com/airbnb/lottie-ios/blob/master/Tests/Samples/Nonanimating/verifyLineHeight.json) on the latest Lottie release.
|Expected|Actual|
|---|---|
|expected|actual|


A30008910-wei:
> After this change the text has moved down and no longer appears visually centered. Would you mind taking a look?

Thanks @calda, I will have a look.

A30008910-wei:
> looking a bit closer, this diff in `Samples/Issues/issue_1448.json` seems like a problem:

Adding a `if diff > 0 { ... }` as in [c9fd88f](https://github.com/airbnb/lottie-ios/pull/1697/commits/c9fd88f298c54274b9258b8b7968352b9262f745) seems to fix the issue, now `issue_1448.json` is no longer shown as a change.

calda (approved):
Nice work, thank you!!

PR #1698, opened 2022-08-04
Fix issue where animation with negative stroke offset would render incorrectly

calda:
This PR fixes an issue where negative stroke offsets would render incorrectly. It looks like Core Animation doesn't support this, so we have to manually shift all of the offset values to be positive.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-04 at 09 19 17 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-04 at 10 01 14 |


erichoracek (approved):
LGTM!

PR #1699, opened 2022-08-04
Support combining shapes with different keyframe timings into the same CAShapeLayer

calda:
This PR adds support for combining shapes with different keyframe timings into the same `CALayer`.

We have to render all shapes in the same `CAShapeLayer`, or their fills may render incorrectly in places where the shapes overlap. We previously only did this when all of the shapes shared the exact same keyframe timing values.

In cases where the shapes have different keyframe timing values, we now combine them by manually interpolating the paths at each frame. This is similar to how we support path trims as of #1612.

| Before | After |
| ----- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-04 at 11 03 38 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-04 at 11 03 04 |


Sources/Private/CoreAnimation/Animations/CombinedShapeAnimation.swift
```diff
@@ -51,3 +51,25 @@ final class CombinedShapeItem: ShapeItem {
...
+
+extension CombinedShapeItem {
+  /// Manually combines the given shape keyframes by manually interpolating at each frame
+  static func manuallyInterpolating(
+    shapes: [KeyframeGroup<BezierPath>],
+    name: String,
+    context: LayerContext)
+    -> CombinedShapeItem
+  {
+    let interpolators = shapes.map { KeyframeInterpolator<BezierPath>(keyframes: $0.keyframes) }
```
erichoracek, line 63: Nice use of some infra from the main thread renderer! Wondering if we should move it to a shared place now

erichoracek (approved):
LGTM! Just one suggestion

Sources/Private/CoreAnimation/Animations/CombinedShapeAnimation.swift
calda: Good idea, I'll move this to `Sources/Private/Utility/Interpolatable`

PR #1705, opened 2022-08-08
Add workaround for issue where airbnb/swift dependency causes Xcode to spin indefinitely at 100% CPU load

calda:
This PR adds a workaround for issue where our dependency on the [airbnb/swift repo](https://github.com/airbnb/swift) causes Xcode to spin indefinitely at 100% CPU load. As discussed in https://github.com/airbnb/swift/pull/183 this is caused by the remote binary dependencies that package uses.

As a workaround, we can remove that dependency from our `Package.swift`. When running `bundle exec rake format:swift` / `bundle exec rake lint:swift`, we can temporarily add the dependency, run the linter/formatter, and then remove the dependency.


Rakefile
```diff
@@ -119,3 +119,40 @@ def xcodebuild(command)
     sh "xcodebuild #{command}"
   end
 end
+
+def formatTool(command)
+  # As of Xcode 13.4 / Xcode 14 beta 4, including airbnb/swift as a dependency
+  # causes Xcode to spin indefinitely at 100% CPU (due to the remote binary dependencies
+  # used by that package). As a workaround, we can specifically add that dependency
+  # to our Package.swift file when linting / formatting and remove it afterwards.
+  packageDefinition = File.read('Package.swift')
```
erichoracek, line 128: Does this work if you execute the rake task from a subdirectory?

erichoracek (approved):
@calda LGTM! Just one minor question

Rakefile
calda: it does!

calda: I'm actually surprised lol, I didn't test it originally and I expected it to not work when I tried it just now

Rakefile
```diff
@@ -119,3 +119,40 @@ def xcodebuild(command)
...
+  packageDefinitionWithFormatDependency = packageDefinition +
+  <<~EOC
+  
+  #if swift(>=5.6)
+  // Add the Airbnb Swift formatting plugin if possible
+  package.dependencies.append(
+    .package(
+      url: "https://github.com/airbnb/swift",
+      .upToNextMajor(from: "1.0.1")))
```
calda, line 137: Oh I almost forgot, since we aren't checking in a `Package.resolved` for this we need to list a specific commit here

calda:
```suggestion
      // Since we don't have a Package.resolved for this, we need to reference a specific commit
      // so changes to the style guide don't cause this repo's checks to start failing.
      .revision("cec29280c35dd6eccba415fa3bfc24c819eae887")))
```

PR #1706, opened 2022-08-09
Fall back to Main Thread rendering engine when attempting to apply `Trim` to `Fill`

calda:
I was browsing the issue backlog and noticed an animation that rendered differently using `.mainThread` and `.automatic`. This turned out to be because the animation was applying a `Trim` to a `Fill`ed shape, which is not natively supported by Core Animation (`Trim`s normally correspond to animations on `strokeStart` / `strokeEnd` for a `Stroke`ed shape). 

| `.automatic`, before | `.automatic`, after |
| ---- | ---- |
| Simulator Screen Shot - iPhone 8 - 2022-08-08 at 17 13 13 | Simulator Screen Shot - iPhone 8 - 2022-08-08 at 17 13 06 |


Sources/Private/CoreAnimation/Animations/ShapeAnimation.swift
```diff
@@ -134,7 +134,7 @@ extension Trim {
 
     // If maximum stroke value is larger than 100%, then we have to create copies of the path
     // so the total path length includes the maximum stroke
-    let startStrokes = adjustedStrokeEnd.keyframes.map { $0.value.cgFloatValue }
+    let startStrokes = adjustedStrokeStart.keyframes.map { $0.value.cgFloatValue }
```
calda, line 137: unrelated but I noticed this typo from a previous PR

erichoracek (approved):

PR #1707, opened 2022-08-09
Fix bug in `AnimationView.play(fromFrame:toFrame:)` with `fromFrame: nil`

### LINKED ISSUE #1233: Default `fromFrame` is incorrect

## This issue is a:
- [ ] Crashing Bug
- [ ] Feature Request

Lottie 3.1.8

## What Platform are you on? 
- [ ] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [ ] Objective-C

## Expected Behavior 
Calling `AnimationView`'s `play(fromFrame:toFrame:loopMode:completion:)` method without specifying the `fromFrame` parameter should play the animation from the current frame.

Also, calling `play(completion:)` when the current frame is not at the start should play the animation from the current frame.

## Actual Behavior
When `play(fromFrame:toFrame:loopMode:completion:)` is called, the animation is played from the frame number that is equivalent to the current "progress".

When `play(completion:)` is called, the animation is played from the start.

## Code Example
```
let animationView = AnimationView(name: "animationWith100Frames")
view.addSubview(animationView)
animationView.currentFrame = 50
// Note: current progress should now be 0.5, because frame 50 is halfway into the animation
animationView.play(toFrame: 100) // Instead of starting at frame 50, the animation starts at frame 0.5
// Or using the line below instead
animationView.play() // Instead of starting at frame 50, the animation starts at frame 0
```

## Possible Solution
This should be simply fixed by changing `currentProgress` on [line 320 of AnimationView.swift](https://github.com/airbnb/lottie-ios/blob/master/lottie-swift/src/Public/Animation/AnimationView.swift#L320) to `currentFrame`.

And by changing `animation.startFrame` on [line 268 of AnimationView.swift](https://github.com/airbnb/lottie-ios/blob/master/lottie-swift/src/Public/Animation/AnimationView.swift#L268) to `currentFrame`.

calda:
This PR fixes an issue in `AnimationView.play(fromFrame:toFrame:)` where if you pass in `fromFrame: nil` it would fall back to using `fromFrame: currentProgress`. Based on the docs, this is supposed to be `fromFrame: currentFrame` (_"Parameter fromFrame: The start frame of the animation. If `nil` the animation will start at the current frame."_)

Fixes #1233.

### DIFF
--- a/Sources/Public/Animation/AnimationView.swift
+++ b/Sources/Public/Animation/AnimationView.swift
@@ -465,7 +465,7 @@ final public class AnimationView: AnimationViewBase {
     }
 
     let context = AnimationContext(
-      playFrom: fromFrame ?? currentProgress,
+      playFrom: fromFrame ?? currentFrame,
       playTo: toFrame,
       closure: completion)
     addNewAnimationForContext(context)


erichoracek (approved):
huh, definitely seems like a typo, surprised it made it this long 😅

PR #1712, opened 2022-08-10
Add support for inverted matte layers

### LINKED ISSUE #1655: Core Animation rendering engine doesn't support `MatteType.invert`

Lottie 3.4.0
## Issue
Hello there! I was experimenting with 3.4.0 (Core Animation) on iOS and I noticed that animated track matte layers do not seem to animate when tested with 3.4.0 

I am exporting with Bodymovin V 5.9.0

Animated mattes may not be supported yet, I am not sure if this is the case!

Thank you for all your efforts! Looking forward to using 3.4.0! 

## Animation JSON

https://lottiefiles.com/113328-square-matte-test
Here is a test that I used to troubleshoot.

calda:
This PR adds support for inverted matte layers to the Core Animation rendering engine. Fixes #1655.

Last time I looked in to this I couldn't figure out how to create an inverted mask from an arbitrary `CALayer`. All of the discussion online ([example](https://stackoverflow.com/a/42238699/2530060)) said to render the layers into an image and then use Core Image to apply the inverted mask. This is how the Main Thread rendering engine works, which is a source of performance issues. That approach doesn't work for the Core Animation engine, since it requires manually re-rendering the image every frame.

We recently added support for blend modes (#1585), which seemed promising for something like this. The [documentation](https://developer.apple.com/documentation/quartzcore/calayer/1410748-compositingfilter) for these is woefully inadequate,  it even says _"This property is not supported on layers in iOS"_ which is just no longer the case. I tried a bunch of the blend modes listed on [this page](https://developer.apple.com/documentation/coregraphics/cgblendmode) and found that `xor` does exactly what we need here 🎉 

| Before (non-inverted masks) | After (inverted masks!)  |
| ----- | ----- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-10 at 10 24 54 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-10 at 10 24 30 |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-10 at 10 25 02 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-10 at 10 25 20 |


Sources/Private/CoreAnimation/Layers/CALayer+setupLayerHierarchy.swift
```diff
@@ -96,15 +99,52 @@ extension CALayer {
...
+    case .invert:
+      guard let maskLayer = try matteLayerModel.makeAnimationLayer(context: context) else {
+        return nil
+      }
+
+      // We can invert the mask layer by having a large solid black layer with the
+      // given mask layer subtracted out using the `xor` blend mode. When applied to the
+      // layer being masked, this creates an inverted mask where only areas _outside_
+      // of the mask layer are visible.
+      // https://developer.apple.com/documentation/coregraphics/cgblendmode/xor
```
erichoracek, line 125: Very cool!

erichoracek (approved):
Nice!

Issue #1717, opened 2022-08-12
Trim Paths timing is out of sync + Alpha Inverted Matte is out of sync with its parent


jhughes696:
## Which Version of Lottie are you using?

Lottie 3.3.0
iOS 15.6
iPhone 12 mini

## Expected Behavior 
![ExpectedBehavior](https://user-images.githubusercontent.com/50207823/184448514-f7c1e8ce-5638-4c9b-bceb-d6dd2fcb55ab.gif)

## Actual Behavior
![iOS_bug](https://user-images.githubusercontent.com/50207823/184448524-62450ce8-a97b-4345-9e40-6cbb9a408285.gif)

## Animation JSON
[https://lottiefiles.com/share/thkvqxz3](url)

calda:
> Secondly, the Alpha Inverted Matte is out of sync, creating a stuttering effect. The Alpha Inverted Matte is parented to another layer, but it's not moving in sync with its parent. So it's creating a stuttering, time delay that's visually apparent due to the Matte layer's position being out of sync with it's parent.
>
> Lastly, when the AlphaInverted Matte is in use, it crashes my LottieFiles previewer app on iOS. For some reason, the Alpha Matte seems to be too much to process?

These issues are fixed by the Core Animation rendering engine, which has much better performance and now supports inverted mattes (as of https://github.com/airbnb/lottie-ios/pull/1712). 

Unfortunately the issue with the stroke timing not matching the matte position also happens in the Core Animation rendering engine. I see issues from a few years ago mentioning similar issues, so there must be a bug somewhere in the foundation of where Lottie handles keyframe timings. If anyone is interested in digging in to this issue, it would be much appreciated!

Also unfortunately, there's a bug in the Core Animation rendering engine that causes the stroke to only render for half of the animation. This also happens with `RenderingEngineOption.automatic`, so I'll get this issue fixed asap. Thanks for the report!

jhughes696:
Oh nice! So for the matte, I just need to make sure my client's eng team is using [Lottie 3.4.0](https://github.com/airbnb/lottie-ios/discussions/1627), is that correct? 

Thanks for looking into/drawing attention to the Trim Paths bug. That would be a huge help to find a fix 🙌🏼

calda:
> Oh nice! So for the matte, I just need to make sure my client's eng team is using https://github.com/airbnb/lottie-ios/discussions/1627, is that correct?

[...]

PR #1721, opened 2022-08-16
Fix issue where large `GradientFill` / `GradientStroke` would be cut off

calda:
This PR fixes an issue I noticed in #1717 where the large `GradientStroke` was being cut off, because it was larger than the hardcoded `GradientRenderLayer.gradientPadding`.

| Before | After |
| ----- | ----- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-16 at 10 16 49 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-16 at 10 16 28 |


erichoracek (approved):
LGTM!

PR #1727, opened 2022-08-23
Fix CoreAnimation `playbackState` race condition

oblador:
#1682 introduced a bug where infinitely looping animations would sometimes report as not playing. I think this is because the property is read while a new animation is scheduled but haven't yet been applied through the `display` lifecycle method. This PR changes the `playbackState` to take the pending animation configuration into consideration which fixes the bug on our end, but happy to see if you have a better approach at fixing this.

### DIFF
--- a/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
+++ b/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
@@ -329,11 +329,18 @@ extension CoreAnimationLayer: RootAnimationLayer {
   }
 
   var isAnimationPlaying: Bool? {
-    switch playbackState {
+    switch pendingAnimationConfiguration?.playbackState {
     case .playing:
-      return pendingAnimationConfiguration?.playbackState == .playing || animation(forKey: #keyPath(animationProgress)) != nil
-    case nil, .paused:
+      return true
+    case .paused:
       return false
+    case nil:
+      switch playbackState {
+      case .playing:
+        return animation(forKey: #keyPath(animationProgress)) != nil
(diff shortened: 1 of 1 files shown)


calda (approved):
This seems reasonable to me, especially for behavior compatibility with the main thread rendering engine that doesn't have this two-phase animation setup flow.

Sources/Private/CoreAnimation/CoreAnimationLayer.swift
```diff
@@ -331,7 +331,7 @@ extension CoreAnimationLayer: RootAnimationLayer {
   var isAnimationPlaying: Bool? {
     switch playbackState {
     case .playing:
-      return animation(forKey: #keyPath(animationProgress)) != nil
+      return pendingAnimationConfiguration?.playbackState == .playing || animation(forKey: #keyPath(animationProgress)) != nil
```
calda, line 334: Why does this `pendingAnimationConfiguration?.playbackState == .playing` check need to be nested in the `playbackState == .playing` branch? 

What happens if `pendingAnimationConfiguration?.playbackState == .playing` when `playbackState == .paused` or `nil`? Is there a reason we only have to do this check when another animation is already playing?

oblador: That's a good point, I guess I was too focused on that this is the line that changed and therefore I wanted to put it there. I'll push a different proposal and you can react to that :-D

calda: Updates lgtm, thanks!

PR #1733, opened 2022-08-26
Added ability to use a custom URLSession in the Animation Downloader

john-brunelle:
Simply added a parameter to the `loadedFrom{` function in the public Animation class extension that allow for a custom URLSession to be passed in. This is helpful for setting such things as request timeout and headers if needed.


calda (approved):
Thanks!

PR #1734, opened 2022-08-26
Fix issue where `CAGradientLayer` locations / color stops were being ignored

calda:
This PR fixes an issue where `CAGradientLayer` locations / color stops were being ignored. I'm pretty surprised it took us this long to find out this wasn't working, considering how many gradients we have in the test suite. I guess none of those use custom gradient locations?

| Before | After |
| ----- | ----- |
| Simulator Screen Shot - iPhone 8 - 2022-08-26 at 16 14 50 | Simulator Screen Shot - iPhone 8 - 2022-08-26 at 16 14 19 |
| Simulator Screen Shot - iPhone 8 - 2022-08-26 at 16 12 43 | Simulator Screen Shot - iPhone 8 - 2022-08-26 at 16 11 57 |


Tests/SnapshotConfiguration.swift
```diff
@@ -47,7 +47,6 @@ extension SnapshotConfiguration {
     "Nonanimating/FirstText": .precision(0.99),
     "Nonanimating/verifyLineHeight": .precision(0.99),
     "Nonanimating/blend_mode_test": .precision(0.99),
-    "Issues/issue_885": .precision(0.99),
```
calda, line 50: I removed this animation because its been flaky and showing up in diffs every time I regenerate snapshots with `isRecording = true`

erichoracek (approved):
😅

PR #1736, opened 2022-08-29
Fix performance issue with line dash patterns

### LINKED ISSUE #1735: Specific animation saturates main thread/freezes app in all rendering modes

Lottie 3.4.2, IOS 15.6.1, Xcode 13.4.1

## Expected Behavior 
Present normally

## Actual Behavior
Caveats:

-Project has `LottieConfiguration.shared.renderingEngine = .coreAnimation` & `LottieConfiguration.shared.decodingStrategy = .dictionaryBased` set, however this occurs with both the main thread and core animation renderers forced as well

Setup:

-Parent SwiftUI View

-Parent presents (via .fullScreenCover(isPresented:)) a child SwiftUI view that has the following Lottie animation.

-The animation was set to "loopOnce" and play() called inside a setup function which is called inside makeUIView() of a UIViewRepresentable container

-The app will freeze/main thread lock for 2-3 seconds in all conditions (debug with attached xcode, debug standalone, release, etc)

-Using _any other animation file_ in my app resolves the issue. Adding this animation back in triggers the behavior 100% of the time.

## Animation JSON
[check_off_pulse_f_green.json.zip](https://github.com/airbnb/lottie-ios/files/9445857/check_off_pulse_f_green.json.zip)

EDIT: 
-The freezing/main thread saturation behavior seems to happen in any context now, for example a page that's displaying one Lottie animation and then changes to this one will freeze the app.

EDIT 2:
-`LottieConfiguration.shared.decodingStrategy = .codable` does not resolve the issue either

calda:
This PR fixes a performance issue with line dash patterns. A `lineDashPattern` of all `0.01` values seems to causes Core Animation to have extremely poor performance, basically hanging the entire OS. Fixes #1735.


erichoracek (approved):
Oh boy 😅

PR #1746, opened 2022-09-12
Make `animationDidSet` `open`

Nirma:
Hi! Great library but I ran into an issue with `AnimatedSwitch`.

While using `AnimatedSwitch` I noticed that setting the animation triggers it to start but I think that as it is in other places there should be an option to set the animation without starting it.

In `AnimatedSwitch`'s base class `AnimatedControl` this is already defined as `open` so I think keeping it `open` in `AnimatedSwitch` and allowing people to subclass and define their own custom behavior would be very helpful.

### DIFF
--- a/Sources/Public/iOS/AnimatedSwitch.swift
+++ b/Sources/Public/iOS/AnimatedSwitch.swift
@@ -67,6 +67,12 @@ open class AnimatedSwitch: AnimatedControl {
     isAccessibilityElement = true
   }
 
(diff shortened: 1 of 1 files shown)


calda (approved):

calda:
@Nirma, you'll need to run the swift linter with `bundle exec rake format:swift` to make CI pass

Nirma:
@calda Thanks! Sorry for forgetting this step!

PR #1750, opened 2022-09-18
Don't play animation on initial assignment

Nirma:
Personally I think that playing the animation on an `AnimatedSwitch` should be triggered on state change and the initial call to `updateOnState` should complete without haptics or animation but is there a reason why `animated` is true?

If not then maybe it would be more predictable to have it not animate the initial animation assignment step. 

What do you think? 

Thanks!

### DIFF
--- a/Sources/Public/iOS/AnimatedSwitch.swift
+++ b/Sources/Public/iOS/AnimatedSwitch.swift
@@ -70,7 +70,7 @@ open class AnimatedSwitch: AnimatedControl {
   // MARK: Open
 
   open override func animationDidSet() {
-    updateOnState(isOn: _isOn, animated: true, shouldFireHaptics: false)
+    updateOnState(isOn: _isOn, animated: animateUpdateWhenChangingAnimation, shouldFireHaptics: false)
   }
 
   open override func endTracking(_ touch: UITouch?, with event: UIEvent?) {
@@ -90,6 +90,9 @@ open class AnimatedSwitch: AnimatedControl {
 
   /// The cancel behavior for the switch. See CancelBehavior for options
   public var cancelBehavior: CancelBehavior = .reverse
+
+  /// If `false` the switch will not play the animation when changing between animations.
+  public var animateUpdateWhenChangingAnimation = true
 
   public override var accessibilityTraits: UIAccessibilityTraits {
     set { super.accessibilityTraits = newValue }


calda (changes requested):
I don't think we can change this, since there may be users relying on this behavior. I'd consider this a breaking change from a behavior perspective, so if its something we want to do we'd have to save it for Lottie 4.0. If you think we should make this change in Lottie 4.0, could you start a Discussion post (ideally with example gifs showing the behavior before and after this change)?

If you need this change in Lottie 3.x, you could add a flag to `AnimatedSwitch` named something like `animateUpdateWhenChangingAnimation: Bool` that controls this, and defaults to `true`.

Nirma:
@calda Yeah this would be a breaking change if users are already relying on the current behavior.

I wasn't thinking of the case of using one `AnimatedSwitch` and swapping out the animation with different animations over the lifetime of the switch since mainly I have been using switches with the same animation getting toggled `on` and `off` throughout the lifetime of the switch.

Ok yeah if its possible what about introducing `animateUpdateWhenChangingAnimation` with a default value of `true` like you suggested? This way behavior remains the same but `AnimatedSwitch` not has the ability to change this parameter.

I made this change in c64f333fb4de77792137374c6f9538a560f8e257

Sources/Public/iOS/AnimatedSwitch.swift
```diff
@@ -70,7 +70,7 @@ open class AnimatedSwitch: AnimatedControl {
   // MARK: Open
 
   open override func animationDidSet() {
-    updateOnState(isOn: _isOn, animated: true, shouldFireHaptics: false)
+    updateOnState(isOn: _isOn, animated: false, shouldFireHaptics: false)
```
calda, line 73: Did you mean to change this to `animated: animateUpdateWhenChangingAnimation` instead of the one below?

Sources/Public/iOS/AnimatedSwitch.swift
```diff
@@ -101,7 +104,7 @@ open class AnimatedSwitch: AnimatedControl {
     set {
       /// This is forwarded to a private variable because the animation needs to be updated without animation when set externally and with animation when set internally.
       guard _isOn != newValue else { return }
-      updateOnState(isOn: newValue, animated: false, shouldFireHaptics: false)
+      updateOnState(isOn: newValue, animated: animateUpdateWhenChangingAnimation, shouldFireHaptics: false)
```
calda, line 107: This is a behavior change since it previously defaulted to `false` but now defaults to `true`

Nirma: @calda This is purely a slip up! Yes I did updated in f028aca5e7d992d2b0dab1889fa7c67eadcbb313

Sources/Public/iOS/AnimatedSwitch.swift
Nirma: @calda This is purely a slip up! Yes I did. updated in f028aca5e7d992d2b0dab1889fa7c67eadcbb313

Sources/Public/iOS/AnimatedSwitch.swift
Nirma: This is a mistake... sorry...

Nirma:
@calda I cleaned up the history and rebased down to one commit in: fd3e10099df585b8d4ec8b726277df0fcaea0013 look ok now? Thanks for your time.

PR #1756, opened 2022-09-26
Update snapshot testing package dependency

calda:
~~[swift-snapshot-testing](https://github.com/pointfreeco/swift-snapshot-testing) `1.10.0` includes changes that were previously only on `main`, so we can switch back to pointing to the public release version.~~

We actually can't use `1.10.0` right now, since it has a minimum `swift-tools-version` of 5.5 and we use 5.3.

Since the `main` branch we point to no longer supports 5.3, we have to pin this repo to specifically use the last commit on `main` that supported 5.3.


erichoracek (approved):
LGTM!

PR #1760, opened 2022-10-07
Bump version to 3.4.4

calda:
I'll release 3.4.4 today, with the set of changes on `master` since 3.4.3: https://github.com/airbnb/lottie-ios/compare/3.4.3...master


Example/Example.xcodeproj/project.pbxproj
```diff
@@ -435,7 +435,7 @@
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				GCC_C_LANGUAGE_STANDARD = gnu11;
 				INFOPLIST_FILE = iOS/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 14.1;
+				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
```
calda, line 438: `bundle exec rake build:example:all` is failing on master with _"compiling for macOS 10.10, but module 'Lottie' has a minimum deployment target of macOS 10.13"_, so I bumped these deployment targets. Not sure how this only just now started failing.

erichoracek (approved):

Example/Example.xcodeproj/project.pbxproj
calda: aha, this is because the macOS 12 GitHub Actions runner was just updated to use Xcode 14 (https://github.com/actions/runner-images/issues/6225). We'll also have to re-render the snapshots, since they seem to depend on the SDK / Xcode version.

PR #1761, opened 2022-10-07
Rename public types that conflict with SwiftUI types

### LINKED ISSUE #1420: Redundant conformance of 'Color'
Hello,
I'm getting error when triyng to extend the SwiftUI.Color because of the Color struc extension.

"redundant conformance of 'Color'"

Thanks in advance!

- [X] This issue isn't related to another open issue

## This issue is a:
- [] Crashing Bug
- [X] Feature Request

## What Platform are you on? 

- [] MacOS
- [X] iOS

## What Language are you in?
- [X] Swift
- [] Objective-C

## Expected Behavior 
I expect to extend SfiwftUI.Color wihtout issues.

## Actual Behavior
I'm getting the "redundant conformance of 'Color'" error because of the extension of the Color struct.

calda:
This PR renames `Animation` and `Color` to `LottieAnimation` and `LottieColor`, to resolve conflicts with the `SwiftUI` `Animation` and `Color` types. For consistency this PR also renames `AnimationView` to `LottieAnimationView`.

Fixes #1420.

We'll release Lottie 3.5 next week, with this being the only change from 3.4.4. This version includes typealiases for the old names, so there is no source breakage. Xcode emits a warning with a fix-it to update callsites to the new name.

In a few weeks we plan on releasing Lottie 4.0. We will _remove_ the old typealiases, which will resolve the naming conflict with SwiftUI.


Sources/Public/Animation/LottieAnimation.swift
```diff
@@ -16,16 +16,25 @@ public enum CoordinateSpace: Int, Codable {
 
 // MARK: - Animation
 
-/// The `Animation` model is the top level model object in Lottie.
+@available(
+  *,
+  deprecated,
+  renamed: "LottieAnimation",
+  message: "`Lottie.Animation` has been renamed to `LottieAnimation`, to prevent conflicts with the `SwiftUI.Animation` type.")
```
erichoracek, line 23: Maybe add a comment that this notice will be removed in Lottie 4.0?

Sources/Private/Model/Text/TextAnimator.swift
```diff
@@ -15,8 +15,8 @@ final class TextAnimator: Codable, DictionaryInitializable {
     let container = try decoder.container(keyedBy: TextAnimator.CodingKeys.self)
     name = try container.decodeIfPresent(String.self, forKey: .name) ?? ""
     let animatorContainer = try container.nestedContainer(keyedBy: TextAnimatorKeys.self, forKey: .textAnimator)
-    fillColor = try animatorContainer.decodeIfPresent(KeyframeGroup<Color>.self, forKey: .fillColor)
-    strokeColor = try animatorContainer.decodeIfPresent(KeyframeGroup<Color>.self, forKey: .strokeColor)
+    fillColor = try animatorContainer.decodeIfPresent(KeyframeGroup<LottieColor>.self, forKey: .fillColor)
+    strokeColor = try animatorContainer.decodeIfPresent(KeyframeGroup<LottieColor>.self, forKey: .strokeColor)
     strokeWidth = try animatorContainer.decodeIfPresent(KeyframeGroup<Vector1D>.self, forKey: .strokeWidth)
     tracking = try animatorContainer.decodeIfPresent(KeyframeGroup<Vector1D>.self, forKey: .tracking)
     anchor = try animatorContainer.decodeIfPresent(KeyframeGroup<Vector3D>.self, forKey: .anchor)
```
erichoracek, line 22: What do you think about also renaming the vectors too since they collide with Apple types in the Spatial framework?

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -86,16 +86,25 @@ extension LottieLoopMode: Equatable {
...
+  deprecated,
+  renamed: "LottieAnimationView",
+  message: "`AnimationView` has been renamed to `LottieAnimationView`, for consistency with the new `LottieAnimation` type.")
+public typealias AnimationView = LottieAnimationView
+
+// MARK: - LottieAnimationView
+
 @IBDesignable
-final public class AnimationView: AnimationViewBase {
+final public class LottieAnimationView: LottieAnimationViewBase {
```
erichoracek, line 99: Have you put any thoughts into what we'd name a SwiftUI `View` for Lottie animations? If both `LottieAnimation` and `LottieAnimationView` are taken I wonder what options we would have

erichoracek (approved):
LGTM! Just a few questions

Sources/Private/Model/Text/TextAnimator.swift
calda: Good idea, didn't know about that problem

Sources/Public/Animation/LottieAnimationView.swift
calda: Good question, `LottieView` is a potential option. How have we been handling this in other cases, where we provide a UIView and SwiftUI View with the same functionality?

Sources/Public/Animation/LottieAnimation.swift
calda: Good idea

Sources/Public/Animation/LottieAnimationView.swift
erichoracek: Sure, I think `LottieView` would work!

PR #1763, opened 2022-10-10
Fix issue where inverted mask layer was clipped incorrectly

calda:
This PR fixes an issue where an inverted mask layer was clipped incorrectly, because the mask shape extended outside the bounds of the layer that it was masking.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-10-10 at 16 34 03 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-10-10 at 16 33 36 |

You can see here the inverted mask layer (red) was smaller than the shape being cut out from it (the blue thumb in the left image), so the final inverted mask was incorrect. The fix is to make that layer larger.

| Mask, before | Mask, after |
| ----- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-10-10 at 15 00 24 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-10-10 at 16 31 52 |


Sources/Private/CoreAnimation/Layers/InfiniteOpaqueAnimationLayer.swift
```diff
@@ -0,0 +1,60 @@
...
+      1)
+  }
+
+  // MARK: Private
+
+  private let additionalPaddingLayer = CALayer()
+
+  /// Additional padding around `self.bounds` that should be filled in with `self.backgroundColor`
+  ///  - This specific value is arbitrary and can be increased if necessary.
+  private let additionalPadding: CGFloat = 10_000.0
```
erichoracek, line 58: "infinite" 😉

erichoracek: I do feel like we have a 10_000 somewhere else, do you think it makes sense to reference them to the same constant?

erichoracek (approved):

Sources/Private/CoreAnimation/Layers/InfiniteOpaqueAnimationLayer.swift
calda: Good question -- `GradientRenderLayer.gradientPadding` is also 10,000 and is conceptually the same value ("good enough" infinity) so I think you're probably right that it makes sense to use the same constant definition for both.

PR #1764, opened 2022-10-11
Use `RenderingEngineOption.automatic` by default

calda:
This PR updates `LottieConfiguration.shared` to use `RenderingEngineOption.automatic` by default.

Starting in Lottie 4.0, we default to using the Core Animation rendering engine for supported animations.


erichoracek (approved):
It's happening!!! 🎉

Issue #1565, opened 2022-05-06
Animation inside UIPageViewController doesn't play


marcelofabri:
## Check these before submitting:
- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [x] This issue isn't related to another open issue

## This issue is a:
- [ ] Non-Crashing Bug (Visual or otherwise)
- [ ] Crashing Bug
- [ ] Feature Request
- [x] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.0

## What Platform are you on? 

- [ ] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [ ] Objective-C

## Expected Behavior 

Animations after the second page inside an `UIPageViewController` don't play. A workaround is to set `backgroundBehavior = .pauseAndRestore`.

This is a regression from Lottie 2 to 3.

## Actual Behavior

They should play

## Code Example

https://github.com/airbnb/lottie-ios/issues/1084

marcelofabri:
I was able to repro it inside the Demo project in a page view controller with the `scroll` mode: https://github.com/marcelofabri/lottie-ios/pull/new/page-view-controller.

However I don't know if there's a good fix here since `backgroundBehavior ` fundamentally works differently. Maybe `pauseAndRestore` should've been the default behavior?

calda:
Closing since there's a known solution

calda:
In Lottie 4.0 we are changing the default behavior of `LottieBackgroundBehavior` from `.pause` to `.pauseAndRestore`, which should fix the underlying issue here.

Issue #1332, opened 2021-03-31
The animation stops when I navigate to other page and go back


IHNEL:
## Lottie-iOS Issue

I create a custom LottieView like below, it works fine for me with looping. However, when I navigate to other page then go back, the animation stops. I don't find anyway to restart or resume it.  Could someone help me please?

```
import SwiftUI
import Lottie
//Following this tutorial: https://designcode.io/swiftui-handbook-lottie-animation 
struct LottieView: UIViewRepresentable {
    var name = "success"
    var loopMode: LottieLoopMode = .loop
    
    func makeUIView(context: UIViewRepresentableContext<LottieView>) -> UIView {
        let view = UIView(frame: .zero)
        let animationView = AnimationView()
        let animation = Animation.named(name)
        animationView.animation = animation
        animationView.contentMode = .scaleAspectFit
        animationView.play()
        animationView.translatesAutoresizingMaskIntoConstraints = false
        animationView.loopMode = loopMode
        view.addSubview(animationView)
        NSLayoutConstraint.activate([
            animationView.heightAnchor.constraint(equalTo: view.heightAnchor),
            animationView.widthAnchor.constraint(equalTo: view.widthAnchor),
        ])
        return view
    }
    func updateUIView(_ uiView: UIView, context: UIViewRepresentableContext<LottieView>) {
    }
}

```

## Check these before submitting:
- [v] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)
- [v] This issue isn't related to another open issue

## This issue is a:
- [v] Non-Crashing Bug (Visual or otherwise)
- [] Crashing Bug
- [] Feature Request
- [] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.0

## What Platform are you on? 

- [] MacOS
- [v] iOS

## What Language are you in?
- [v] Swift
- [] Objective-C

BannaGo:
i meet the same problem

007-Heroa:
animationView.backgroundBehavior = .pauseAndRestore

BeyondChallenge:
same problem

fchiusolo:
+1

ssoubhik:
The same Problem is on TabView also.
Animation freezes if I slide back to a previous tab

bluerainxty:
i meet the same problem

ngay881855:
[...]

PR #1770, opened 2022-10-12
Update `LottieBackgroundBehavior.default` from `.pause` to `.pauseAndRestore` for Main Thread rendering engine

calda:
As discussed in https://github.com/airbnb/lottie-ios/discussions/1649, I've seen some confusion about `AnimationView.backgroundBehavior`. It currently defaults to `.pause` when using the main thread rendering engine, which seems unexpected to many users. For example:
 - https://github.com/airbnb/lottie-ios/issues/1332
 - https://github.com/airbnb/lottie-ios/issues/1565

We have an opportunity to update this in Lottie 4.0, so this PR changes the default from `.pause` to `.pauseAndRestore`. This matches the behavior of the Core Animation engine, which isn't paused when backgrounded.

### DIFF
--- a/Sources/Public/Animation/LottieAnimationView.swift
+++ b/Sources/Public/Animation/LottieAnimationView.swift
@@ -16,10 +16,11 @@ public enum LottieBackgroundBehavior {
   case stop
 
   /// Pause the animation in its current state. The completion block is called.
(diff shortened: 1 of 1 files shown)


erichoracek (approved):
🎉 will save so much time

PR #1772, opened 2022-10-17
Adds ability to not play the frame of the end marker (#1758)

lennartstolz:
LottieAnimationView: Adds ability to not play the frame of the provided end marker.

Discussion: https://github.com/airbnb/lottie-ios/discussions/1758


calda (approved):
Thanks!

PR #1777, opened 2022-10-17
Update LottieConfiguration to use DecodingStrategy. dictionaryBased by default

marcelofabri:
See https://github.com/airbnb/lottie-ios/discussions/1776

### DIFF
--- a/Sources/Public/LottieConfiguration.swift
+++ b/Sources/Public/LottieConfiguration.swift
@@ -152,7 +152,7 @@ public enum DecodingStrategy: Hashable {
   /// Use Codable. This is was the default strategy introduced on Lottie 3, but should be rarely
   /// used as it's slower than `dictionaryBased`. Kept here for any possible compatibility issues
   /// that may come up, but consider it soft-deprecated.
-  case codable
+  case legacyCodable
 
   /// Manually deserialize a dictionary into an Animation.
   /// This should be at least 2-3x faster than using Codable and due to that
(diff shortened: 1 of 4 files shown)


Sources/Public/LottieConfiguration.swift
```diff
@@ -149,11 +149,13 @@ extension RenderingEngine: RawRepresentable, CustomStringConvertible {
 
 /// How animation files should be decoded
 public enum DecodingStrategy: Hashable {
-  /// Use Codable. This is the default strategy introduced on Lottie 3.
+  /// Use Codable. This is was the default strategy introduced on Lottie 3, but should be rarely
+  /// used as it's slower than `dictionaryBased`. Kept here for any possible compability issues
+  /// that may come up, but consider it soft-deprecated.
```
marcelofabri, line 154: is this too strong?

calda: I expect we'll never completely _remove_ the `Codable` conformances on our models, since it seems reasonable to continue providing a `Codable` conformance for convenience (and since removing this would be source-breaking).

I do think it makes sense to call this case "legacy" or "deprecated" though, since it's no longer recommended. I think we may as well change it to `case legacyCodable` now. Thoughts?

calda (approved):

Sources/Public/LottieConfiguration.swift
marcelofabri: We could still have the `Codable` conformance, but eventually remove this convenience around using it. I'll rename it to `legacyCodable` for now.

Issue #1774, opened 2022-10-17
LRUAnimationCache should be smarter?


marcelofabri:
## Which Version of Lottie are you using?

Lottie 3.5.0

## Expected Behavior 

`LRUAnimationCache` should:

1. have a memory limit
2. evict when getting memory warnings

I'm curious why we're not using a `NSCache` instead.

## Actual Behavior

The limit is based on number of animations, which can widely vary. It also never clears any cached animations under memory pressure.

calda:
I'm open to providing alternative, smarter cache implementations, and updating the default cache used in our animation helper functions. I don't think we can remove or change the behavior of the existing `LRUAnimationCache` though, since that would be source breaking for consumers.

marcelofabri:
Yeah, we can keep it around, but having the right default is the most important thing IMO.

marcelofabri:
Also found https://github.com/airbnb/lottie-ios/pull/1411 which is pretty similar with what I had in mind

PR #1779, opened 2022-10-18
Allow customizing default AnimationCacheProvider

marcelofabri:
Motivated by https://github.com/airbnb/lottie-ios/issues/1774, I think it's important that we have a single customization point for which animation cache should be used, instead of using `LRUAnimationCache.sharedCache` everywhere.

### DIFF
--- a/Sources/Public/Animation/LottieAnimationViewInitializers.swift
+++ b/Sources/Public/Animation/LottieAnimationViewInitializers.swift
@@ -23,7 +23,7 @@ extension LottieAnimationView {
     name: String,
     bundle: Bundle = Bundle.main,
     imageProvider: AnimationImageProvider? = nil,
-    animationCache: AnimationCacheProvider? = LottieConfiguration.shared.animationCache,
+    animationCache: AnimationCacheProvider? = LottieAnimationCache.shared,
     configuration: LottieConfiguration = .shared)
   {
     let animation = LottieAnimation.named(name, bundle: bundle, subdirectory: nil, animationCache: animationCache)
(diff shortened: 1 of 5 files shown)


Sources/Public/Animation/LottieAnimationViewInitializers.swift
```diff
@@ -23,7 +23,7 @@ extension LottieAnimationView {
     name: String,
     bundle: Bundle = Bundle.main,
     imageProvider: AnimationImageProvider? = nil,
-    animationCache: AnimationCacheProvider? = LRUAnimationCache.sharedCache,
+    animationCache: AnimationCacheProvider? = LottieConfiguration.shared.animationCache,
```
marcelofabri, line 26: I don't love that we both have an `animationCache` param and a `configuration` one and the `animationCache` isn't coming from the `configuration`.

However I couldn't think about any good ways to deal with this since a `nil` `animationCache` is supported.

Sources/Public/AnimationCache/AnimationCacheProvider.swift
```diff
@@ -11,7 +11,7 @@ import Foundation
 /// can increase performance when loading an animation multiple times.
 ///
 /// Lottie comes with a prebuilt LRU Animation Cache.
-public protocol AnimationCacheProvider {
+public protocol AnimationCacheProvider: AnyObject {
```
marcelofabri, line 14: this is potentially a breaking change, but:

1. I don't think many people are rolling out their own cache providers
2. we kind of need a cache to be a reference type in order to avoid copying all over the place

Sources/Public/Animation/LottieAnimationViewInitializers.swift
calda: Hm, this is quite the gotcha. 

What if we created a separate `LottieAnimationCache.default` singleton that we use for this, instead of adding it to `LottieConfiguration`? 

That would have the added benefit of letting `LottieConfiguration` stay a simple value type with synthesized Equatable / Hashable conformances.

marcelofabri: Updated!

calda (approved):

PR #1785, opened 2022-10-19
Adds .lottie file load capability to LottieAnimation

eharrison:
Adds support to parse .lottie files
the dotLottie format (dotlottie.io) is a zip file with animation.json bundled along with image resources, and manifest

Motivation:
- Make .lottie file a new standard for lottie animations
- The update is already supported in Android player (https://github.com/airbnb/lottie-android/pull/1660)

Changes
- Added internal classes of ZIPFoundation (required to decompress the .lottie file)
- Added .lottie file parsing as a fallback to loading json animation data
- .lottie file also applies speed, loop mode and imageProviderSettings according to it's content


eharrison:
Hello! Here's Evandro. Head of mobile @ LottieFiles. Sending this as an effort to make .lottie more accessible. Please let me know if you see any issues. Thank you! :)

Sources/Public/Animation/DotLottie/ZIPFoundation/Archive+BackingConfiguration.swift
```diff
@@ -0,0 +1,128 @@
+//
+//  Archive+BackingConfiguration.swift
+//  ZIPFoundation
+//
+//  Copyright © 2017-2021 Thomas Zoechling, https://www.peakstep.com and the ZIP Foundation project authors.
+//  Released under the MIT License.
+//
+//  See https://github.com/weichsel/ZIPFoundation/blob/master/LICENSE for license information.
```
calda, line 8: I see you've included the source of ZIPFoundation directly. I think it would be preferable to incorporate ZIPFoundation through a package manager. Have you explored that approach?

Sources/Public/Animation/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,119 @@
...
+// Copyright © 2020 LottieFiles. All rights reserved.
+//
+
+import Foundation
+
+/// Detailed .lottie file structure
+public struct DotLottieFile {
+  public let localUrl: URL
+  
+  public static let manifestFileName: String = "manifest.json"
```
calda, line 15: Ideally we'd limit the set of new public APs to only include types / methods / properties that are strictly necessary for consumers of the library. For example, in this file it seems like everything should be internal other than the initializers.

Sources/Public/Animation/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,119 @@
...
+    return animation
+  }
+
+  /// Image provider for `LottieAnimationView`
+  public var imageProvider: FilepathImageProvider? {
+    !images.isEmpty ? FilepathImageProvider(filepath: imagesUrl) : nil
+  }
+    
+  /// Constructor with url.
+  /// Returns nil if is not a .lottie file and decompression failed
```
calda, line 75: This comment appears to be incorrect -- this initializer throws but never returns nil

Sources/Public/Animation/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,119 @@
...
+  public var imagesUrl: URL {
+    localUrl.appendingPathComponent("\(DotLottieFile.imagesFolderName)")
+  }
+  
+  /// All images in images folder
+  public var images: [URL] {
+    FileManager.default.urls(for: imagesUrl) ?? []
+  }
+  
+  /// Returns main (first) lottie animation of file
```
calda, line 56: If a .lottie file contains multiple animations, how does the user access them?

Sources/Public/Animation/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,119 @@
...
+  /// Constructor with url.
+  /// Returns nil if is not a .lottie file and decompression failed
+  /// - Parameters:
+  ///  - url: URL to .lottie file
+  public init(url: URL) throws {
+    self.localUrl = DotLottieUtils.animationsDirectoryURL(for: url)
+    guard url.isDotLottieFile else {
+      throw DotLottieError.invalidFileFormat
+    }
+    try decompress(from: url, in: localUrl)
```
calda, line 83: I see this unzips the .lottie file to a location in the temporary directory, and then leaves it there indefinitely (until the system removes it later). 

Since the temporary directory is a resource we share with the app using Lottie, I think we should be a bit more hygienic about this. I think ideally we would unzip the `.lottie` file to a temporary directory, load the contents of that directory into memory, and then delete the temporary directory. Thoughts?

Sources/Public/Animation/LottieAnimationHelpers.swift
```diff
@@ -154,13 +154,21 @@ extension LottieAnimation {
...
-      guard let dict = json as? [String: Any] else {
-        throw InitializableError.invalidInput
+      do {
+        let json = try JSONSerialization.jsonObject(with: data)
+        guard let dict = json as? [String: Any] else {
+          throw InitializableError.invalidInput
+        }
+        return try LottieAnimation(dictionary: dict)
+      } catch {
+        return try DotLottieFile(data: data).mainAnimation()
```
calda, line 170: Loading a `DotLottieFile` is a significantly more expensive operation that parsing a `LottieAnimation`, so I think this is potentially a surprising change. 

For example, I think it's probably acceptable in many circumstances to parse a `LottieAnimation` on the main thread, but I would never expect to decompress a zip file and perform other file operations on the main thread.

Since `.lottie` support is completely new functionality, I think it would be better to provide new public APIs specifically for loading `.lottie` files. These APIs should probably be asynchronous so the `.lottie` file decompression doesn't happen on the main thread.

Sources/Public/Animation/LottieAnimation.swift
```diff
@@ -157,4 +157,7 @@ public final class LottieAnimation: Codable, DictionaryInitializable {
   /// Markers
   let markers: [Marker]?
   let markerMap: [String: Marker]?
+    
+  /// DotLottie configuration to setup the player
+  var dotLottieConfiguration: DotLottieConfiguration?
```
calda, line 162: This seems reasonable to me. I like the idea of the `DotLottieFile` object returning an array of the `LottieAnimation`s that it contains. That lets us avoid having to add a bunch of new public methods on `LottieAnimationView`, which is nice.

Sources/Public/Animation/DotLottie/DotLottieManifest.swift
```diff
@@ -0,0 +1,48 @@
...
+// LottieFiles
+//
+// Created by Evandro Harrison Hoffmann on 27/06/2020.
+// Copyright © 2020 LottieFiles. All rights reserved.
+//
+
+import Foundation
+
+/// Manifest model for .lottie File
+public struct DotLottieManifest: Codable {
```
calda, line 12: This should be internal

Sources/Public/Animation/DotLottie/DotLottieAnimation.swift
```diff
@@ -0,0 +1,37 @@
+//
+// DotLottieAnimation.swift
+// Pods
+//
+// Created by Evandro Harrison Hoffmann on 28/06/2021.
+//
+
+import Foundation
+
+public struct DotLottieAnimation: Codable {
```
calda, line 10: This should probably be internal

Sources/Public/Animation/DotLottie/DotLottieUtils.swift
```diff
@@ -0,0 +1,127 @@
...
+// DotLottieUtils.swift
+// LottieFiles
+//
+// Created by Evandro Harrison Hoffmann on 27/06/2020.
+// Copyright © 2020 LottieFiles. All rights reserved.
+//
+
+import Foundation
+
+public struct DotLottieUtils {
```
calda, line 11: All of these helpers should be internal

calda (commented):
Left some comments. At a high level there are a few things we need to figure out:
 1. Can we incorporate ZIPFoundation using a package manager, instead of simply pasting its source into our repo?
 2. Can we load the `.lottie` file contents into memory and then delete the temporary directory we created when uncompressing it?
 3. Instead of modifying the existing methods that create `LottieAnimation`s, can we add a set of asynchronous methods specifically for loading `.lottie` files? From there the user could access any of the animations included in the file.
 
This will also need to include snapshot tests that render animations from `.lottie` files, and the code will need to be formatted according to [Airbnb's Swift style guide](https://github.com/airbnb/swift) (you can run `bundle exec rake format:swift` to do this automatically).

Sources/Public/DotLottie/Utility/DotLottieImageProvider.swift
```diff
@@ -0,0 +1,70 @@
...
+    imageCache.object(forKey: asset.name as NSString)
+  }
+
+  // MARK: Internal
+
+  let filepath: URL
+
+  // MARK: Private
+
+  private var imageCache: NSCache<NSString, CGImage> = .init()
```
calda, line 48: In this case I think we actually _shouldn't_ use `NSCache`.

The system will remove entires from an `NSCache` when under memory pressure. In our case we can't reload these images later, since the files no longer exists.

```suggestion
  private var images: [String: CGImage] = [:]
```

Sources/Public/DotLottie/Utility/DotLottieManifest.swift
```diff
@@ -0,0 +1,54 @@
...
+// LottieFiles
+//
+// Created by Evandro Harrison Hoffmann on 27/06/2020.
+// Copyright © 2020 LottieFiles. All rights reserved.
+//
+
+import Foundation
+
+/// Manifest model for .lottie File
+struct DotLottieManifest: Codable {
```
calda, line 12: These are a few files here under `/Sources/Public/DotLottie` that don't have any public declarations. Could you move them to something like `Sources/Private/DotLottie`?

Tests/SnapshotTests.swift
```diff
@@ -235,18 +249,28 @@ extension SnapshotConfiguration {
...
+        dotLottie: lottie,
+        configuration: configuration,
+        logger: logger)
+    } else {
+      return nil
+    }
 
     // Set up the animation view with a valid frame
     // so the geometry is correct when setting up the `CAAnimation`s
-    animationView.frame.size = animation.snapshotSize
```
calda, line 249: I don't think we can remove this without breaking the snapshot tests

Sources/Public/DotLottie/DotLottie.swift
```diff
@@ -0,0 +1,153 @@
...
+public final class DotLottie {
+
+  // MARK: Lifecycle
+
+  /// Loads `DotLottie` from `URL` containing .lottie file.
+  ///
+  /// - Parameters:
+  ///  - url: URL to .lottie file
+  ///  - Returns: Deserialized `DotLottie`. Optional.
+  init(url: URL) throws {
```
calda, line 24: Is it possible to make these initializers `async`?

calda: These initializers to a lot more work than `LottieAnimation.init` so it would be nice to help folks avoid calling this on the main thread

Sources/Public/DotLottie/Utility/DotLottieAnimation.swift
```diff
@@ -0,0 +1,51 @@
...
+  var speed: Double
+
+  /// 1 or -1
+  var direction: Int? = 1
+
+  /// mode - "bounce" | "normal"
+  var mode: String? = "normal"
+
+  /// URL to animation, to be set internally
+  public var animationUrl: URL?
```
calda, line 30: Does this need to be public?

Sources/Public/DotLottie/DotLottieHelpers.swift
```diff
@@ -0,0 +1,200 @@
...
+
+import Foundation
+
+extension DotLottie {
+
+  /// A closure for an Animation download. The closure is passed `nil` if there was an error.
+  public typealias DotLottieDownloadClosure = (DotLottie?) -> Void
+
+  /// Returns the list of `DotLottieAnimation` in the file
+  public var dotLottieAnimations: [DotLottieAnimation] {
```
calda, line 16: Does this need to be public? It seems like it's only used internally when loading the dotLottie archive contents in to memory

Sources/Public/DotLottie/Utility/DotLottieAnimation.swift
```diff
@@ -0,0 +1,51 @@
+//
+// DotLottieAnimation.swift
+// Pods
+//
+// Created by Evandro Harrison Hoffmann on 28/06/2021.
+//
+
+import Foundation
+
+public struct DotLottieAnimation: Codable {
```
calda, line 10: Does this need to be public?

Sources/Public/DotLottie/Utility/DotLottieImageProvider.swift
```diff
@@ -0,0 +1,70 @@
...
+#if os(iOS) || os(tvOS) || os(watchOS) || targetEnvironment(macCatalyst)
+import UIKit
+#elseif os(macOS)
+import AppKit
+#endif
+
+// MARK: - DotLottieImageProvider
+
+/// Provides an image for a lottie animation from a provided Bundle.
+public class DotLottieImageProvider: AnimationImageProvider {
```
calda, line 18: Does this need to be public?

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -402,6 +427,30 @@ final public class LottieAnimationView: LottieAnimationViewBase {
...
+  /// be loaded up and set to the beginning of its timeline.
+  /// The loopMode, animationSpeed and imageProvider will be set according
+  /// to lottie file settings
+  /// - Parameters:
+  ///   - dotLottie: Lottie file to play
+  ///   - animationId: Internal animation id to play. Optional
+  ///   Defaults to play first animation in file.
+  public func setDotLottie(
+    _ dotLottie: DotLottie?,
+    animationId: String? = nil)
```
calda, line 441: I wonder if a name like this would be more idiomatic:

```suggestion
  public func loadAnimation(
    _ animationId: String? = nil,
    from _ dotLottieFile: DotLottie)
```

I'd like to avoid "set" if possible. Open to other ideas as well.

calda (commented):
Looking better!

calda:
In noticed CI is failing when using Xcode 12, presumably because support SPM dependencies in Xcode projects wasn't added until Xcode 13. 

In https://github.com/airbnb/lottie-ios/pull/1787 I bumped the minimum supported Xcode version to Xcode 13 (which bumps the minimum supported Swift version to 5.5). This should fix that CI failure, and will also let use use newer Swift features like async / await.

eharrison:
Amazing! Thank you for the comments! I'll check them out.  
Also, I may need help with Carthage, haven't done this before and not sure how the dependency is added.

calda:
Hah, I _also_ don't know how to add a dependency to a Carthage library. I can help take a look. If we can't get it to work by default we may have to wrap the ZIPFoundation code in `#canImport` directives and require Carthage consumers to integrate ZIPFoundation manually.

Sources/Public/DotLottie/DotLottie.swift
```diff
@@ -0,0 +1,153 @@
...
+// Copyright © 2020 LottieFiles. All rights reserved.
+//
+
+import Foundation
+import ZIPFoundation
+
+// MARK: - DotLottie
+
+/// Detailed .lottie file structure
+public final class DotLottie {
```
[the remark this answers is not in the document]
eharrison: Yes, makes sense!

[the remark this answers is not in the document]
eharrison: I'll keep DotLottieFile

marcelofabri:
Can we explore ways to make this dependency opt-in since a lot of people won’t need it?

for example:
- Make this another module
- Expose a ZIPProvider protocol that consumers are expect to implement in their apps and inject in LottieConfiguration or something similar


Besides adding a dependency for people who might not need it, this could cause conflicts if another version of the dependency is already used by the app.

calda:
That makes sense to me. We could put all of the `.lottie`-related functionality in a separate `DotLottie` module that includes the ZIPFoundation dependency and has to be imported separately.

This should also fix our Carthage build issue, since we could simply not make the new module available on Carthage.

Thoughts @eharrison?

eharrison:
Yes, I agree that the dependency is not ideal. It can definitely cause problems for users that already use ZIPFoundation.

Although, it would be awesome to have dotLottie directly inside lottie-ios, making it easier to use. 

Perhaps, let me spend some time trying not to need the dependency, as it's literally called in a single place to decompress the file. There might be a simpler solution for this. I'll research and experiment.

Sources/Public/Animation/LottieAnimation.swift
```diff
@@ -157,4 +157,7 @@ public final class LottieAnimation: Codable, DictionaryInitializable {
   /// Markers
   let markers: [Marker]?
   let markerMap: [String: Marker]?
+
+  /// DotLottie configuration to setup the player
+  var dotLottieConfiguration: DotLottieConfiguration?
```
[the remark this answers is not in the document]
eharrison: Having this var helps on removing logic using the animations in the dotLottie file. It compiles information from manifest, image provider and DotLottieAnimation. Alternative I can think of would be to have this configuration in a separate dictionary inside DotLottieFile and then besides searching for the animation with id, also search for the configuration with id.

eharrison:
Hey hey, so I have added the necessary Zip dependency inside the app, cleaning up everything we don't need to make it run in the supported platforms. Let me know if you find any issue.

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -0,0 +1,204 @@
...
+      }
+      task.resume()
+    }
+  }
+
+  /// Returns animation with id
+  /// - Parameter id: id to animation. Specified in .lottie file manifest. Optional
+  /// Defaults to first animation
+  /// - Returns: LottieAnimation with id. Optional
+  public func animation(id: String? = nil) -> LottieAnimation? {
```
calda, line 185: I think this would be more idiomatic

```suggestion
  public func animation(for id: String? = nil) -> LottieAnimation? {
```

Sources/Public/Animation/LottieAnimation.swift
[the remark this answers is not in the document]
calda: > have this configuration in a separate dictionary inside DotLottieFile and then besides searching for the animation with id, also search for the configuration with id.

That sounds reasonable, I think this would be better

Sources/Public/DotLottie/Cache/DotLottieCacheProvider.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  DotLottieCacheProvider.swift
+//  Lottie
+//
+//  Created by Evandro Hoffmann on 20/10/22.
+//
+
+import Foundation
+/// `DotLottieCacheProvider` is a protocol that describes a DotLottie Cache.
```
calda, line 9:
```suggestion

/// `DotLottieCacheProvider` is a protocol that describes a DotLottie Cache.
```

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,152 @@
+//
+// DotLottie.swift
+// LottieFiles
+//
+// Created by Evandro Harrison Hoffmann on 27/06/2020.
+// Copyright © 2020 LottieFiles. All rights reserved.
```
calda, line 6: Can you remove the "Copyright © 2020 LottieFiles" lines from these files?

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,152 @@
...
+  ///  - Returns: Deserialized `DotLottie`. Optional.
+  init(data: Data, filename: String) throws {
+    fileUrl = DotLottieUtils.tempDirectoryURL.appendingPathComponent(filename.asFilename())
+    try decompress(data: data, to: fileUrl)
+  }
+
+  // MARK: Public
+
+  /// List of `LottieAnimation` in the file
+  public var animations: [LottieAnimation] = []
```
calda, line 45: Does this need to be public? I think it may be better to keep this internal and have consumers use the `DotLottieFile`-specific methods on `LottieAnimationView`. A `LottieAnimation` from a `DotLottieFile` isn't fully self-contained and may not play correctly without the full `DotLottieFile` context.

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,152 @@
...
+
+  /// Image provider for animations
+  public var imageProvider: AnimationImageProvider?
```
calda, line 48: Same here, it may make sense to keep this internal.

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,152 @@
...
+
+  // MARK: Internal
+
+  /// Manifest.json file loading
+  var manifest: DotLottieManifest? {
+    let path = fileUrl.appendingPathComponent(DotLottieFile.manifestFileName)
+    return try? DotLottieManifest.load(from: path)
```
calda, line 55: Should we make this `lazy` or something so it's only loaded once, instead of being loaded repeatedly?

calda: If not we may want to call this `loadManifest()` to make it clear that it's reading a file from disk each call

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,152 @@
...
+  /// - Parameters:
+  ///  - url: url to .lottie file
+  ///  - destinationURL: url to destination of decompression contents
+  private func decompress(from url: URL, to destinationURL: URL) throws {
+    try? FileManager.default.removeItem(at: destinationURL)
+    try FileManager.default.createDirectory(at: destinationURL, withIntermediateDirectories: true, attributes: nil)
+    try FileManager.default.unzipItem(at: url, to: destinationURL)
+    loadContent()
+    try? FileManager.default.removeItem(at: destinationURL)
+    try? FileManager.default.removeItem(at: url)
```
calda, line 104: Nice!

Sources/Public/DotLottie/Utility/DotLottieAnimation.swift
```diff
@@ -0,0 +1,52 @@
+//
+// DotLottieAnimation.swift
+// Pods
+//
+// Created by Evandro Harrison Hoffmann on 28/06/2021.
+//
+
+import Foundation
+
+struct DotLottieAnimation: Codable {
```
calda, line 10: These files with only `internal` declarations should be in `Sources/Private/Model/DotLottie`

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -0,0 +1,204 @@
...
+
+  /// Loads a DotLottie model from a bundle by its name. Returns `nil` if a file is not found.
+  ///
+  /// - Parameter name: The name of the lottie file without the lottie extension. EG "StarAnimation"
+  /// - Parameter bundle: The bundle in which the lottie is located. Defaults to `Bundle.main`
+  /// - Parameter subdirectory: A subdirectory in the bundle in which the lottie is located. Optional.
+  /// - Parameter dotLottieCache: A cache for holding loaded lotties. Defaults to `LRUDotLottieCache.sharedCache`. Optional.
+  ///
+  /// - Returns: Deserialized `DotLottie`. Optional.
+  public static func named(
```
calda, line 27: Can we make this `async` since it's doing some heavy file operations? We don't want consumers to be calling this from the main thread.

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -0,0 +1,152 @@
...
+public final class DotLottieFile {
+
+  // MARK: Lifecycle
+
+  /// Loads `DotLottie` from `URL` containing .lottie file.
+  ///
+  /// - Parameters:
+  ///  - url: URL to .lottie file
+  ///  - Returns: Deserialized `DotLottie`. Optional.
+  init(url: URL) throws {
```
calda, line 23: Can we make this `async` since it's doing some heavy file operations? We don't want this to be called from the main thread

calda (commented):
Looking good! Mostly minor comments. I agree that embedding the pieces of functionality from ZIPFoundation that we need as source is the best approach here, thanks for slimming it down to the bare essentials.

I still think the `DotLottieFile` initializer / `load` / `decompress` functions should be marked as `async` so they can't be used directly from the main thread. Thoughts?

calda (approved):
Thanks for the contribution!

rasberik:
Currently there is issue when `play()` method getting called but animation is not loaded yet. 
Especially for local `.lottie` files, 

``` swift 
// AnimationViewInitializers
DotLottieFile.named(name, bundle: bundle, dotLottieCache: dotLottieCache) { result in
      guard case Result.success(let lottie) = result else { return }
      self.loadAnimation(animationId, from: lottie)
    }
```
there is currently no way to know when it is ready to play or how to properly handle this. Most of my dotLottie files are not playing, just stuck in first frame because it returns in following block:

``` swift 
// LottieAnimationView
public func play(completion: LottieCompletionBlock? = nil) {
      return //here
    }
    ...
  }
```

I think its better add some additional ways to handle it, like may be something like `whenAnimationReady: () -> Void` which will be common for any `LottieAnimationView`?

rasberik:
@eharrison @calda

calda:
@rasberik, could you file an issue and ideally also provide a sample project that reproduces the issue?

rasberik:
> @rasberik, could you file an issue and ideally also provide a sample project that reproduces the issue?

Thanks, I prepared a PR here: https://github.com/airbnb/lottie-ios/pull/1802
Sample project creation might take me a while, files Im using cant be shared unfortunately 🙏

rasberik:
@calda 
Will this be enough as reproduction evidence?

https://github.com/rasberik/lottie-ios/pull/1/files

calda:
Helpful, thanks!

rasberik:
@calda I separated tests and added to https://github.com/airbnb/lottie-ios/pull/1802

PR #1787, opened 2022-10-21
Update minimum Swift version to 5.5

calda:
Lottie 4.0 is a good opportunity for us to update our minimum-supported Swift version.

This PR updates this repo to require Swift 5.5 and Xcode 13.0+

### DIFF
--- a/Sources/Private/CoreAnimation/Animations/GradientAnimations.swift
+++ b/Sources/Private/CoreAnimation/Animations/GradientAnimations.swift
@@ -193,12 +193,12 @@ extension GradientShapeItem {
       for colorIndex in 0..<numberOfColors {
         let colorStartIndex = colorIndex * 4
 
-        let colorLocation = Double(colorComponents[colorStartIndex])
+        let colorLocation = CGFloat(colorComponents[colorStartIndex])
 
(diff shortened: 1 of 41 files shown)


Sources/Private/CoreAnimation/Animations/GradientAnimations.swift
```diff
@@ -193,12 +193,12 @@ extension GradientShapeItem {
       for colorIndex in 0..<numberOfColors {
         let colorStartIndex = colorIndex * 4
 
-        let colorLocation = CGFloat(colorComponents[colorStartIndex])
+        let colorLocation = Double(colorComponents[colorStartIndex])
```
calda, line 196: banished `CGFloat` to the shadow realm

erichoracek (approved):

Sources/Private/CoreAnimation/Animations/GradientAnimations.swift
calda: One of these `CGFloat` changes causes the Xcode 13.2 build to fail with a weird and unhelpful `Illegal instruction: 4` error, so I reverted those changes

PR #1794, opened 2022-11-01
Add snapshot tests for dot lottie animations, and other cleanup

calda:
#1785 added support for dot lottie animations. This PR adds snapshot tests for these animations, plus other cleanup for the changes in that PR.


Lottie.xcodeproj/project.pbxproj
```diff
@@ -2658,7 +2658,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
 				GENERATE_INFOPLIST_FILE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 15.0;
```
calda, line 2661: Bumped the test target to 15.0 so we can use async/await in the tests

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -9,26 +9,40 @@ import Foundation
...
-  public typealias DotLottieLoadClosure = (Result<DotLottieFile, Error>) -> Void
-
-  // MARK: DotLottie file (Loading)
+  /// Loads a DotLottie model from a bundle by its name. Returns `nil` if a file is not found.
+  ///
+  /// - Parameter name: The name of the lottie file without the lottie extension. EG "StarAnimation"
+  /// - Parameter bundle: The bundle in which the lottie is located. Defaults to `Bundle.main`
+  /// - Parameter subdirectory: A subdirectory in the bundle in which the lottie is located. Optional.
+  /// - Parameter dotLottieCache: A cache for holding loaded lotties. Defaults to `LRUDotLottieCache.sharedCache`. Optional.
+  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
```
calda, line 18: Added `async` variants for all of these methods

calda:
fyi @eharrison

Sources/Public/Animation/LottieAnimation.swift
```diff
@@ -157,7 +157,4 @@ public final class LottieAnimation: Codable, DictionaryInitializable {
   /// Markers
   let markers: [Marker]?
   let markerMap: [String: Marker]?
-
-  /// DotLottie configuration to setup the player
-  var dotLottieConfiguration: DotLottieConfiguration?
```
calda, line 162: Removed this property in favor of a `DotLottieFile.Animation` type that includes a `LottieAnimation` and `DotLottieConfiguration`

Sources/Public/Animation/LottieAnimationHelpers.swift
```diff
@@ -164,6 +164,29 @@ extension LottieAnimation {
     }
   }
 
+  /// Loads a Lottie animation asynchronously from the URL.
+  ///
+  /// - Parameter url: The url to load the animation from.
+  /// - Parameter animationCache: A cache for holding loaded animations. Defaults to `LottieAnimationCache.shared`. Optional.
+  ///
+  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
+  public static func loadedFrom(
```
calda, line 173: I added `async` methods for loading `DotLottieFile`s so figured we may as well also add this one for remote `LottieAnimation`s

Sources/Private/Model/DotLottie/DotLottieImageProvider.swift
```diff
@@ -38,12 +38,12 @@ class DotLottieImageProvider: AnimationImageProvider {
...
 
   func imageForAsset(asset: ImageAsset) -> CGImage? {
-    imageCache.object(forKey: asset.name as NSString)
+    images[asset.name]
   }
 
   // MARK: Private
 
-  private var imageCache: NSCache<NSString, CGImage> = .init()
+  private var images = [String: CGImage]()
```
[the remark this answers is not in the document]
erichoracek: If we store the `filepath.url` couldn't we recover from a memory pressure eviction? It seems risky to have a bunch of `CGImage`s in memory and not be able to release them on memory pressure warning as it may lead to apps being jetsam'd

Sources/Public/Animation/LottieAnimationHelpers.swift
erichoracek: Nice!

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -9,26 +9,40 @@ import Foundation
...
   public static func named(
     _ name: String,
     bundle: Bundle = Bundle.main,
     subdirectory: String? = nil,
-    closure: @escaping DotLottieLoadClosure,
-    dotLottieCache: DotLottieCacheProvider? = DotLottieCache.sharedCache)
+    dotLottieCache: DotLottieCacheProvider? = DotLottieCache.sharedCache,
+    handleResult: @escaping (Result<DotLottieFile, Error>) -> Void)
   {
     DispatchQueue.global().async {
```
erichoracek, line 47: Seems like if you wanted you could potentially check for task cancellation in these methods via `Task.isCancelled`  to support async cancellation

erichoracek (approved):
@calda Good stuff! LGTM with a few suggestions/question

Sources/Private/Model/DotLottie/DotLottieImageProvider.swift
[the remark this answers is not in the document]
calda: When loading a dot lottie file, we currently:
 1. decompress the zip archive to a temporary directory
 2. load the file contents into memory
 3. delete the files in the temporary directory

I think deleting the temporary unzipped files after loading them is a good idea because:
 1. I think we don't want to be leaving behind a large number of files in the host app's temporary directory. The OS cleans these up in theory but I'm not sure we should rely on that.
 2. Since the OS can clean up those files at any time in theory, I don't think its safe to assume that they will still exist at any point in the future

That means that we don't have a way to gracefully reload the images if they get removed from the `NSCache`, since the filepath URLs are no longer valid.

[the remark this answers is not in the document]
calda: I suppose we could re-decompress the zip archive to reload the images if necessary after they're evicted from the cache, but that functionality doesn't current exist.

Tests/AnimationKeypathTests.swift
```diff
@@ -78,8 +82,8 @@ final class AnimationKeypathTests: XCTestCase {
     configuration: LottieConfiguration,
     function: String = #function,
     line: UInt = #line)
-  {
-    let hierarchyKeypaths = hierarchyKeypaths(animationName: animationName, configuration: configuration)
+  async {
```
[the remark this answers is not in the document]
calda: hah, yeah Swift Format is formatting `async` and `throws` effects pretty strangely, the only way I've been able to get this to not look funny is by putting the effects on the same line as the closing paren. At some point I'll probably need to tweak how SwiftFormat works here.

Sources/Private/Model/DotLottie/DotLottieImageProvider.swift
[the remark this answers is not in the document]
erichoracek: Gotcha, sounds good. Perhaps we could add a note about adding that behavior if needed?

PR #1795, opened 2022-11-02
Add Rounded Corners

johnny-duo:
This PR resolves issue #1523: 

https://user-images.githubusercontent.com/108750639/199562241-4510d730-42a7-4466-a78e-99f8efd88a84.mp4

PR #1796, opened 2022-11-02
Add rounded corner support to Core Animation rendering engine

calda:
This PR adds rounded corner support to the Core Animation rendering engine, following #1795 which added this support for the Main Thread rendering engine


calda:
fyi @johnny-duo

Sources/Private/Utility/Primitives/BezierPathRoundExtension.swift
```diff
@@ -43,11 +43,8 @@ extension CompoundBezierPath {
 }
 
 extension BezierPath {
-  // Round corners of a single bezier
+  // Computes a new `BezierPath` with each corner rounded based on the given `radius`
   func roundCorners(radius: CGFloat) -> BezierPath {
-    guard radius > 0 else {
```
calda, line 48: I had to remove this, because it caused artifacts when Core Animation tried to interpolate between a path with `roundedCorners(radius: 0)` to any other value

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-11-02 at 16 01 44 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-11-02 at 16 01 04 |

I added back the equivalent logic to the call site from the Main Thread rendering engine, which doesn't have this issue

erichoracek (approved):

PR #1801, opened 2022-11-07
Update Core Animation rendering engine to use device RGB color space

### LINKED ISSUE #1800: Improper background (white veil) with coreAnimation rendering engine.
There's a difference in the rendering of an animation between the `mainThread` and `coreAnimation` rendering engines on iOS.
When rendering with `coreAnimation`, a slight white veil appears on top of part of the screen during the whole animation.

The issue is reproduced on an iPad pro with iOS 16.1

Lottie 3.5.0 but the issue was already reproducible with 3.4.1

## Expected Behavior 
With `mainThread` rendering engine, the behaviour is correct.
IMG_0196

## Actual Behavior
With `coreAnimation` rendering engine, the behaviour is not correct.
IMG_0195

## Animation JSON
We have not made this JSON ourselves, it has been provided by our client.
[1440x1080.txt](https://github.com/airbnb/lottie-ios/files/9953417/1440x1080.txt)

Thanks for your help.
Aurélien.

calda:
This PR updates the Core Animation rendering engine to use the same color space as the Main Thread rendering engine (`CGColorSpaceCreateDeviceRGB()`). Fixes #1800.

Currently the Main Thread rendering engine and Core Animation rendering engine use different color spaces, which can cause the two engines to render colors differently. For consistency between the two implementations, we should use the same color space. 

`CGColorSpaceCreateDeviceRGB()` produces the exact same color as the web player:

Screen Shot 2022-11-07 at 10 04 00 AM

`CGColor(red:green:blue:alpha:)` creates a lighter color in this case:

Screen Shot 2022-11-07 at 10 04 15 AM

Since `CGColorSpaceCreateDeviceRGB()` is the existing behavior, and best matches other platforms, we should be using that instead.


Sources/Private/MainThread/LayerContainers/CompLayers/MaskContainerLayer.swift
```diff
@@ -103,8 +103,8 @@ private class MaskLayer: CALayer {
     addSublayer(maskLayer)
     anchorPoint = .zero
     maskLayer.fillColor = mask.mode == .add
-      ? CGColor(colorSpace: CGColorSpaceCreateDeviceRGB(), components: [1, 0, 0, 1])
-      : CGColor(colorSpace: CGColorSpaceCreateDeviceRGB(), components: [0, 1, 0, 1])
+      ? .rgb(1, 0, 0)
```
calda, line 106: Consolidated all `CGColor.init` call sites to the same helper

erichoracek (approved):

PR #1802, opened 2022-11-10
Add on `onAnimationReady` function support for delayed animation loading

rasberik:
## Problem scope

Currently there is issue `play()` method getting called after `LottieAnimationView` has been initialized, but `animation` is not loaded yet for newly added `.lottie` format, which needs file processing before animation is ready.

``` swift 
// AnimationViewInitializers
DotLottieFile.named(name, bundle: bundle, dotLottieCache: dotLottieCache) { result in
      guard case Result.success(let lottie) = result else { return }
      self.loadAnimation(animationId, from: lottie)
    }
```
there is currently no way to know when it is completed or way to properly handle it. 
I propose solution that doesn't require complete rewrite of view lifecycle after changing to `.lottie` format.

``` swift 
// LottieAnimationView
public func play(completion: LottieCompletionBlock? = nil) {
      // current behavior for play when using `.lottie` files is returned due to missing animation
      // behavior might success depend on file size that needs to be processed, not tested
      return
    }
    ...
  }
```

## How to Reproduce

Simply create `LottieAnimationView` using any local `.dotlottie` resource based initializer, bigger file size will guarantee the reproduction (30kb+). Key is to call play fast enough after calling the initializer

``` swift
let lottieAnimationView = LottieAnimationView(dotLottieName: resourceName, bundle: bundle)
// ... layout ...
lottieAnimationView.play()
```

You can refer to [added tests](https://github.com/airbnb/lottie-ios/pull/1802/files#diff-dba4e64f11b3b3a5d42dd612ab352daa4bf326c71f1aae47124ecd2c2cb21b60) that can confirm the problem.

## Proposed Solution

Until more proper alternative is found, I would like to suggest using functional method `onAnimationReady` that will return a closure with self (LottieAnimationView).
This helps to keep code changes to minimum, and works similarly to all local and remote types.
It does not alter existing behavior, and works as convenient method for those who might need it.

#### Before:
``` swift
func play(completion: CompletionHandler? {
   view.play(completion: completion)
}
```

#### After:
``` swift
func play(completion: CompletionHandler?) {
   lotView.onAnimationReady { view in
      view.play(completion: completion) // any type of `play(...)` or other methods on view can be called without self capture in most use cases
   }
}
```

### DIFF
--- a/Tests/AnimationReadyTests.swift
+++ b/Tests/AnimationReadyTests.swift
@@ -15,14 +15,13 @@ final class AnimationReadyTests: XCTestCase {
 
   func test_dotLottieFile_hasAnimation_onAnimationReady() {
     let exp = XCTestExpectation(description: "onAnimationReady expectation")
-    exp.expectedFulfillmentCount = 1
 
     let animationView = prepareDotLottieAnimationView()
     animationView.onAnimationReady { _ in
       XCTAssertNotNil(animationView.animation)
       exp.fulfill()
     }
-    wait(for: [exp], timeout: 0.1)
+    wait(for: [exp], timeout: 0.5)
   }
 
   func test_jsonFile_hasAnimation_onInitialize() {
@@ -32,7 +31,6 @@ final class AnimationReadyTests: XCTestCase {
 
   func test_jsonFile_hasAnimation_onAnimationReady() {
     let exp = XCTestExpectation(description: "onAnimationReady expectation")
-    exp.expectedFulfillmentCount = 1
 
     let animationView = prepareJsonAnimationView()
     animationView.onAnimationReady { _ in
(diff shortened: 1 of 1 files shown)


rasberik:
@calda Added tests to this PR, but we can remove them later as I think they dont add any specific value

Tests/AnimationReadyTests.swift
```diff
@@ -0,0 +1,76 @@
...
+
+@testable import Lottie
+
+@MainActor
+final class AnimationReadyTests: XCTestCase {
+
+  // MARK: Internal
+
+  func test_dotLottieFile_hasAnimation_onInitialize() {
+    XCTExpectFailure("This is expected failure due to loading delay")
```
rasberik, line 11: [Note] This is the underlying issue, that is different from `test_jsonFile_hasAnimation_onInitialize`

calda:
e.g. you will be able to do:

```swift
let animationView = LottieAnimationView(
  dotLottieName: "DotLottie/animation",
  completion: { animationView in
    animationView.play()
  })
```

PR #1810, opened 2022-11-11
Add completion callbacks to LottieAnimationView DotLottie initializers

calda:
When using the various `LottieAnimationView` DotLottie initializers, there is currently no way to actually configure the animation playback after the animation is loaded. This is because the `DotLottieFile` is loaded asynchronously, but  there was no callback for when it finished loaded:

```swift
let animationView = LottieAnimationView(
  dotLottieName: "DotLottie/animation",
  bundle: .module,
  subdirectory: Samples.directoryName)

XCTAssertNotNil(animationView.animation) // 🛑 not loaded yet, this is still nil
```

These initializers now take a callback which lets you configure the animation view once the animation is loaded:

```swift
let animationView = LottieAnimationView(
  dotLottieName: "DotLottie/animation",
  bundle: .module,
  subdirectory: Samples.directoryName,
  completion: { animationView, error in
    XCTAssertNotNil(animationView.animation) // ✅
  })
```

As suggested in #1802 we also provide a codepath that can be used to configure a `LottieAnimationView` once the animation finishes loading, regardless of the initializer called. This simplifies adoption of dot lottie files since there is a single codepath that works regardless of the animation format:

```swift
let animationView: LottieAnimationView

if loadDotLottieFile {
  // Loads the .lottie file asynchronously
  animationView = LottieAnimationView(dotLottieName: "animation")
} else {
  // Loads the .json file synchronously
  animationView = LottieAnimationView(name: "animation")
}

animationView.animationLoaded = { animationView, animation in
  // If using a .lottie file, this is called once the file finishes loading.
  // If using a .json file, this is called immediately (since the animation is loaded synchronously).
  animationView.play()
}
```

### DIFF
--- a/Sources/Public/Animation/LottieAnimationView.swift
+++ b/Sources/Public/Animation/LottieAnimationView.swift
@@ -225,6 +225,45 @@ final public class LottieAnimationView: LottieAnimationViewBase {
   public var animation: LottieAnimation? {
     didSet {
       makeAnimationLayer(usingEngine: configuration.renderingEngine)
+
+      if let animation = animation {
+        animationLoaded?(self, animation)
+      }
+    }
+  }
+
+  /// A closure that is called when `self.animation` is loaded. When setting this closure,
+  /// it is called immediately if `self.animation` is non-nil.
+  ///
+  /// When initializing a `LottieAnimationView`, the animation will either be loaded
+  /// synchronously (when loading a `LottieAnimation` from a .json file on disk)
+  /// or asynchronously (when loading a `DotLottieFile` from disk, or downloading
+  /// an animation from a URL). This closure is called in both cases once the
+  /// animation is loaded and applied, so can be a useful way to configure this
+  /// `LottieAnimationView` regardless of which initializer was used. For example:
+  ///
+  /// ```
+  /// let animationView: LottieAnimationView
+  ///
+  /// if loadDotLottieFile {
+  ///   // Loads the .lottie file asynchronously
+  ///   animationView = LottieAnimationView(dotLottieName: "animation")
+  /// } else {
+  ///   // Loads the .json file synchronously
+  ///   animationView = LottieAnimationView(name: "animation")
+  /// }
+  ///
+  /// animationView.animationLoaded = { animationView, animation in
+  ///   // If using a .lottie file, this is called once the file finishes loading.
+  ///   // If using a .json file, this is called immediately (since the animation is loaded synchronously).
+  ///   animationView.play()
+  /// }
+  /// ```
+  public var animationLoaded: ((_ animationView: LottieAnimationView, _ animation: LottieAnimation) -> Void)? {
+    didSet {
+      if let animation = animation {
+        animationLoaded?(self, animation)
+      }
     }
   }
 
--- a/Sources/Public/Animation/LottieAnimationViewInitializers.swift
+++ b/Sources/Public/Animation/LottieAnimationViewInitializers.swift
@@ -57,6 +57,7 @@ extension LottieAnimationView {
   public convenience init(
     url: URL,
     imageProvider: AnimationImageProvider? = nil,
+    session: URLSession = .shared,
     closure: @escaping LottieAnimationView.DownloadClosure,
     animationCache: AnimationCacheProvider? = LottieAnimationCache.shared,
     configuration: LottieConfiguration = .shared)
@@ -67,7 +68,7 @@ extension LottieAnimationView {
     } else {
       self.init(animation: nil, imageProvider: imageProvider, configuration: configuration)
 
-      LottieAnimation.loadedFrom(url: url, closure: { animation in
+      LottieAnimation.loadedFrom(url: url, session: session, closure: { animation in
         if let animation = animation {
           self.animation = animation
           closure(nil)
@@ -75,31 +76,6 @@ extension LottieAnimationView {
           closure(LottieDownloadError.downloadFailed)
         }
       }, animationCache: animationCache)
-    }
-  }
-
-  /// Loads a Lottie animation asynchronously from the URL
-  ///
-  /// - Parameter url: The url to load the animation from.
-  /// - Parameter imageProvider: An image provider for the animation's image data.
-  /// If none is supplied Lottie will search in the main bundle for images.
-  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-  public convenience init(
-    url: URL,
-    imageProvider: AnimationImageProvider? = nil,
-    animationCache: AnimationCacheProvider? = LottieAnimationCache.shared,
-    configuration: LottieConfiguration = .shared) async throws
-  {
-    if let animationCache = animationCache, let animation = animationCache.animation(forKey: url.absoluteString) {
-      self.init(animation: animation, imageProvider: imageProvider, configuration: configuration)
-    } else {
-      self.init(animation: nil, imageProvider: imageProvider, configuration: configuration)
-
-      guard let animation = await LottieAnimation.loadedFrom(url: url, animationCache: animationCache) else {
-        throw LottieDownloadError.downloadFailed
-      }
-
-      self.animation = animation
     }
   }
 
@@ -122,33 +98,6 @@ extension LottieAnimationView {
   }
 
   // MARK: DotLottie
-
-  /// Loads a Lottie animation from a .lottie file in the supplied bundle.
-  ///
-  /// - Parameter filePath: The name of the lottie file without the lottie extension. EG "StarAnimation"
-  /// - Parameter bundle: The bundle in which the lottie is located. Defaults to `Bundle.main`
-  /// - Parameter subdirectory: A subdirectory in the bundle in which the lottie is located. Optional.
-  /// - Parameter animationId: Animation id to play. Optional
-  /// Defaults to first animation in file
-  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-  public convenience init(
-    dotLottieName name: String,
-    bundle: Bundle = Bundle.main,
-    subdirectory: String? = nil,
-    animationId: String? = nil,
-    dotLottieCache: DotLottieCacheProvider? = DotLottieCache.sharedCache,
-    configuration: LottieConfiguration = .shared) async throws
-  {
-    self.init(dotLottie: nil, animationId: animationId, configuration: configuration)
-
-    let dotLottieFile = try await DotLottieFile.named(
-      name,
-      bundle: bundle,
-      subdirectory: subdirectory,
-      dotLottieCache: dotLottieCache)
-
-    loadAnimation(animationId, from: dotLottieFile)
-  }
 
   /// Loads a Lottie animation from a .lottie file in the supplied bundle.
   ///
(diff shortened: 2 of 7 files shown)


Sources/Public/Animation/LottieAnimationViewInitializers.swift
```diff
@@ -100,90 +125,195 @@ extension LottieAnimationView {
...
   /// Defaults to first animation in file
+  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
   public convenience init(
     dotLottieName name: String,
     bundle: Bundle = Bundle.main,
+    subdirectory: String? = nil,
     animationId: String? = nil,
     dotLottieCache: DotLottieCacheProvider? = DotLottieCache.sharedCache,
-    configuration: LottieConfiguration = .shared)
+    configuration: LottieConfiguration = .shared) async throws
```
calda, line 140: I'm not entirely sure how useful these `async` animation view initializers are, since in most use cases I expect you would rather do all of the view hierarchy setup in one pass and then further configure the animation view in the callback. Seems like it probably doesn't hurt to provide them though. Any thoughts?

Sources/Public/Animation/LottieAnimationViewInitializers.swift
```diff
@@ -198,3 +328,12 @@ extension LottieAnimationView {
...
 }
+
+// MARK: - Foundation.Bundle + Sendable
+
+/// Necessary to suppress warnings like:
+/// ```
+/// Non-sendable type 'Bundle' exiting main actor-isolated context in call to non-isolated
+/// static method 'named(_:bundle:subdirectory:dotLottieCache:)' cannot cross actor boundary
+/// ```
+extension Foundation.Bundle: @unchecked Sendable { }
```
calda, line 339: image 

😛

rasberik:
Thank you. 

Although unfortunately with this approach it will need major changes in current codebase in order to switch from json.

It is difficult to achieve proper decoupling where view doesn't care about type of animation, as it cant make the assumption that animation is ready or not. 

The use cases when filename and format is received from other sources of truth, and the responsibility of calling play is also decided by other sources, not necessarily same context with initialization. (Eg lazy view initialization and play on viewdidappear)

calda:
Good point, I see the appeal of supporting a single codepath for configuring an animation view regardless of what initializer was used. I'll add something like the `animationDidLoad: (Animation) -> Void` callback you suggested on Monday.

calda:
@rasberik, looking closer at the implementation I'm not sure if that change is a good idea.

One key problem is that when loading a `LottieAnimation`, `self.animation` is set synchronously before the `init` is finished. So when the consumer provides a value for the `animationDidLoad` closure, `self.animation` is already non-nil:

```swift
let animationView = LottieAnimationView(name: "LottieJsonFileName")

// self.animation is already non-nil when we set up this closure:

self.animationDidSet = { animation in
  // ...
}
```

The implementation in https://github.com/airbnb/lottie-ios/pull/1802 handles this by hopping between dispatch queues before calling `animationDidSet`, causing it to be called a few run loop cycles later. 

We'd want this `animationDidSet` closure to be called immediately when `self.animation` is updated. Otherwise there would always be few run loop cycles / frames between when the animation is loaded and the animation view itself is actually set up. But we can't use a synchronous `animationDidSet` closure with this `init` callback pattern unless we pass it in as a parameter to the initializer (which I think defeats the purpose).

rasberik:
@calda I understand, this is one of the reasons I chose slightly different name because animationDidLoad is not really on _didLoad_. 
Among alternatives, what do you think about giving optional configuration I mentioned above.

There is code on play that returns when animation is nil:
```
      return
    }
```
So by adding some optional parameters, it might be good idea to let users handle this case. This should not directly interfere with synchronous flow

``` swift
enum AnimationNotReadyBehavior {
  case interrupt
  case playWhenReady
}
```

calda:
I added an `animationLoaded` closure with the following behavior:
 1. when a new `animation` is set, `animationLoaded` is called
 2. when the `animationLoaded` closure is set, but `animation` is already non-nil`, `animationLoaded` is called
 
 That means this pattern works no matter which `LottieAnimationView` initializer you use:

```swift
let animationView: LottieAnimationView

if loadDotLottieFile {
  // Loads the .lottie file asynchronously
  animationView = LottieAnimationView(dotLottieName: "animation")
} else {
  // Loads the .json file synchronously
  animationView = LottieAnimationView(name: "animation")
}

animationView.animationLoaded = { animationView, animation in
  // If using a .lottie file, this is called once the file finishes loading.
  // If using a .json file, this is called immediately (since the animation is loaded synchronously).
  animationView.play()
}
```

@erichoracek, this PR is ready for review when you have a chance

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -225,6 +225,45 @@ final public class LottieAnimationView: LottieAnimationViewBase {
...
+  ///   // If using a .lottie file, this is called once the file finishes loading.
+  ///   // If using a .json file, this is called immediately (since the animation is loaded synchronously).
+  ///   animationView.play()
+  /// }
+  /// ```
+  public var animationLoaded: ((_ animationView: LottieAnimationView, _ animation: LottieAnimation) -> Void)? {
+    didSet {
+      if let animation = animation {
+        animationLoaded?(self, animation)
+      }
```
rasberik, line 266: [Q] After it is called, it is not set to nil and potentially can be called again if animation didSet triggers. I guess it is intended?

rasberik:
Thank you for updates, it is somewhat similar to the [original PR](https://github.com/airbnb/lottie-ios/pull/1802/files#diff-82a7a8ce67766a8c773f539f2150e0cc9ce2ca8d53fe0fcabd67516279c17f1eR613) approach 👍

Tests/AnimationViewTests.swift
```diff
@@ -0,0 +1,45 @@
...
+      completion: { animationView, error in
+        XCTAssertNil(error)
+        XCTAssertNotNil(animationView.animation)
+        expectation.fulfill()
+      })
+
+    XCTAssertNil(animationView.animation)
+    wait(for: [expectation], timeout: 1.0)
+  }
+
```
manishsilverlabs, line 72: extra space

Sources/Public/Animation/LottieAnimationViewInitializers.swift
Alain-17: Close all

calda:
ptal @erichoracek

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -225,6 +225,45 @@ final public class LottieAnimationView: LottieAnimationViewBase {
   public var animation: LottieAnimation? {
     didSet {
       makeAnimationLayer(usingEngine: configuration.renderingEngine)
+
+      if let animation = animation {
+        animationLoaded?(self, animation)
+      }
```
erichoracek, line 231: Can we check whether the animation `!==` to the previous animation before we do this work? Seems like as-is we could potentially call this more than we want

Sources/Public/Animation/LottieAnimationViewInitializers.swift
```diff
@@ -78,6 +78,31 @@ extension LottieAnimationView {
...
 
+  /// Loads a Lottie animation asynchronously from the URL
+  ///
+  /// - Parameter url: The url to load the animation from.
+  /// - Parameter imageProvider: An image provider for the animation's image data.
+  /// If none is supplied Lottie will search in the main bundle for images.
+  @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
+  public convenience init(
+    url: URL,
+    imageProvider: AnimationImageProvider? = nil,
```
erichoracek, line 89: I'd imagine we want to pass through the `session: URLSession = .shared` to the `LottieAnimation.loadedFrom` to allow consumers to specify custom headers etc.

Sources/Public/Animation/LottieAnimationViewInitializers.swift
```diff
@@ -78,6 +78,31 @@ extension LottieAnimationView {
...
+    configuration: LottieConfiguration = .shared) async throws
+  {
+    if let animationCache = animationCache, let animation = animationCache.animation(forKey: url.absoluteString) {
+      self.init(animation: animation, imageProvider: imageProvider, configuration: configuration)
+    } else {
+      self.init(animation: nil, imageProvider: imageProvider, configuration: configuration)
+
+      guard let animation = await LottieAnimation.loadedFrom(url: url, animationCache: animationCache) else {
+        throw LottieDownloadError.downloadFailed
+      }
```
erichoracek, line 100: It seems like we should propagate the error from the `loadedFrom` call, as-is it seems to be discarded outright which makes it impossible to get any insight into this error

Sources/Public/Animation/LottieAnimationViewInitializers.swift
erichoracek: Hmm I would say it's probably an anti-pattern to construct a view asynchronously since that would prevent adding the view to the view hierarchy until its animation has loaded. Has anyone requested these initializers? I think creating an internal `Task` that loads the animation would be the expectation, similar to how `AsyncImage` works

Sources/Public/Animation/LottieAnimationViewInitializers.swift
erichoracek: Would you be able to add a note explaining that this conformance doesn't conflict with conformances defined in other modules? As-is this would be a concern but seems like it's not an issue

erichoracek (approved):
@calda Looks great, just a few suggestions/comments/questions

Sources/Public/Animation/LottieAnimationViewInitializers.swift
calda: Since `Sendable` is a marker protocol without any runtime component, this doesn't cause conflicts with other modules (I checked that multiple modules in the same graph can define this extension). Will mention this in the comment.

Sources/Public/Animation/LottieAnimationViewInitializers.swift
calda: Agreed, this doesn't seem like a good idea. Will remove it.

Sources/Public/Animation/LottieAnimationView.swift
calda: Good idea

calda: Oh, we don't have access to the previous animation here -- this is a callback for the `animationLoaded` closure itself. The docs say:

> A closure that is called when `self.animation` is loaded. When setting this closure, it is called immediately if `self.animation` is non-nil.

This is a bit non-standard but is required for this pattern to work:

```swift
let animationView: LottieAnimationView

if loadDotLottieFile {
  // Loads the .lottie file asynchronously
  animationView = LottieAnimationView(dotLottieName: "animation")
} else {
  // Loads the .json file synchronously
  animationView = LottieAnimationView(name: "animation")
}

animationView.animationLoaded = { animationView, animation in
  // If using a .lottie file, this is called once the file finishes loading.
  // If using a .json file, this is called immediately (since the animation is loaded synchronously).
  animationView.play()
}
```

PR #1824, opened 2022-11-29
Use Git Large File Storage for snapshot images

calda:
This PR adopts [Git Large File Storage](https://git-lfs.github.com/) for our snapshot images. This should reduce the default size of the repo by ~130MB (the size of `Tests/__Snapshots__`) and will prevent the git history from growing arbitrarily large.


calda:
After merging this PR I'll rewrite the git history of `master` to use LFS, so a typical checkout done by a package manager will no longer need to download any of these images or their existing git history. 

To prevent breaking any existing projects referencing specific commits on master (e.g. either though using explicit commit shas in their `Package.swift` / `Podfile` / `Cartfile`, or as recorded in `Package.resolved` / `Podfile.lock` / `Cartfile.resolved`), I'll publish a tag of the current `master` branch. This makes sure that existing commits can still be downloaded even if they're no longer present on `master`. I confirmed that all three package managers support this type of change.

calda:
ptal @erichoracek

erichoracek (approved):
Thanks for working on this, glad we can take this approach

erichoracek:
@calda One thought—do we need a guide for contributors to pull later changes after force pushing to `master`?

PR #1843, opened 2022-12-08
Add script for building .xcframework, instructions to README

calda:
This PR adds a script for building `Lottie.xframework`, which we'll be attaching to releases from now on.

I built a `Lottie.xcframework.zip` using `bundle exec rake build:xcframework` and attached it to the release for 4.0.1: https://github.com/airbnb/lottie-ios/releases/tag/4.0.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,16 @@ You can pull the [Lottie Github Repo](https://github.com/airbnb/lottie-ios/) and
 
 ### Swift Package Manager
 
-To install Lottie using [Swift Package Manager](https://github.com/apple/swift-package-manager)  you can follow the [tutorial published by Apple](https://developer.apple.com/documentation/xcode/adding_package_dependencies_to_your_app) using the URL for the Lottie repo with the current version:
+To install Lottie using [Swift Package Manager](https://github.com/apple/swift-package-manager) you can follow the [tutorial published by Apple](https://developer.apple.com/documentation/xcode/adding_package_dependencies_to_your_app) using the URL for the Lottie repo with the current version:
 
-1. In Xcode, select “File” → “Swift Packages” → “Add Package Dependency”
+1. In Xcode, select “File” → “Add Packages...”
 1. Enter https://github.com/airbnb/lottie-ios.git
+
+or you can add the following dependency to your `Package.swift`:
+
+```swift
+.package(url: "https://github.com/airbnb/lottie-ios.git", from: "4.0.0")
+```
 
 ### CocoaPods
 Add the pod to your Podfile:
@@ -52,6 +58,7 @@ After installing the cocoapod into your project import Lottie with
 ```swift
 import Lottie
 ```
+
 ### Carthage
 Add Lottie to your Cartfile:
 ```
@@ -63,6 +70,27 @@ And then run:
 carthage update
 ```
 In your application targets “General” tab under the “Linked Frameworks and Libraries” section, drag and drop lottie-ios.framework from the Carthage/Build/iOS directory that `carthage update` produced.
+
+### Binary dependency (`xcframework`)
+
+Lottie is also available as a `xcframework` binary dependency. Lottie's git repository is somewhat large (300+ MB), so using an `xcframework` (~10MB) reduces the amount of data that needs to be downloaded.
+
+Since Lottie 4.0.1, a `Lottie.xcframework.zip` archive is attached to each release. You can find the latest release [here](https://github.com/airbnb/lottie-ios/releases/latest). 
+
+`xcframework`s can be installed manually in your Xcode project. If using Swift Package Manager, you can add the following dependency to your `Package.swift`:
+
+```swift
+// You can either use a remote binary dependency where SPM downloads the xcframework from GitHub
+.binaryTarget(
+  name: "Lottie",
+  url: "https://github.com/airbnb/lottie-ios/releases/download/4.0.1/Lottie.xcframework.zip",
+  checksum: "b6d8b0b81975d91965b8bb00cffb0eae4b3d94538b6950a90bc1366afd5d4239")
+
+// Or you can download the xcframework locally and include it in your project's git repo:
+.binaryTarget(
+  name: "Lottie",
+  path: "path/to/Lottie.xcframework")
+```
 
 ### Data collection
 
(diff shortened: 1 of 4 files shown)


Rakefile
```diff
@@ -43,6 +43,28 @@ namespace :build do
...
+    xcodebuild(
+      '-create-xcframework ' +
+      '-framework .build/archives/Lottie_iOS.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-framework .build/archives/Lottie_iOS_Simulator.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-framework .build/archives/Lottie_macOS.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-framework .build/archives/Lottie_tvOS.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-framework .build/archives/Lottie_tvOS_Simulator.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-output .build/archives/Lottie.xcframework')
+    Dir.chdir('.build/archives') do
+      sh 'zip -r Lottie.xcframework.zip Lottie.xcframework'
```
erichoracek, line 64: Assuming that this will result in an artifact being uploaded to the GitHub action that can be attached to a release? Might be worth having a comment to that effect, or if not it would be nice to use the `.xcframework` from a CI run rather than one that was built locally

README.md
```diff
@@ -64,6 +71,27 @@ carthage update
 ```
 In your application targets “General” tab under the “Linked Frameworks and Libraries” section, drag and drop lottie-ios.framework from the Carthage/Build/iOS directory that `carthage update` produced.
 
+### Binary dependency (`xcframework`)
+
+Lottie is also available as a `xcframework` binary dependency. Lottie's git repository is somewhat large (300+ MB), so using an `xcframework` (~10MB) reduces the amount of data that needs to be downloaded.
+
+Since Lottie 4.0.1, a `Lottie.xcframework.zip` archive is attached to each release. You can find the latest release [here](https://github.com/airbnb/lottie-ios/releases/latest).
```
erichoracek, line 78: Where should folks look to find the checksum?

Rakefile
```diff
@@ -43,6 +43,28 @@ namespace :build do
       xcodebuild('build -scheme "Example (tvOS)" -destination "platform=tvOS Simulator,name=Apple TV" -workspace Lottie.xcworkspace')
     end
   end
+
+  desc 'Builds an xcframework for all supported platforms'
+  task :xcframework do
+    sh 'rm -rf .build/archives'
+    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (iOS)" -destination generic/platform=iOS -archivePath ".build/archives/Lottie_iOS" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES')
```
erichoracek, line 50: Maybe we could factor out the `SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES` into a constant since they're the same on every xcodebuild invocation?

Rakefile
```diff
@@ -43,6 +43,28 @@ namespace :build do
       xcodebuild('build -scheme "Example (tvOS)" -destination "platform=tvOS Simulator,name=Apple TV" -workspace Lottie.xcworkspace')
     end
   end
+
+  desc 'Builds an xcframework for all supported platforms'
+  task :xcframework do
+    sh 'rm -rf .build/archives'
```
erichoracek, line 49: Maybe we could move `.build/archives/` into a constant as well?

Rakefile
```diff
@@ -43,6 +43,28 @@ namespace :build do
...
+    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (tvOS)" -destination generic/platform=tvOS -archivePath ".build/archives/Lottie_tvOS" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES')
+    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (tvOS)" -destination "generic/platform=tvOS Simulator" -archivePath ".build/archives/Lottie_tvOS_Simulator" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES')
+    xcodebuild(
+      '-create-xcframework ' +
+      '-framework .build/archives/Lottie_iOS.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-framework .build/archives/Lottie_iOS_Simulator.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-framework .build/archives/Lottie_macOS.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-framework .build/archives/Lottie_tvOS.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-framework .build/archives/Lottie_tvOS_Simulator.xcarchive/Products/Library/Frameworks/Lottie.framework ' + 
+      '-output .build/archives/Lottie.xcframework')
```
erichoracek, line 62: Suggest using `join(" ")` on an array of strings so we don't need the trailing spaces on these

erichoracek (approved):
@calda Looks good, just a few questions!

Rakefile
calda: I was planning on uploading the xcframework to the release manually for now

> if not it would be nice to use the .xcframework from a CI run rather than one that was built locally

Should we prefer publishing artifacts built by CI rather than ones built locally? Are there any differences you have in mind?

README.md
calda: Not sure the best way to handle this. I image you'd have to download the artifact and run a command like `swift package compute-checksum Lottie.xcframework.zip`. I can add a note about this to the readme.

Rakefile
erichoracek: The ones published by CI are generally safer (e.g. against infected developer machines) and more easily reproducible for contributors

README.md
erichoracek: Maybe we could include it somewhere in the release notes?

Rakefile
calda: Good idea, I added a CI step that builds and uploads the `Lottie.xcframework.zip` file

README.md
calda: Good idea, we'll include the checksum in the release notes moving forward

Rakefile
calda: Good idea

PR #1848, opened 2022-12-12
Updating currentFrame on MainThreadAnimationLayer to reflect the actual current value

goergisn:
This PR adds updates the currentFrame property on  MainThreadAnimationLayer to reflect the actual current value.

## Background
The `currentFrame` value never represented the actual value (verified on Lottie 3.1.5) of the animation progress but was mainly a way to update the animation. The introduction of `CoreAnimationLayer` added the expectation that `currentProgress` on the `AnimationView` actually returns the elapsed animation progress. This PR introduces that behavior for the MainThreadAnimationLayer.

### DIFF
--- a/Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift
+++ b/Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift
@@ -126,6 +126,7 @@ final class MainThreadAnimationLayer: CALayer, RootAnimationLayer {
   public override func display() {
     guard Thread.isMainThread else { return }
     var newFrame: CGFloat
+    defer { currentFrame = newFrame }
     if
       let animationKeys = animationKeys(),
       !animationKeys.isEmpty


calda:
> The currentFrame value never represented the actual value (verified on Lottie 3.1.5) of the animation progress but was mainly a way to update the animation

If this wasn't the case before, what was the value of `currentFrame` when the animation was playing?

calda:
This may be an intentional difference between `currentProgress` and `realtimeAnimationProgress`. Does `realtimeAnimationProgress` work as you expect?

I'm slightly worried about changing the behavior of the existing property, since it would be a breaking change for consumers. If `realtimeAnimationProgress` has the behavior you expect then we may not want to make any changes here.

goergisn:
`realtimeAnimationProgress` is actually exactly what I need, thanks for pointing it out. Closing the PR and looking into how to make the documentation clearer 👌

calda:
Awesome!

PR #1857, opened 2022-12-14
Add support for display P3 color space

alexisbronchart:
This PR adds support for display P3 color space. The changes are opt-in only, anyone willing to use the P3 color space would have to explicitly enable through a new flag in `LottieConfiguration` like so:
```
LottieConfiguration.shared.useDisplayP3 = true
```

### DIFF
--- a/Sources/Public/LottieConfiguration.swift
+++ b/Sources/Public/LottieConfiguration.swift
@@ -1,6 +1,8 @@
 // Created by Cal Stephens on 12/13/21.
 // Copyright © 2021 Airbnb Inc. All rights reserved.
 
+import QuartzCore
+
 // MARK: - LottieConfiguration
 
 /// Global configuration options for Lottie animations
@@ -10,10 +12,12 @@ public struct LottieConfiguration: Hashable {
 
   public init(
     renderingEngine: RenderingEngineOption = .automatic,
-    decodingStrategy: DecodingStrategy = .dictionaryBased)
+    decodingStrategy: DecodingStrategy = .dictionaryBased,
+    colorSpace: CGColorSpace = CGColorSpaceCreateDeviceRGB())
   {
     self.renderingEngine = renderingEngine
     self.decodingStrategy = decodingStrategy
+    self.colorSpace = colorSpace
   }
 
   // MARK: Public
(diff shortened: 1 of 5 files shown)


Sources/Public/LottieConfiguration.swift
```diff
@@ -32,6 +36,19 @@ public struct LottieConfiguration: Hashable {
...
   public var decodingStrategy: DecodingStrategy
 
+  /// Use the display P3 color space for rendering instead of the default RGB
+  ///  - Defaults to `false`
+  public var useDisplayP3: Bool
+}
+
+extension LottieConfiguration {
+    var renderingColorSpace: CGColorSpace {
+        if let colorspace = CGColorSpace(name: CGColorSpace.displayP3), self.useDisplayP3 {
```
marcelofabri, line 46: a small optimization would be checking `useDisplayP3` first so the colorspace is not created if `useDisplayP3` is `false` (the default behavior)

alexisbronchart: Updated!

Sources/Public/LottieConfiguration.swift
```diff
@@ -32,6 +36,22 @@ public struct LottieConfiguration: Hashable {
   /// The decoding implementation to use when parsing an animation JSON file
   public var decodingStrategy: DecodingStrategy
 
+  /// Use the display P3 color space for rendering instead of the default RGB
+  ///  - Defaults to `false`
+  public var useDisplayP3: Bool
```
calda, line 41: Rather than having a boolean for toggling Display P3, could we have a `animationColorSpace: CGColorSpace` property? That API would be more flexible since it lets consumers use any color space. We could include instructions / sample code in the doc comment on how to set the color space to Display P3.

Example/iOS/AppDelegate.swift
```diff
@@ -20,6 +20,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     Configuration.applyCurrentConfiguration()
 
+    LottieConfiguration.shared.useDisplayP3 = true
```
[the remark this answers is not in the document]
calda: Yeah, please remove this before merging. If we want to support testing multiple color spaces in the Example app we could add an entry to the configuration menu that lets you switch between RGB and Display P3.

Sources/Public/LottieConfiguration.swift
alexisbronchart: Done. I can follow up with a PR on the doc repo but basically this can be used like so:
```
if let displayP3 = CGColorSpace(name: CGColorSpace.displayP3) {
    LottieConfiguration.shared.colorSpace = displayP3
}
```

calda: Thanks!

calda (approved):
Thank you! Nice and simple :)

PR #1858, opened 2022-12-15
Fix heap buffer overflow in CGColor+RGB extensions

anlaital:
Running Lottie with Address Sanitizer (ASan) enabled reveals a heap buffer overflow when `CGColor`s are constructed. This is regression from https://github.com/airbnb/lottie-ios/pull/1801.

This happens because the provided `components` array does not include enough elements as per [init's documentation](https://developer.apple.com/documentation/coregraphics/cgcolor/1455927-init):
> An array of intensity values describing the color. The array should contain n+1 values that correspond to the n color components in the specified color space, followed by the alpha component. Each component value should be in the range appropriate for the color space. Values outside this range will be clamped to the nearest correct value.

This causes the internal CoreGraphics `create_color` function to access array index out of bounds.

Two things that should be noted:
- I could not verify the fix to `gray` function works in the context of the framework since I could not find an animation that uses it in the project. It might also be worth it to add unit tests for them that could be run with ASan enabled? I did craft some code by hand and verified that the current implementation causes heap buffer overflow and that this PR fixes it.
- I did not understand the calls to `copy` so I removed them, but I can restore them in case they are necessary. Should also then add comments describing why it's done as it's non-obvious to me at least. Based on the original PR it looks like they were added due to iOS 11 minimum deployment target, but I don't think that's necessary since the used API is available since iOS 2.0.

### DIFF
--- a/Sources/Private/Utility/Extensions/CGColor+RGB.swift
+++ b/Sources/Private/Utility/Extensions/CGColor+RGB.swift
@@ -6,22 +6,20 @@ import QuartzCore
 extension CGColor {
   /// Initializes a `CGColor` using the given `RGB` values
   static func rgb(_ red: CGFloat, _ green: CGFloat, _ blue: CGFloat) -> CGColor {
-    CGColor(
-      colorSpace: CGColorSpaceCreateDeviceRGB(),
-      components: [red, green, blue])!
-      .copy(alpha: 1)!
+    rgba(red, green, blue, 1.0)
   }
 
   /// Initializes a `CGColor` using the given grayscale value
   static func gray(_ gray: CGFloat) -> CGColor {
     CGColor(
       colorSpace: CGColorSpaceCreateDeviceGray(),
-      components: [gray])!
-      .copy(alpha: 1)!
+      components: [gray, 1.0])!
   }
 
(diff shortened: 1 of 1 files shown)


marcusziade (approved):
Thanks for the explanation 👍 

Agreed about documenting `.copy`

chaseklingelzen:
Hi @anlaital we are experiencing Stack buffer overflow when Address Sanitizer is checked. Is this a potentially related issue? Happy to provide any more context that I can. 

Screen Shot 2022-12-15 at 11 12 34 AM

anlaital:
@chaseklingelzen this fixes exactly that issue.

chaseklingelzen:
@anlaital thanks for the quick response! We are trying to determine if this is causing app crashes on production. Is this known to actually cause crashes?

Kondamon (approved):
Looks fine! 👍

calda (approved):
Thanks for the fix!

Kondamon:
@calda When are you going to release (4.0.2) this fix?

calda:
We released 4.1.0 today https://github.com/airbnb/lottie-ios/releases/tag/4.1.0

PR #1872, opened 2022-12-20
Fix issue where ShapeTransform could incorrectly affect stroke width

### LINKED ISSUE #1855: Using rendering engine .coreAnimation showing animation differently than before
## Which Version of Lottie are you using

Lottie 4.0.1

## Expected Behavior 

- .coreAnimation RenderingEngine show same as .mainThread engine and [web-player](https://lottiefiles.com/web-player) playing

## Actual Behavior

- different with web-player and .mainThread engine

## Animation JSON

- https://static.toss.im/lotties/general/loan-load.json

calda:
This PR fixes an issue where `ShapeTransform`s could incorrectly affect stroke width. Fixes #1855.

In this case, the shape layer is structured like this:

 - Stroke (10px width)
 - Group
   - Shape Transform (3x scale)
   - Circle

This was being transformed into a single `CAShapeLayer` with a 3x scale and a 10px stroke. This caused the stroke to be rendered incorrectly with a width of 30px. 

Since we can't really change the structure of how the stroke is rendered, the best fix is to continue rendering a single `CAShapeLayer` but with a 3x scale and a 3.333px stroke (which is 10px wide when actually rendered).

| Before (Core Animation engine) | After (Core Animation engine) | Main Thread engine |
| ---- | ----- | ----- |
| Simulator Screen Shot - iPhone 14 Pro - 2022-12-20 at 10 16 49 | Simulator Screen Shot - iPhone 14 Pro - 2022-12-20 at 11 28 49 | Simulator Screen Shot - iPhone 14 Pro - 2022-12-20 at 11 29 06 |


erichoracek (approved):
Nice find and fix!

PR #1873, opened 2022-12-20
Revert workaround for Xcode 14.0 SPM binary dependency issue (#1705)

calda:
This PR reverts #1705, which was a workaround for an issue in Xcode 14.0 where SPM binary dependencies would cause Xcode to spin indefinitely at 100% CPU load. This has been fixed in Xcode 14.1+, so we don't need to use the workaround anymore.

This also fixes an issue where `Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved` was often unnecessarily marked as modified by git, since the Airbnb formatter tool was being added locally when running the formatter but not present on master.

### DIFF
--- a/Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Lottie.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/airbnb/swift",
         "state": {
           "branch": null,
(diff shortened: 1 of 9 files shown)


erichoracek (approved):
Glad this was fixed! Can we merge the same Fix in Epoxy too?

calda:
Yep, here you go: https://github.com/airbnb/epoxy-ios/pull/132

PR #1874, opened 2022-12-20
Fix issue where some top-level shape items would be silently ignored.

### LINKED ISSUE #1853: Core Animation Engine regression: Can not handle top items next to a group

Lottie 4.0.1

## Expected Behavior 
When using the Main Thread Engine the initial (hollowed) star is displayed.

https://user-images.githubusercontent.com/36905/207494595-17bff305-6141-4d1a-9435-784650ee2955.mp4

## Actual Behavior
When using Core Animation Engine the initial star is not displayed.

https://user-images.githubusercontent.com/36905/207494643-7e0ae616-50cf-4da2-8d12-584c3227c3fa.mp4

## Animation JSON
[STAR_RATING_BUTTON_ANIMATION.json.zip](https://github.com/airbnb/lottie-ios/files/10223944/STAR_RATING_BUTTON_ANIMATION.json.zip)

## Discussion
The layer called "STAR_01" has top level items and a group.
The top level items next to "Group 1" can not be handled in `ShapeLayer.makeGroupLayers(...)`.
Additionally a fill (Fill 1) should be applied to the group ("Group 1").

Structure:
- Shape name: Path 2
- Fill name: Fill 2
- Group name: Group 1
   * Shape name: Path 1
   * ShapeTransform name: Transform
- Fill name: Fill 1

I attempted a solution by creating a group for the top level items, but I'm not sure how to handle the case of applying "Fill 1" to the correct group (Group 1).

```diff
    // If this shape doesn't have any groups but just has top-level shape items,
    // we can create a placeholder group with those items. (Otherwise the shape items
    // would be silently ignored, since we expect all shape layers to have a top-level group).
    if groupItems.isEmpty, parentGroup == nil {
      groupItems = [Group(items: otherItems, name: "")]
      otherItems = []
    }

+    if parentGroup == nil, !groupItems.isEmpty, !otherItems.isEmpty {
+        groupItems.append(Group(items: otherItems, name: "Other Items"))
+        otherItems = []
+    }

    // `ShapeItem`s either draw a path, or modify how a path is rendered.
```

calda:
This PR fixes an issue where some top-level shape items were being silently ignored. Fixes #1853.

In this example the shape layer is structured like this:

 - Path 1
 - Fill (Black)
 - Group
   - Path 2
 - Fill (White)

We perviously didn't handle cases where there were both top-level shape-drawing item _and_ a top-level group. This was causing us to only render a single shape layer with Path 2 and no fill color (meaning we effectively rendered neither of the two paths).

Instead we need to render two shape layers -- Path 1 with a black fill, and Path 2 with a white fill. We can do this by computing the shape render groups, and having child groups inherit any items that would be silently ignored due to not being associated with a path. (e.g. in this case `[Path 1, Fill (Black), Fill (White)]` is a single shape render group `[Path 1, Fill (Black)]`, leaving the `Fill (White)` item unused).

This transforms the shape layer structure to:

- Path 1
- Fill (Black)
- Group
  - Path 2
  - Fill (White, inherited from parent)

which is something we can handle rendering without any issues.

| Before | After |
| --- | ---- |
| Simulator Screen Shot - iPhone 14 Pro - 2022-12-20 at 14 38 26 | Simulator Screen Shot - iPhone 14 Pro - 2022-12-20 at 14 39 00 |


Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
```diff
@@ -210,25 +210,21 @@ extension CALayer {
...
-      }
-    } else {
-      inheritedItemsForChildGroups = []
-    }
+    // Any child items that wouldn't be included in a valid shape render group
+    // need to be applied to child groups (otherwise they'd be silently ignored).
+    let inheritedItemsForChildGroups = otherItems
+      .map { ShapeItemLayer.Item(item: $0, groupPath: parentGroupPath) }
+      .shapeRenderGroups
+      .unusedItems
```
calda, line 227: Both of these changes are a superset of the previous logic. Always nice when we can just further-generalize existing logic rather than adding more special cases.

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
```diff
@@ -414,5 +412,16 @@ extension Array where Element == ShapeItemLayer.Item {
...
+
+    var unusedItems = [ShapeItemLayer.Item]()
+    for renderGroup in renderGroups {
+      // All valid render groups must have a path, otherwise the items wouldn't be rendered
+      if renderGroup.pathItems.isEmpty {
+        unusedItems.append(contentsOf: renderGroup.otherItems)
+      }
+    }
+
+    renderGroups.removeAll(where: { $0.pathItems.isEmpty })
```
erichoracek, line 424: Rather than doing two passes over `renderGroups` could we iterate its indexes in reverse and remove items by index as we find empty ones?

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
```diff
@@ -354,15 +350,17 @@ extension Collection {
...
   /// Shape items that modify the appearance of the shapes rendered by this group
   var otherItems: [ShapeItemLayer.Item] = []
 }
 
 extension Array where Element == ShapeItemLayer.Item {
-  /// Splits this list of `ShapeItem`s into groups that should be rendered together as individual units
-  var shapeRenderGroups: [ShapeRenderGroup] {
+  /// Splits this list of `ShapeItem`s into groups that should be rendered together as individual units,
+  /// plus the remaining items that were not included in any group.
+  var shapeRenderGroups: (groups: [ShapeRenderGroup], unusedItems: [ShapeItemLayer.Item]) {
```
erichoracek, line 363: Maybe something like this to make it more clear than `groups.groups`?
```suggestion
  var shapeRenderGroups: (used: [ShapeRenderGroup], unused: [ShapeItemLayer.Item]) {
```

erichoracek (approved):
Looks good to me, just a few small suggestions

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
calda: Good idea

Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
calda: Good idea

Issue #1091, opened 2019-11-29
Unable to set the frame of CompatibleAnimationView


whisterlee:
Hello! 

## Check these before submitting:

- [x] The issue doesn't involve an [Unsupported Feature](https://github.com/airbnb/lottie-ios/blob/master/README.md#supported-after-effects-features)

- [x] This issue isn't related to another open issue

## This issue is a:

- [x] Non-Crashing Bug (Visual or otherwise)
- [ ] Crashing Bug
- [ ]  Feature Request
- [ ] Regression (Something that once worked, but doesn't work anymore)

## Which Version of Lottie are you using?

Lottie 3.0

## What Platform are you on? 

- [ ] MacOS

- [x] iOS

## What Language are you in?

- [ ] Swift
- [x] Objective-C

## Expected Behavior 

CompatibleAnimationView should be the same size as the parent view

## Actual Behavior

CompatibleAnimationView is not the same size as the parent view

![image](https://user-images.githubusercontent.com/17923550/69847605-8e5a9080-12b2-11ea-8bdc-d32d0718fe5f.png)

## Code Example

- (void)viewDidLoad
{
    [super viewDidLoad];
    
    CGFloat timeLineViewW = 276.f;
    CGFloat timeLineViewH = 96.f;
    CGFloat timeLineViewX = (self.view.bounds.size.width - timeLineViewW) / 2.f;
    CGFloat timeLineViewY = (self.view.bounds.size.width - timeLineViewH) / 2.f;
    UIView *timeLineView = [[UIView alloc] initWithFrame:CGRectMake(timeLineViewX, timeLineViewY, timeLineViewW, timeLineViewH)];
    timeLineView.backgroundColor = [UIColor blueColor];
    [self.view addSubview:timeLineView];
    
    CompatibleAnimationView *animationView = [[CompatibleAnimationView alloc] initWithFrame:timeLineView.bounds];
    animationView.contentMode = UIViewContentModeScaleAspectFit;
    animationView.loopAnimationCount = -1;
    animationView.compatibleAnimation = [[CompatibleAnimation alloc] initWithName:@"timeline" bundle:[NSBundle mainBundle]];
    [animationView play];
    [timeLineView addSubview:animationView];
}

## Animation JSON

[timeline.json.zip](https://github.com/airbnb/lottie-ios/files/3903965/timeline.json.zip)

farfromrefug:
I am facing the same issue!

farfromrefug:
[...]

PR #1878, opened 2022-12-22
Fix frame setting issue for CompatibleAnimationView.

snow-google:
As discussed on the GitHub page in the past, CompatibleAnimationView does not correctly set up the constaints as to make the underlying AnimationView fill the entirity of the frame provided to CompatibleAnimationView. https://github.com/airbnb/lottie-ios/issues/1091

This fix is the same that was discussed in the issue thread and resolves the frame setting issue.


calda:
I don't think this change is backwards-compatible, it seems like it could easily be a breaking change for consumers who have code that relies on `translatesAutoresizingMaskIntoConstraints` defaulting to `false`. Could you share more details about your issue? Why does this fix the problem? Is there another solution we could use that would have less effect on existing consumers?

snow-google:
As it is, the compatible view is a wrapper on the underlying animation view. When the compatible view has its frame set, the underlying animation view does not get these updates. This is not a problem when using the aniamtion view directly which means that it is only ObjC compatible views that have this unexpected behavior. In essense, the abstraction and encapsulation that the compatible view is supposed to create are broken by the current behavior which requires compatible view users to investigate how to resolve the issue. It is possible that the workaround previously discussed on this forum makes it less likely for people to report this issue, but for my use case, overriding the value in our source code is incredibly challenging.

calda (approved):
Ok, thanks for the explanation. I think this change makes sense considering `translatesAutoresizingMaskIntoConstraints` also defaults to `true` for `LottieAnimationView` -- seems like we'd want both to have the same behavior.

I went ahead and also added a test case for `CompatibleAnimationView` to the snapshot test suite.

snow-google:
Thank you so much for the work to accept and verify this change! I appreciate the responsiveness of the team!

PR #1881, opened 2022-12-23
Add some missing utility to CompatibleAnimationView.

snow-google:
CompatibleAnimationView only has one way of initializing instances, with already initialized CompatibleAnimation instances. CompatibleAnimation only has one way of initializing instances, with a resource name and bundle. The Objective-C Lottie library had two ways of initializing views that are not available with today's APIs:
  - Initializing directly from an NSDictionary
  - Initializing from a remote URL resource

This change adds constructors for these use cases. Additionally, there was no way to configure the rendering engine to use from the compatibility classes. The initializers have been updated to allow for this configuration on a per-view basis.

Finally, a duration getter API was added along with some helpful comments around naming implicit intro naming conventions that make calling these methods from Objective-C code quite tricky today.

### DIFF
--- a/Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
+++ b/Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
@@ -40,16 +40,117 @@ public final class CompatibleAnimation: NSObject {
   private let bundle: Bundle
 }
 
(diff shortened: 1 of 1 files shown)


calda (approved):
Thanks!

snow-google:
Thank you so much for reviewing, approving, and polishing this change!

Issue #1880, opened 2022-12-23
Main Thread renderer shape layer issue


snow-google:
The main thread renderer does not render shape layers with certain combinations of contentMode and frame height/width. The easiest repro is when the content mode is set to aspect fit. In this scenario, having an animationView with a larger height than width causes the animation to not render while having a larger width than height seems to work fine. The CoreAnimation renderer has a different issue with this same animation. Here is the issue in case it is helpful: https://github.com/airbnb/lottie-ios/pull/1878. There are similar issues when using contentMode aspect fill.

## Which Version of Lottie are you using?

Lottie 3.5.0

## Expected Behavior 

When width of AnimationView exceeds height (correct):
![width bigger](https://user-images.githubusercontent.com/121244810/209380577-98018e2b-dfea-4588-92c1-0d658cfb72fd.png)

## Actual Behavior

When height of AnimationView exceeds width (incorrect):
![height bigger](https://user-images.githubusercontent.com/121244810/209380721-36b88fd3-9ff8-4e3d-b887-d2c3879af3d2.png)

## Animation JSON

[...]

PR #1884, opened 2022-12-29
Change veryLargeRect size to fix issue with masks (#1880)

nevil:
When using scaleToFill / scaleAspectFit / scaleAspectFill mode with an animation that has a subtract mask, the animation could sometimes fail to render. `veryLargeRect` is used as the size of the "bounds" rect from which the path is subtracted.

**Alternative considered**
I digged through the old code and found that the old Objective-C implementation actually passed in the bounds size.
I guess that could be an alternative solution if needed.

```objective-c
- (void)updateForFrame:(NSNumber *)frame withViewBounds:(CGRect)viewBounds {
   if ([self hasUpdateForFrame:frame]) {
     LOTBezierPath *path = [_pathInterpolator pathForFrame:frame cacheLengths:NO];

     if (self.maskNode.maskMode == LOTMaskModeSubtract) {
       CGMutablePathRef pathRef = CGPathCreateMutable();
       CGPathAddRect(pathRef, NULL, viewBounds);
       CGPathAddPath(pathRef, NULL, path.CGPath);
       self.path = pathRef;
       self.fillRule = @"even-odd";
     } else {
       self.path = path.CGPath;
     }

     self.opacity = [_opacityInterpolator floatValueForFrame:frame];
   }
 }
```
https://github.com/airbnb/lottie-ios/commit/502b18ff#diff-233d9c9fcd51dcb99d9884451e850073e2712d14d3c05e0e91bae1c551abe91eR42

Video capture before/after the change

iPhone14 Pro Simulator
| Before | After |
| ------ | ------ |
| <video src="https://user-images.githubusercontent.com/36905/209914152-a2eb6eee-c7ce-4dc6-bd19-5efb9a7fe115.mp4" /> | <video src="https://user-images.githubusercontent.com/36905/209914166-f734f5b6-8bf6-478a-8d77-56ba8fe3ee64.mp4" /> |

### DIFF
--- a/Sources/Private/MainThread/LayerContainers/CompLayers/MaskContainerLayer.swift
+++ b/Sources/Private/MainThread/LayerContainers/CompLayers/MaskContainerLayer.swift
@@ -84,10 +84,10 @@ final class MaskContainerLayer: CALayer {
 extension CGRect {
   static var veryLargeRect: CGRect {
     CGRect(
-      x: -100_000_000,
-      y: -100_000_000,
-      width: 200_000_000,
-      height: 200_000_000)
+      x: -10_000_000,
+      y: -10_000_000,
+      width: 20_000_000,
+      height: 20_000_000)
   }
 }
 


Sources/Private/MainThread/LayerContainers/CompLayers/MaskContainerLayer.swift
```diff
@@ -84,10 +84,10 @@ final class MaskContainerLayer: CALayer {
 extension CGRect {
   static var veryLargeRect: CGRect {
     CGRect(
-      x: -100_000_000,
-      y: -100_000_000,
-      width: 200_000_000,
-      height: 200_000_000)
+      x: -10_000_000,
```
nevil, line 87: How large is a large rect? 🤷‍♂️

calda (approved):
Thanks for the fix! Nice find.

nevil:
Seems like the same test is failing each time, but it shouldn't really be affected by the change to `veryLargeRect` as it isn't used in that test. 🤔 

```
✖ testParsing_simpleAnimation, XCTAssertEqualWithAccuracy failed: ("1.3376318674468077") is not equal to ("2.0") +/- ("0.65")
```

calda:
The performance tests are flaky sometimes -- we just need to increase the threshold

PR #1888, opened 2022-12-29
Fix issue where Fill following other Fill would be silently ignored

### LINKED ISSUE #1887: Gradient fill are not showing in version 4.0.1
Hi. I'm having an issue.
when I use Lottie 4.0.1 version, cannot render gradient fill 
Any help would be greatly appreciated.

Lottie 4.0.1

## Animation JSON
[lottie Test.json.zip](https://github.com/airbnb/lottie-ios/files/10320717/lottie.Test.json.zip)

calda:
This PR fixes an issue where a `Fill` item following another `Fill` item would be silently ignored, because it would create a render group with no path-drawing items. Fixes #1887. 

For example, this animation has [Circle, Gradient Fill (White), Gradient Fill (Black)]. Previous we would split this into [Circle, Gradient Fill (White)] and [Gradient Fill (Black)], leaving the black fill unused. Now we split this into [Circle, Gradient Fill (White)] and [Circle, Gradient Fill (Black)].

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 14 Pro Max - 2022-12-29 at 14 11 57 | Simulator Screen Shot - iPhone 14 Pro Max - 2022-12-29 at 14 12 26 |


Sources/Private/CoreAnimation/Layers/ShapeLayer.swift
```diff
@@ -118,6 +118,9 @@ final class GroupLayer: BaseAnimationLayer {
   ///    structure of the `ShapeLayer` data model, we track that info here.
   private let groupPath: [String]
 
+  /// Child group items contained in this group. Correspond to a child `GroupLayer`
```
erichoracek, line 121:
```suggestion
  /// Child group items contained in this group. Corresponds to a child `GroupLayer`
```

erichoracek (approved):
LGTM!

PR #1894, opened 2023-01-04
Fix issue where negative `scale.x` values didn't render correctly on iOS 14

### LINKED ISSUE #1882: Old iOS versions have rendering issues [iOS 13]

Lottie 4.0.1 commit 0ceb3ece

## Expected Behavior 
CleanShot 2022-12-26 at 14 31 51@2x

## Actual Behavior
wrong

This issue happens for iOS 13 devices when I use `LottieConfiguration(renderingEngine: .automatic)`.
When I use `LottieConfiguration(renderingEngine: .mainThread)` everything goes back to normal.
Animation works fine on iOS16 with both rendering engines.

calda:
This PR fixes an issue where negative `scale.x` values didn't render correctly on iOS 14 when using `RenderingEngineOption.automatic`. Fixes #1882.

I imagine there's a way to support this properly on iOS 14 and earlier, but I think it's better to just throw a compatibility error in this case:
 1. The existing code path for this doesn't work correctly in snapshot tests so any changes we make are risky and have the potential to cause invisible regressions
 2. It seems good to avoid having two complex code paths with substantially different behavior, especially considering we don't include CI tests for older OS versions.

| Before | After | 
| ---- | ---- |
| Simulator Screen Shot - iPhone 6s (iOS 14) - 2023-01-04 at 11 34 18 | Simulator Screen Shot - iPhone 6s (iOS 14) - 2023-01-04 at 11 38 38 |


erichoracek (approved):

Sources/Private/CoreAnimation/Animations/TransformAnimations.swift
```diff
@@ -146,17 +146,42 @@ extension CALayer {
...
+    //    log a compatibility error on devices running older OSs.
     if TestHelpers.snapshotTestsAreRunning {
-      if transformModel.scale.keyframes.contains(where: { $0.value.x < 0 }) {
+      if hasNegativeXScaleValues {
         context.logger.warn("""
           Negative `scale.x` values are not displayed correctly in snapshot tests
           """)
       }
     } else {
+      if !osSupportsNegativeScaleValues, hasNegativeXScaleValues {
```
erichoracek, line 179: I'm a bit confused how this fixes the issue, isn't this just logging an extra warning?

calda: The `logCompatibilityIssue` and `compatibilityAssert` methods cause `RenderingEngineOption.automatic` to detect that the animation is unsupported and fall-back to using the main thread rendering engine.

erichoracek: Ah, got it!

PR #1896, opened 2023-01-05
Use Lottie swift package independent of Lottie Workspace and Example Xcode Project

adozenlines:
- Moves dependencies from Example and Lottie Workspace down to the Lottie swift package manifest file, allowing the swift package to be used independently of the Xcode sample workspaces. 
- Xcode Example project now uses the Lottie swift package locally.
- Excluded the snapshot images folder during a package checkout to reduce project bloat. This separates swift package test dependencies from the Xcode for production uses. If needed, a checkout of the project repo manually for development and testing purposes of the Lottie project.
- Restructured the Test folder to Test/LottieTest moving all resources under LottieTest.


calda:
Thanks for the suggestions @adozenlines. These are pretty significant changes to the project structure, which I'd want to discuss extensively before merging. Can you explain the motivation for these changes? (particularly the `Package.swift` changes -- I'm not sure why we'd want to make this change).

calda:
> Excluded the snapshot images folder during a package checkout to reduce project bloat. This separates swift package test dependencies from the Xcode for production uses. If needed, a checkout of the project repo manually for development and testing purposes of the Lottie project.

What do you mean "excluded during package checkout"? Since this content is part of the git repository it will have to be downloaded / checked-out no matter where the files live in the repo.

> Restructured the Test folder to Test/LottieTest moving all resources under LottieTest.

Is there a reason we should do this? I'd prefer not to unless there's a really good reason.

calda:
These changes also break out CI piplelines and actions configured via in the `Rakefile`.

adozenlines:
> Thanks for the suggestions @adozenlines. These are pretty significant changes to the project structure, which I'd want to discuss extensively before merging. Can you explain the motivation for these changes? (particularly the `Package.swift` changes -- I'm not sure why we'd want to make this change).

The general practice is to open the swift package and try to build it manually if you want to contribute to the framework itself. This change, as noted, moves the Epoxy & Difference dependencies into the Lottie package, where building & testing the package is independent of the Xcode projects Lottie & Example under the Workspace Lottie. The only change here is Packages.swift and the creation of a new folder, "LottieTest" which is now the test target for unit testing the package.

adozenlines:
The purpose of the request is to remove all of the images during a package check under __SnapShot__ or at least __SnapShotTest__ the create unwanted project bloat as they are never needed during a project build.

calda:
We use `Lottie.xcworkspace` when making and testing changes to Lottie. One reason is that we have to continue supporting Carthage, which builds via the `Lottie.xcworkspace` and not `Package.swift`. It's important to keep the `Lottie.xcworkspace` up-to-date or the Carthage build will break. Since managing Xcode projects is a bit more manual than managing `Package.swift` projects, I've found that the easiest workflow is to just always use the `xcworkspace`.

Also, using the `xcworkspace` lets us have internal test-only dependencies (like Epoxy and swift-snapshot-testing). If we add those to our `Package.swift` then they become dependencies for all consumers who use Lottie, which isn't ideal since the public  Lottie package doesn't even use those dependencies.

I'm going to close this PR -- feel free to [start a discussion](https://github.com/airbnb/lottie-ios/discussions) if you have questions or more suggestions about our project setup.

adozenlines:
> We use `Lottie.xcworkspace` when making and testing changes to Lottie. One reason is that we have to continue supporting Carthage, which builds via the `Lottie.xcworkspace` and not `Package.swift`. It's important to keep the `Lottie.xcworkspace` up-to-date or the Carthage build will break. Since managing Xcode projects is a bit more manual than managing `Package.swift` projects, I've found that the easiest workflow is to just always use the `xcworkspace`.
> 
> Also, using the `xcworkspace` lets us have internal test-only dependencies (like Epoxy and swift-snapshot-testing). If we add those to our `Package.swift` then they become dependencies for all consumers who use Lottie, which isn't ideal since the public Lottie package doesn't even use those dependencies.
> 
> I'm going to close this PR -- feel free to [start a discussion](https://github.com/airbnb/lottie-ios/discussions) if you have questions or more suggestions about our project setup.

Go ahead and close the PR I will open a issue to exclude "__SnapShotTest__" from the package manifest

calda:
If you're concerned about the checkout size of the project git repo (300MB+, which is admittedly quite large), there's a discussion about this [here](https://github.com/airbnb/lottie-ios/discussions/1807) where we're exploring solutions: https://github.com/airbnb/lottie-ios/discussions/1807

adozenlines:
> If you're concerned about the checkout size of the project git repo (300MB+, which is admittedly quite large), there's a discussion about this [here](https://github.com/airbnb/lottie-ios/discussions/1807) where we're exploring solutions: #1807

See the package manifest in this pull request it solves the issue.

calda:
>  I will open a issue to exclude "SnapShotTest" from the package manifest

I believe this is already the case -- our `Package.swift` manifest only references the `Sources` folder, which doesn't include our snapshot tests.

adozenlines:
> > I will open a issue to exclude "SnapShotTest" from the package manifest
> 
> I believe this is already the case -- our `Package.swift` manifest only references the `Sources` folder, which doesn't include our snapshot tests.

When the package is checkout everything is downloaded.

calda:
In what way does this fix the bloat? The snapshot files would still present on disk even if they aren't included in the generated Package.swift workspace. Does these files being present in the generated workspace negatively impact your flow somehow, other than just contributing to the download size of the repository / checkout?

adozenlines:
> In what way does this fix the bloat? The snapshot files would still present on disk even if they aren't included in the generated Package.swift workspace. Does these files being present in the generated workspace negatively impact your flow somehow, other than just contributing to the download size of the repository / checkout?

There will be no `__SnapShots__` on checkout visible in the file tree under Test.

PR #1892, opened 2023-01-03
Core Animation Engine - Set currentFrame when animation is finished

nevil:
For the Core Animation Engine the `currentFrame` is calculated based on the `animationProgress` placeholder animation.
The `animationProgress` value is never copied from the presentation layer to the model layer.

Because of that value of `currentFrame` is lost once the place holder animation has reached completion and was automatically removed.

This PR changes the placeholder animation to remain and instead be manually removed in the `animationDidStop(..)` delegate.
That way it is possible to persist the `currentFrame` value.

Attempts to solve issues:
#1877
#1865

| Before | After |
| ------- | ---- |
| <video src="https://user-images.githubusercontent.com/36905/210312095-e07c547f-b007-4b15-9fc5-5effb7241a16.mp4" /> | <video src="https://user-images.githubusercontent.com/36905/210312050-26b9b7a6-a7cc-4152-9c74-69689ee34870.mp4" /> |

*Notes*
#1877 is not fully fixed as there still is an issue when setting the switch to off from the start. The wrong frame is shown.

__Warning__
I'm pretty certain that there are cases that I have missed and that I have now broken. 🙇‍♂️

PR #1897, opened 2023-01-05
Fix issue where `LottieAnimationView.currentFrame` would be incorrect after playing animation with `LottieLoopMode.playOnce`

### LINKED ISSUE #1865: `LottieAnimationView`'s `play(toFrame:)` does not update `currentFrame`

Latest master, commit hash: 0ceb3ece6feffeb213e6cc6b9fcea0b3b8ad507f

When I call `play(toFrame: someFrame)` on a `LottieAnimationView` and read `animationView.currentFrame` after the animation has completed, I get back `0.0` instead of `someFrame`.

## Expected Behavior 
After `play(toFrame: someFrame)` completes, I would expect that `animationView.currentFrame == someFrame`

## Actual Behavior
After `play(toFrame: someFrame)` completes, `animationView.currentFrame == 0.0`.

I tested this with the Lottie example workspace (Lottie.xcworkspace) by adding

```
      animationView.play(toFrame: 86) { _ in
          print("Current frame = \(self.animationView.currentFrame)")
      }
``` 

at the bottom of AnimationPreviewViewController.viewDidLoad and commenting out the CADisplayLink related logic (updateAnimation, animationCallback) and opening `9squares_AlBoardman`. I also changed the loopMode of the animation view in this file to `playOnce`

### LINKED ISSUE #1877: AnimatedSwitch behaves differently with coreAnimationEngine vs mainThreadEngine
Latest commit 3b907922757b550484405d6d708089d1941a9c47

## Description
We are using an AnimatedSwitch with the default settings for the progress for each state.
We do not call `setProgressForState(...)`
```swift
  fileprivate var onStartProgress: CGFloat = 0
  fileprivate var onEndProgress: CGFloat = 1
  fileprivate var offStartProgress: CGFloat = 1
  fileprivate var offEndProgress: CGFloat = 0
```

When animating back from on to off state the animation does not behave as expected.
I believe the root cause is that `currentFrame` is 0 at this point for the coreAnimationEngine but not for the mainThreadEngine.

Because of that, in `addNewAnimationForContext(...)` the playFrom value will be forced to 0:
```swift
        switch loopMode {
        // When playing exactly once (and not looping), we can just set the
        // `playFrom` time to be the `currentFrame`. Since the animation duration
        // is based on `playFrom` and `playTo`, this automatically truncates the
        // duration (so the animation stops playing at `playFrom`).
        case .playOnce:
          animationContext.playFrom = currentFrame
```

With the mainThreadEngine the currentFrame will be set to the expected value in didAnimationStop but this is not the case for the coreAnimationEngine.

```swift
    if let animationLayer = animationLayer, let key = animationKey {
      animationLayer.removeAnimation(forKey: key)
      if flag {
        animationLayer.currentFrame = (anim as! CABasicAnimation).toValue as! CGFloat
      }
    }
```

Temporary workaround, not sure if it is accurate :-)
Force setting the currentProgress also for coreAnimationEngine
```diff
-        // For the Main Thread rendering engine, we freeze the animation at the expected final progress
-        // once the animation is complete. This isn't necessary on the Core Animation engine.
-        if finished, !(self.animationView.animationLayer is CoreAnimationLayer) {
+       if finished {
            self.animationView.currentProgress = finalProgress
        }
```

## Expected Behavior 

https://user-images.githubusercontent.com/36905/209103417-eff097ab-a3ff-4519-9443-ea5dd0d17a7c.mp4

## Actual Behavior

https://user-images.githubusercontent.com/36905/209103485-d7eadef8-a282-43eb-8b40-ffc7ddf96ca4.mp4

## Animation JSON
[myListButton.json.zip](https://github.com/airbnb/lottie-ios/files/10285647/myListButton.json.zip)

calda:
This PR fixes an issue where `LottieAnimationView.currentFrame` would incorrectly be `0` after playing an animation with `LottieLoopMode.playOnce` and using the Core Animation rendering engine. This issue was breaking `AnimatedSwitch` in some situations, and was also inconsistent with the main thread rendering engine. Fixes #1865. Fixes #1877.


Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -1390,6 +1390,13 @@ final public class LottieAnimationView: LottieAnimationViewBase {
         }
       }
 
+      // If attempting to play a zero-duration animation, just pause on that single frame instead
+      if animationContext.playFrom == animationContext.playTo {
```
calda, line 1394: This change was also required to fix https://github.com/airbnb/lottie-ios/issues/1877 -- the logic here was causing us to attempt to play a zero-duration animation, which seems to not be supported properly by the Core Animation engine. Instead we can just pause the animation at that frame.

Tests/AnimationViewTests.swift
```diff
@@ -61,4 +61,66 @@ final class AnimationViewTests: XCTestCase {
     wait(for: [expectation], timeout: 1.0)
   }
 
+  func testPlayFromFrameToFrame() {
```
calda, line 64: Pulled in this test case from https://github.com/airbnb/lottie-ios/pull/1892 -- thanks @nevil!

erichoracek (approved):

nevil:
@calda 

There is a small behavior difference between Main Thread Engine and Core Animation Engine.
In the sample app, play animation issue_1877.
- Play once
- Once animation is paused, select play once again.

With Main Thread engine the animation plays from frame 0.
With the Core Animation Engine the animation doesn't play.


Not sure if this is worth fixing...

calda:
Mind filing an issue for this remaining problem? I'll take a look.

PR #1908, opened 2023-01-09
Fix issue where `play(fromProgress: 1, toProgress: 0)` animation would not actually play

### LINKED ISSUE #1901: Minor AnimatedSwitch behavior difference between the Core Animation Engine and the Main Thread Engine

Latest master (03eeda861705c742af3ede84711f7c3ace1ac7e2 at this time)

## Description
Relates to #1877 

Can be reproduced using the sample app.
- Set the engine to Main Thread Engine.
- Open Controls Demo
--> Observe that the `Issue #1877` switch is animating.

- Set the engine to Core Animation Engine.
- Open Controls Demo
--> Observe that the `Issue #1877` switch is set to the final frame immediately.

I guess it is a question of how closely the Core Animation Engine should copy Main Thread Engine's behavior...

calda:
This PR fixes an issue where animations would be ignored when `toFrame == currentFrame`. For example, when playing from 100% to 0% when the animation view is created (and `currentFrame == 0`), the animation was previously ignored. Fixes #1901 and #1907.

| Before | After |
| ---- | ---- |
|  |  |

### DIFF
--- a/Sources/Public/Animation/LottieAnimationView.swift
+++ b/Sources/Public/Animation/LottieAnimationView.swift
@@ -1374,7 +1374,7 @@ final public class LottieAnimationView: LottieAnimationViewBase {
         // `playFrom` time to be the `currentFrame`. Since the animation duration
         // is based on `playFrom` and `playTo`, this automatically truncates the
         // duration (so the animation stops playing at `playFrom`).
-        //  - Don't do this is the animation is already at that frame
+        //  - Don't do this if the animation is already at that frame
         //    (e.g. playing from 100% to 0% when the animation is already at 0%)
         //    since that would cause the animation to not play at all.
         case .playOnce:
--- a/Sources/Public/Animation/LottieAnimationView.swift
+++ b/Sources/Public/Animation/LottieAnimationView.swift
@@ -1374,8 +1374,13 @@ final public class LottieAnimationView: LottieAnimationViewBase {
         // `playFrom` time to be the `currentFrame`. Since the animation duration
         // is based on `playFrom` and `playTo`, this automatically truncates the
         // duration (so the animation stops playing at `playFrom`).
+        //  - Don't do this is the animation is already at that frame
+        //    (e.g. playing from 100% to 0% when the animation is already at 0%)
+        //    since that would cause the animation to not play at all.
         case .playOnce:
-          animationContext.playFrom = currentFrame
+          if animationContext.playTo != currentFrame {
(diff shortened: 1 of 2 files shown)


Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -1374,8 +1374,13 @@ final public class LottieAnimationView: LottieAnimationViewBase {
         // `playFrom` time to be the `currentFrame`. Since the animation duration
         // is based on `playFrom` and `playTo`, this automatically truncates the
         // duration (so the animation stops playing at `playFrom`).
+        //  - Don't do this is the animation is already at that frame
+        //    (e.g. playing from 100% to 0% when the animation is already at 0%)
+        //    since that would cause the animation to not play at all.
         case .playOnce:
-          animationContext.playFrom = currentFrame
+          if animationContext.playTo != currentFrame {
```
calda, line 1381: This is basically the same logic that the Main Thread rendering engine has here: https://github.com/airbnb/lottie-ios/blob/master/Sources/Public/Animation/LottieAnimationView.swift#L1422

Tests/AnimationViewTests.swift
```diff
@@ -65,6 +65,7 @@ final class AnimationViewTests: XCTestCase {
     let tests: [(fromFrame: AnimationFrameTime?, toFrame: AnimationFrameTime)] = [
       (fromFrame: nil, toFrame: 10),
       (fromFrame: 8, toFrame: 14),
+      (fromFrame: 14, toFrame: 0),
```
calda, line 68: Added a test case for this, which fails on master

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -1374,8 +1374,13 @@ final public class LottieAnimationView: LottieAnimationViewBase {
         // `playFrom` time to be the `currentFrame`. Since the animation duration
         // is based on `playFrom` and `playTo`, this automatically truncates the
         // duration (so the animation stops playing at `playFrom`).
+        //  - Don't do this is the animation is already at that frame
```
erichoracek, line 1377:
```suggestion
        //  - Don't do this if the animation is already at that frame
```

calda: ty

calda: Xcode has spell check but what I really need seems to be grammar check

PR #1909, opened 2023-01-11
Add BSD-2-Clause to Lottie license file.

snow-google:
This license is referenced in /Sources/Private/Utility/Primitives/UnitBezier.swift but not added to the LICENSE file.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,3 +199,27 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+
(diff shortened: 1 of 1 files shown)


calda:
I think we're in compliance with the license for this code since we retain the license / copyright notice in the `UnitBezier.swift` source file when distributing Lottie as source. I'd prefer to not modify the project's LICENSE file unless there's some indication that this would be mandatory to comply with the license. Let me know if you think there are any particular reasons this would be necessary.

snow-google:
Thanks for the context! Whenever we pull in this library, we have to add that license to the LICENSE file to comply with out compliance requirements. If this is not something other people run into, then I will move forward with a local patch. Thanks

PR #1921, opened 2023-01-18
Add info about lottie-spm repo

calda:
This PR adds information about the new [lottie-spm](https://github.com/airbnb/lottie-spm) repo. I think it makes sense for this to be the recommended way to integrate Lottie via Swift Package Manager.

I'll release 4.1.0 after merging this PR.


README.md
```diff
@@ -71,29 +75,6 @@ carthage update
 ```
 In your application targets “General” tab under the “Linked Frameworks and Libraries” section, drag and drop lottie-ios.framework from the Carthage/Build/iOS directory that `carthage update` produced.
 
-### Binary dependency (`xcframework`)
```
calda, line 74: I don't think we need to recommend this approach anymore, since it was primarily a workaround for the large repo size when using SPM. I think the UX for integrating with the lottie-spm repo is much better than manually managing XCFrameworks / checksums. Using lottie-spm also probably makes it easier for folks to stay up-to-date (since the standard `swift package update` flow still works).

calda:
ptal @erichoracek

erichoracek (approved):
@calda LGTM! Frustrating that we need to use this workaround, it would be good to elevate this to the SPM folks in some form

calda:
I filed an issue on SPM here: https://github.com/apple/swift-package-manager/issues/6062

PR #1923, opened 2023-01-19
Update Package.resolved version in Lottie.xcodeproj to fix Swift Package Index support Xcode 13.2.1 / Swift 5.5

calda:
I noticed Swift Package Index was reporting that we don't support Swift 5.5. Chatting with them in https://github.com/SwiftPackageIndex/SwiftPackageIndex-Server/issues/2225 pointed to an issue with the `Package.resolved` inside the `Lottie.xcodeproj` bundle. Updating it from `"version" : 2` to `"version" : 1` should fix the issue.


erichoracek (approved):
huh!

PR #1924, opened 2023-01-19
Fix issue where XCFramework archive was double-zipped and rejected by SPM

calda:
The XCFramework uploaded by the CI job was double-zipped (a `.zip` that contains a `.zip`), causing it to be rejected by SPM. This PR fixes the job to upload a `.zip` that actually works when used as an SPM dependencies.


erichoracek (approved):
😅

PR #1930, opened 2023-01-23
#1927 Add rotation on x and y axis for animation

ipodishima:
Following the discussion on #1927, I needed to rotate objects "for real" along the x / y axis. Especially on Text layers.
It might benefit the community so creating a PR, if ever it's not creating too much of a regression


Sources/Private/MainThread/LayerContainers/Utility/LayerTransformNode.swift
```diff
@@ -18,13 +18,17 @@ final class LayerTransformProperties: NodePropertyMap, KeypathSearchable {
...
-    rotation = NodeProperty(provider: KeyframeInterpolator(keyframes: transform.rotation.keyframes))
+    rotationX = NodeProperty(provider: KeyframeInterpolator(keyframes: transform.rotationX.keyframes))
+    rotationY = NodeProperty(provider: KeyframeInterpolator(keyframes: transform.rotationY.keyframes))
+    rotationZ = NodeProperty(provider: KeyframeInterpolator(keyframes: transform.rotationZ.keyframes))
     opacity = NodeProperty(provider: KeyframeInterpolator(keyframes: transform.opacity.keyframes))
 
     var propertyMap: [String: AnyNodeProperty] = [
       "Anchor Point" : anchor,
       "Scale" : scale,
-      "Rotation" : rotation,
```
calda, line 27: Removing support for `"*.Rotation"` animation key paths would be a breaking change for anyone relying on this functionality. Is there a way to continue supporting `Rotation` here as well as X/Y/Z?

calda: You can see in [Tests/__Snapshots__/AnimationKeypathTests/testMainThreadEngineKeypathLogging.Switch.txt](https://github.com/airbnb/lottie-ios/pull/1930/files#diff-5bd94bedf0fc0d5657b62b8eb004d4c4b8fc57257dc302ae516535cecf4c4d13) how this is part of the public API and how it would cause a breaking change

ipodishima: Yes. I can bring it back. I wasn't sure how this was used

calda:
Nice! Could you please also add at least one sample animation json (like at `Tests/Samples/Issues/pr_1930.json`) that includes X and Y rotation? We'll want a test case for this functionality.

ipodishima:
Yes! Generating it as we speak. I was waiting for the PR number.
Also, you'll see (I'll tag you) but the screenshotting is not working properly where as the sample code runs perfectly

ipodishima:
@calda See this one. idk why but it's not properly screenshotting the animation.
Could also be because of the way I created the files. I started with this [lottie](https://lottiefiles.com/93914-rotation-04) and updated the rotation key in the json

calda:
Pretty weird, but that's ok. I've seen issues like this before where certain Core Animation functionality (particularly with transforms) works at runtime in the simulator but doesn't work correctly in the snapshot tests. Here's one example: https://github.com/airbnb/lottie-ios/blob/420c411204208d0fa8ef12e8c847e93d93444191/Sources/Private/CoreAnimation/Animations/TransformAnimations.swift#L189-L194

Running this locally is seems like we have that problem for X and Y rotations. They look good in the simulator which is good enough. If we accidentally break or change this code in the future I'd expect it would still cause a diff in the snapshot, which is what really matters. Could you add a log message like the one above for the X / Y rotations in `TransformAnimations.addRotationAnimation`? That's good enough for me.

Sources/Private/Model/Objects/Transform.swift
```diff
@@ -169,15 +203,21 @@ final class Transform: Codable, DictionaryInitializable {
...
+  /// The rotation of the transform on Z axis.
+  let rotationZ: KeyframeGroup<LottieVector1D>
 
   /// The opacity of the transform.
   let opacity: KeyframeGroup<LottieVector1D>
 
-  /// Should always be nil.
-  let rotationZ: KeyframeGroup<LottieVector1D>?
+  /// Rotation. Will always be nil (It's here for `Codable`)
+  let rotation: KeyframeGroup<LottieVector1D>?
```
calda, line 222: Is there a way for us to remove this property, since it's now unused? Why would we need to keep it around for `Codable` -- could we just keep its coding key `case rotation = "r"` defined above?

README.md
```diff
@@ -88,7 +88,7 @@ We always appreciate contributions from the community. To make changes to the pr
 
 All pull requests with new features or bug fixes that affect how animations render should include snapshot test cases that validate the included changes. 
   - To add a new sample animation to the snapshot testing suite, you can add the `.json` file to `Tests/Samples`. Re-run the snapshot tests to generate the new snapshot image files.
-  - To update existing snapshots after making changes, you can set `isRecording = true` in `SnapshotTests.swift` and then re-run the snapshot tests.
+  - To update existing snapshots after making changes, you can set `AssertSnapshot.isRecording = true` in `SnapshotTests.swift` and then re-run the snapshot tests.
```
calda, line 91: What is this change for? I don't see `AssertSnapshot` defined anywhere

Sources/Private/Model/Objects/Transform.swift
ipodishima: It's `Codable` screaming the `CodingKeys.rotation` should have a property on the `Transform` class
Screenshot 2023-01-23 at 15 49 17
I think there is a way to trick it by having an enum wrapping the different keys for the `rotationZ`

Something like


```
enum RotationZ: Codable {
   case r(...)
   case rz(...)

   enum CodingKeys: ... {
         case r, rz
   }
} 
```

I can explore this avenue if you want.
I don't really like having a ghost property either

README.md
ipodishima: Oooooups my bad, bad paste

You should read `SnapshotTesting.isRecording = true`

Took me some minutes to figure out where to set this property to `true`

calda:
There seems to be something off with `Issues/issue_1800.json` -- these letters are getting clipped strangely while animating, when using the Core Animation rendering engine. Otherwise all of the tests pass without any issues.

image

README.md
calda: I usually just add `isRecording = true` as the first line of `SnapshotTests.testMainThreadRenderingEngine` or `SnapshotTests.testCoreAnimationRenderingEngine`, like this:

```swift
  func testCoreAnimationRenderingEngine() async throws {
    isRecording = true
    try await compareSampleSnapshots(configuration: LottieConfiguration(renderingEngine: .coreAnimation))
  }
```

calda: We could potentially commit `isRecording = false` here so it's more obvious where you're supposed to set it to `true`

ipodishima:
Yes, when I run on generation mode, it's updating it (see the T appearing later in the new)
Screenshot 2023-01-23 at 15 56 41

I found out why. In the PR_1800.json, there were rx and ry properties set `"rx":{"a":0,"k":0,"ix":8},"ry":{"a":1,"k":[{"i":{"x":[0.833],"y":[0.833]},"o":{"x":[0.167],"y":[0.167]},"t":41,"s":[-125]},{"t":59,"s":[0]}],"ix":9}`

So I just removed them given it's messing with the screenshot things and is now working correctly

I think I fixed all your comments

Sources/Private/MainThread/LayerContainers/Utility/LayerTransformNode.swift
ipodishima: Done

README.md
ipodishima: Done

calda:
I would prefer we don't modify that animation and instead fix the runtime issue. The issue in the snapshot test doesn't really matter. When you test the unmodified animation on lottiefiles.com or with the Main Thread rendering engine in the simulator you can see the x/y rotation animations work great. The only issue is when testing it in the simulator with the Core Animation rendering engine.

ipodishima:
OK so you would prefer that I keep the changes on issue_1800.json, but instead update the two snapshots failing (75%)?

calda:
Yes, and we'll also need to investigate why the animation plays incorrectly at runtime when using the Core Animation rendering engine. If we can't fix that issue I think we'd need to mark this feature as only supported by the Main Thread rendering engine.

ipodishima:
Not sure that we're talking about the same thing

The issue_1800 renders correctly on the sim with the CoreAnimation engine. The screenshot failed in the test because it contains some `rx` and `ry` properties on the letters, which messes up how the screenshot behaves.
- If I remove those properties from the json, both the animation and the screenshot (aka tests) are fine
- If I keep those properties, then I have to update one of the frame of the snapshot because it does not display the T on the same 75% timing.

Regarding what I did on the PR, aka `rx` and `ry` it runs without issues on runtime, on the CoreAnimation engine

calda:
> The issue_1800 renders correctly on the sim with the CoreAnimation engine. The screenshot failed in the test because it contains some `rx` and `ry` properties on the letters, which messes up how the screenshot behaves.

I don't think this is the case from testing it locally. I see these letters being clipped incorrectly while the rotation animation is playing:

https://github.com/airbnb/lottie-ios/pull/1930#issuecomment-1400481517

Can you share a gif or screen recording of what you see locally? It's possible we're seeing different animations somehow

ipodishima:
Sure

iPhone 14 pro sim running iOS 16.2

https://user-images.githubusercontent.com/968830/214089879-f38208b4-b3c7-4871-aac2-bbea67db0775.mp4

calda:
I think you're seeing a different animation because of the changes to the animation json in https://github.com/airbnb/lottie-ios/pull/1930/commits/5a146529da9cb0b1cf59475013d83e8327967985. When I revert that commit and use the original un-modified sample json, this is what I see:

https://user-images.githubusercontent.com/1811727/214092178-21cc6a58-227a-4b1f-8047-b107b0ca539b.mov

calda:
If we aren't able to find a fix for this then the easiest way to unblock this PR would be to add a `compatibilityAssert` to `TransformAnimations.addRotationAnimation` like this:

```swift
// Due to issues with Samples/Issues/issue_1800.json, we can't enable X/Y rotation by default
try context.compatibilityAssert(
  !containsXRotationValues && !containsYRotationValues,
  "The Core Animation rendering engine doesn't fully-support X / Y rotation.")
```

This would make it so `issue_1800` still animates correctly when using `RenderingEngineOption.automatic` (by falling-back to the main thread rendering engine).

Sources/Private/Model/Objects/Transform.swift
calda: Would also be nice to make this property private if possible

ipodishima:
But I need it running on CoreAnimation ahahah, because my project is indeed using Lottie in an `AVSynchronizedLayer`.
But this one is gonna be a PR for later, given it's still a WIP (but a working WIP)

Don't you think it's rather an artefact issue from the `rx`/`ry` values which where not supposed to be here on the `issue_1800.json` instead?
The bad thing would be that anyone using files with those artefacts might have an issue running the new update with this PR though :(

calda:
> The bad thing would be that anyone using files with those artefacts might have an issue running the new update with this PR though :(

Yeah, exactly 😔

> But I need it running on CoreAnimation ahahah, because my project is indeed using Lottie in an AVSynchronizedLayer.

You can always force Lottie to use the Core Animation rendering engine by passing in `renderingEngineOption: .coreAnimation` instead of `.automatic` (the default). Lottie would emit an assertion, but you can suppress the assertion by configuring the `LottieLogger` like `LottieLogger.shared = .printToConsole` (don't call `assertionFailure` and instead just call `print`).

We could merge this PR with a `compatibilityAssert` and then try to investigate the issue in a follow-up. If you file an issue about this I can find time to take a look.

Sources/Private/CoreAnimation/Animations/TransformAnimations.swift
```diff
@@ -255,6 +255,11 @@ extension CALayer {
           `rotation.y` values are not displayed correctly in snapshot tests
           """)
       }
+    } else {
+      // Due to issues with Samples/Issues/issue_1800.json, we can't enable X/Y rotation by default
+      try context.compatibilityAssert(
+        !containsXRotationValues && !containsYRotationValues,
+        "The Core Animation rendering engine doesn't fully-support X / Y rotation.")
     }
```
calda, line 263: We should do this even when running the snapshot tests. One of the test cases uses these assertions to track which sample json files are/aren't supported by the Core Animation rendering engine (in this case, `Test/__Snapshots__/AutomaticEngineTests/testAutomaticEngineDetection.Issues-issue_1800.txt`)

```suggestion
    }
    
    // Due to issues with Samples/Issues/issue_1800.json, we can't enable X/Y rotation by default
    try context.compatibilityAssert(
      !containsXRotationValues && !containsYRotationValues,
      "The Core Animation rendering engine doesn't fully-support X / Y rotation.")
```

calda (approved):
Thanks for the great contribution!

ipodishima:
Got it

Just made the change on private for `rotate` and added the `compatibilityAssert`

Also, I've been digging a bit about the issue and I might have a hint. Documented it on #1932 #1932

Sources/Private/Model/Objects/Transform.swift
ipodishima: Private was enough. Good suggestion, was overthinking it ahah

Sources/Private/CoreAnimation/Animations/TransformAnimations.swift
ipodishima: OK I see. Done it, just waiting for the UT to pass before committing

calda:
Good news, I found a solution and pushed a few commits. Now the animation plays correctly:



Your notes in #1932 were helpful -- the letters were clipping behind the white circle. This doesn't happen with the Main Thread rendering engine.

Inconveniently, the Xcode view debugger doesn't with with the Core Animation rendering engine. It _does_ work with the Main Thread rendering engine though. We can see here that the entirety of the letter layers are in front of the circle, with respect to the z axis:

Screenshot 2023-01-23 at 1 34 32 PM

This appears to be because of how the Main Thread rendering engine's base layer type inserts an additional `contentsLayer` in the hierarchy and applies the transform to that `contentsLayer` instead of the parent layer. That must affect how Core Animation orders / spaces layers in the Z axis.

I updated the Core Animation rendering engine to work the same way in https://github.com/airbnb/lottie-ios/pull/1930/commits/7fb751a4bd3db665f2bc8c65f270a2195646a88f, and that resolved the issue!

ipodishima:
Nice job!!

Thanks for the really really quick help and jump on the PR

PR #1937, opened 2023-01-25
make LottieAnimationView open & expose key methods

lehmacdj:
At Duolingo we disable certain animations to comply with the reduced motion and soon will also be disabling certain animations in order to improve performance on older devices. To implement this functionality we currently have a class that wraps a LottieAnimationView forwarding method calls to its child LottieAnimationView while performing some extra checks around some of the methods.

It would be neater if we could override some of the methods in LottieAnimationView instead of wrapping it. This is possible in our android app because LottieAnimationView [isn't declared as final](https://github.com/airbnb/lottie-android/blob/master/lottie/src/main/java/com/airbnb/lottie/LottieAnimationView.java) but on ios it is.

Thoughts on allowing overriding?

I exposed only animation methods (e.g. play, pause, stop) & frame progress properties, because those seem like the methods that make the most sense to make open to overriding, and are sufficient for our needs.

### DIFF
--- a/Sources/Public/Animation/LottieAnimationView.swift
+++ b/Sources/Public/Animation/LottieAnimationView.swift
@@ -346,61 +346,6 @@ open class LottieAnimationView: LottieAnimationViewBase {
   /// Value Providers that have been registered using `setValueProvider(_:keypath:)`
   public private(set) var valueProviders = [AnimationKeypath: AnyValueProvider]()
 
-  /// Sets the current animation time with a Progress Time
-  ///
-  /// Note: Setting this will stop the current animation, if any.
-  /// Note 2: If `animation` is nil, setting this will fallback to 0
-  public var currentProgress: AnimationProgressTime {
-    set {
-      if let animation = animation {
-        currentFrame = animation.frameTime(forProgress: newValue)
-      } else {
-        currentFrame = 0
-      }
-    }
-    get {
-      if let animation = animation {
-        return animation.progressTime(forFrame: currentFrame)
-      } else {
-        return 0
-      }
-    }
-  }
-
-  /// Sets the current animation time with a time in seconds.
-  ///
-  /// Note: Setting this will stop the current animation, if any.
-  /// Note 2: If `animation` is nil, setting this will fallback to 0
-  public var currentTime: TimeInterval {
-    set {
-      if let animation = animation {
-        currentFrame = animation.frameTime(forTime: newValue)
-      } else {
-        currentFrame = 0
-      }
-    }
-    get {
-      if let animation = animation {
-        return animation.time(forFrame: currentFrame)
-      } else {
-        return 0
-      }
-    }
-  }
-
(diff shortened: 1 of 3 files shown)


Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -178,6 +178,221 @@ final public class LottieAnimationView: LottieAnimationViewBase {
     commonInit()
   }
 
+  // MARK: Open
+
+  /// Sets the current animation time with a Progress Time
+  ///
+  /// Note: Setting this will stop the current animation, if any.
+  /// Note 2: If `animation` is nil, setting this will fallback to 0
+  open var currentProgress: AnimationProgressTime {
```
calda, line 187: I see the rationale of wanting to override the `play` methods to provide custom functionality. What use case do you have in mind for overriding `currentProgress` / `currentTime` / `currentFrame`? What sort of additional logic would you need to layer on top of these?

lehmacdj: My also allowing these to be overridden was speculative and can be left out of this PR. We aren't actually overriding these at the moment, I was just anticipating that we might want to in the future.

For example, if we are ignoring play requests (e.g. in response to reduce motion) and keeping the animation static, it might also makes sense to ignore changes to the current frame to avoid the animation jumping. We don't have a concrete case like this yet though.

calda: Cool, in that case I'd prefer if we left these properties as just `public`. Totally fine with marking the various `play` methods as open though!

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -186,6 +346,61 @@ final public class LottieAnimationView: LottieAnimationViewBase {
   /// Value Providers that have been registered using `setValueProvider(_:keypath:)`
   public private(set) var valueProviders = [AnimationKeypath: AnyValueProvider]()
 
+  /// Sets the current animation time with a Progress Time
+  ///
+  /// Note: Setting this will stop the current animation, if any.
+  /// Note 2: If `animation` is nil, setting this will fallback to 0
+  public var currentProgress: AnimationProgressTime {
```
calda, line 353: Can you revert the changes to these properties? Otherwise LGTM

calda (approved):

Sources/Public/Animation/LottieAnimationView.swift
lehmacdj: you want me to make `currentProgress`/`currentFrame`/`currentTime` open again? Got the change to make them public in very fast before you commented this

calda: Sorry, I mean make it so those properties aren't included this PR / the diff. It looks like they were moved to a different place in the file.

calda (approved):
Thanks!

PR #1944, opened 2023-02-02
Expose AnimationTextProvider in objc wrapper

ciss1995:
## This PR updates CompatibleAnimationView to expose textProvider. 

Exposing `textProvider` in `CompatibleAnimationView` will allow usage of Lottie in Objective-C access to the api to dynamically set texts in animations. 

Updates to `CompatibleAnimationView`:
- New class `CompatibleAnimationTextProvider` with class member returning an `AnimationTextProvider`.
- New public class member to CompatibleAnimationView to set textProvider from `CompatibleAnimationTextProvider`.

### DIFF
--- a/Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
+++ b/Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
@@ -190,9 +190,7 @@ public final class CompatibleAnimationView: UIView {
   @objc
   public var compatibleAnimationTextProvider: CompatibleAnimationTextProvider? {
     didSet {
-      if let textProvider = compatibleAnimationTextProvider?.textProvider {
-        animationView.textProvider = textProvider
-      }
+      animationView.textProvider = compatibleAnimationTextProvider?.textProvider ?? DefaultTextProvider()
     }
   }
 
--- a/Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
+++ b/Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
@@ -470,8 +470,7 @@ public final class CompatibleAnimationTextProvider: NSObject {
   // MARK: Lifecycle
 
   @objc
-  public init(values: [String: String])
-  {
+  public init(values: [String: String]) {
     self.values = values
     super.init()
   }
@@ -479,7 +478,7 @@ public final class CompatibleAnimationTextProvider: NSObject {
   // MARK: Internal
 
   internal var textProvider: AnimationTextProvider? {
-    return DictionaryTextProvider(self.values)
+    DictionaryTextProvider(values)
   }
 
   // MARK: Private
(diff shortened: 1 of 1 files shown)


Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
```diff
@@ -187,6 +187,15 @@ public final class CompatibleAnimationView: UIView {
     }
   }
 
+  @objc
+  public var compatibleAnimationTextProvider: CompatibleAnimationTextProvider? {
+    didSet {
+      if let textProvider = compatibleAnimationTextProvider?.textProvider {
+        animationView.textProvider = textProvider;
```
calda, line 194: 😄 

```suggestion
        animationView.textProvider = textProvider
```

Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
```diff
@@ -451,4 +460,30 @@ public final class CompatibleAnimationView: UIView {
...
+  public init(values: [String: String])
+  {
+    self.values = values
+    super.init()
+  }
+
+  // MARK: Internal
+
+  internal var textProvider: AnimationTextProvider? {
+    return DictionaryTextProvider(self.values);
```
calda, line 482:
```suggestion
    return DictionaryTextProvider(self.values)
```

calda (approved):
Thank you!

Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
ciss1995: Ah thanks :)

calda:
I noticed the Swift linter check was failing, I pulled this branch, ran `bundle exec rake format:swift`, and pushed a commit.

ciss1995:
> I noticed the Swift linter check was failing, I pulled this branch, ran `bundle exec rake format:swift`, and pushed a commit.

Awesome! Good catch! Thanks a lot.

Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift
```diff
@@ -187,6 +187,15 @@ public final class CompatibleAnimationView: UIView {
     }
   }
 
+  @objc
+  public var compatibleAnimationTextProvider: CompatibleAnimationTextProvider? {
+    didSet {
+      if let textProvider = compatibleAnimationTextProvider?.textProvider {
+        animationView.textProvider = textProvider
+      }
```
calda, line 195: We probably want to handle the case where `compatibleAnimationTextProvider` is set to `nil`. Previously this setter did nothing in that case, so `animationView.textProvider` would still have the previous value.

```suggestion
      animationView.textProvider = compatibleAnimationTextProvider?.textProvider ?? DefaultTextProvider()
```

campersau:
This ties the implementation to use a dictionary. Exposing the `CompatibleAnimationTextProvider` as a protocol might be more flexible?

calda:
Maybe we should call this `CompatibleDictionaryTextProvider` instead to leave things open for future directions like that. What do you think @ciss1995?

ciss1995:
> Maybe we should call this `CompatibleDictionaryTextProvider` instead to leave things open for future directions like that. What do you think @ciss1995?

Sure! Sounds great to me

calda:
Thanks all! Appreciate it.

PR #1956, opened 2023-02-15
Add dispatch queue parameter to dotLottie loading methods

mkj-is:
## Reason

In some cases dotLottie animations will not be loaded correctly due to asynchronous nature of the loading mechanism.

This is due to race condition inside `DotLottieFile` loading methods. It uses `DispatchQueue.global()` which is concurrent. Loading the animation might fail in these examples:

- Initialisation of more animation views (typical for cells in collection/table view) with the same animation.
- If you have some stateful animation view (let's say play/pause button) and the user taps it quickly, then there might be race condition in case the animation was not loaded in time before.

```swift
playButton.updateAnimation(true)
playButton.updateAnimation(false)
playButton.updateAnimation(true)
playButton.updateAnimation(false)
```

```swift
func updateAnimation(_ playing: Bool) {
    DotLottieFile.asset(playing ? "pause" : "play") { success in
        guard case let .success(animation) = result else { return }
        animationView.loadAnimation(from: animation)
    }
}
```

Of course this depends on the usage of Lottie in your code, but it makes rewrites to async loading much more complex. Fixing this outside of Lottie is possible, by keeping flags with information about animation loading, wrapping everything in OperationQueue etc. But all of these solutions are hard to implement and would cause additional overhead.

## Solution

Add parameter to pass dispatch queue to dotLottie loading methods to enable passing serial or higher priority queue.

### Adding DispatchQueue to Lottie configuration

Can be combined with this solution. Did not add this since Lottie configuration has only few flags and did not want to add any unnecessary feature flags to cause feature flag hell.

### Using serial queue by default.

Wanted to make this backward compatible and do not break current behavior.

### DIFF
--- a/Sources/Public/DotLottie/DotLottieFileHelpers.swift
+++ b/Sources/Public/DotLottie/DotLottieFileHelpers.swift
@@ -36,15 +36,17 @@ extension DotLottieFile {
   /// - Parameter bundle: The bundle in which the lottie is located. Defaults to `Bundle.main`
   /// - Parameter subdirectory: A subdirectory in the bundle in which the lottie is located. Optional.
   /// - Parameter dotLottieCache: A cache for holding loaded lotties. Defaults to `LRUDotLottieCache.sharedCache`. Optional.
+  /// - Parameter dispatchQueue: A dispatch queue used to load animations. Defaults to `DispatchQueue.global()`. Optional.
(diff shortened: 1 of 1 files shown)


calda (approved):
Thanks!

calda (approved):

PR #1964, opened 2023-02-17
Fix issue where rounded corners algorithm would return incorrect path for shapes without any rounded corners

calda:
This PR fixes an issue where the rounded corners algorithm used in #1795 would return an incorrect path for shapes that don't actually have any rounded corners.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 8 - 2023-02-17 at 11 49 41 | Simulator Screen Shot - iPhone 8 - 2023-02-17 at 11 49 54 |


thedrick (approved):
LGTM @calda

PR #1966, opened 2023-02-19
Fix compilation of iOS example project in Xcode 14.3 beta

mkj-is:
Removes 2 unnecessary conditional unwraps in example project, which now throw compilation error in the new Xcode 14.3 beta.

Screenshot 2023-02-19 at 22 18 35

### DIFF
--- a/Example/iOS/Views/AnimatedButtonRow.swift
+++ b/Example/iOS/Views/AnimatedButtonRow.swift
@@ -57,33 +57,31 @@ final class AnimatedButtonRow: UIView, EpoxyableView {
     guard let content = content else { return }
 
     group.setItems {
-      if let animationName = content.animationName {
-        GroupItem<AnimatedButton>(
-          dataID: DataID.animatedButton,
-          content: content,
-          make: {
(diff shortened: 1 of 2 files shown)


Example/iOS/Views/AnimatedButtonRow.swift
```diff
@@ -57,33 +57,31 @@ final class AnimatedButtonRow: UIView, EpoxyableView {
     guard let content = content else { return }
 
     group.setItems {
-      if let animationName = content.animationName {
```
calda, line 60: Weird! I wonder how this compiled before considering `content.animationName` is non-optional. Must have been a compiler bug.

calda (approved):
Thanks for the fix

Example/iOS/Views/AnimatedButtonRow.swift
calda: Filed an issue for this on the Swift repo since I think this source break was probably unintentional: https://github.com/apple/swift/issues/63808

Issue #1967, opened 2023-02-20
possibility to load dotLottie files synchronously


ilendemli:
## Which Version of Lottie are you using?
Lottie 4.1.3

In our use case, we have to use the dotLottie file synchronously and I couldn't find any method that could provide this.

PR #1968, opened 2023-02-20
add support for loading DotLottieFile by name and filename synchronously

ilendemli:
We need to be able to load `dotLottie` files synchronously in out project. This PR enables us to create `DotLottieFile` directly.

Reference: https://github.com/airbnb/lottie-ios/issues/1967

### DIFF
--- a/Sources/Public/DotLottie/DotLottieFileHelpers.swift
+++ b/Sources/Public/DotLottie/DotLottieFileHelpers.swift
@@ -47,7 +47,7 @@ extension DotLottieFile {
     handleResult: @escaping (Result<DotLottieFile, Error>) -> Void)
   {
     dispatchQueue.async {
-      let result = named(name, bundle: bundle, subdirectory: subdirectory, dotLottieCache: dotLottieCache)
+      let result = SynchronouslyBlockingCurrentThread.named(name, bundle: bundle, subdirectory: subdirectory, dotLottieCache: dotLottieCache)
 
       DispatchQueue.main.async {
         handleResult(result)
@@ -83,7 +83,7 @@ extension DotLottieFile {
     handleResult: @escaping (Result<DotLottieFile, Error>) -> Void)
   {
     dispatchQueue.async {
-      let result = loadedFrom(filepath: filepath, dotLottieCache: dotLottieCache)
+      let result = SynchronouslyBlockingCurrentThread.loadedFrom(filepath: filepath, dotLottieCache: dotLottieCache)
 
       DispatchQueue.main.async {
         handleResult(result)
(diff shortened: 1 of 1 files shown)


Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -268,4 +217,71 @@ extension DotLottieFile {
...
+      let lottie = dotLottieCache.file(forKey: filepath)
+    {
+      return .success(lottie)
+    }
+
+    do {
+      /// Decode the lottie.
+      let url = URL(fileURLWithPath: filepath)
+      let data = try Data(contentsOf: url)
+      let lottie = try DotLottieFile(data: data, filename: url.deletingPathExtension().lastPathComponent)
```
calda, line 240: I don't think we should provide a synchronous method for loading DotLottie files. This initializer performs file system operations, like decompressing `data` to a temporary directly and reading various files. This would block the current thread, so we need to make sure we aren't making it too easy for folks to accidentally call this from the main thread.

One reason I think it's especially important to not provide synchronous loading methods for `DotLottieFile`s is because we _do_ provide synchronous loading methods for `LottieAnimation`s (which are less problematic to load on the main thread). If we provide synchronous methods for both I worry the difference would be too subtle.

calda:
@ilendemli, can you share more details about your use case here? Why do you want to load these files synchronously? Are you loading them on a background thread or on the main thread? Why can't you use the existing async methods?

ilendemli:
@calda In our case, when the app starts we show a static image of the first frame of the lottie animation as the launch screen. When the app is loaded into memory we show a splash screen which is basically the same as the launch screen but with the dotlottie loaded and animated. If we load the animation asynchronously then there is a white flash/screen until the dotlottie is loaded and displayed. I know this is blocking the main thread during the operation. But as it is during the startup process of the app, it's less noticeable. I would suggest to inform the user to use the asynchronous methods whenever possible, but in this (and a few other cases where we have to use dotlottie files) we have to use synchronous loading.

calda:
Intentionally doing this on the main thread seems relatively uncommon and something I think we probably don't want to "officially" support for now. I'm open to reconsidering if we get more requests for this in the future, though. Have you explored using a something like a [`DispatchGroup`](https://stackoverflow.com/questions/42484281/waiting-until-the-task-finishes) to manually block the main thread while waiting for the async operation to complete? That seems like it would work for your use case without having to change the API of Lottie itself.

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -268,4 +217,71 @@ extension DotLottieFile {
     }
   }
 
+  /// Loads an DotLottie from a specific filepath synchronously.
+  /// - Parameter filepath: The absolute filepath of the lottie to load. EG "/User/Me/starAnimation.lottie"
+  /// - Parameter dotLottieCache: A cache for holding loaded lotties. Defaults to `LRUDotLottieCache.sharedCache`. Optional.
+  public static func loadedFrom(
```
calda, line 223: I could also imagine being ok with merging this change if we give this method a more scary sounding name like:

```suggestion
  public static func loadSynchronouslyBlockingCurrentThread(
```

with documentation explaining that this performs file system operations like decompressing a zip archive.

What do you think?

calda (approved):
This feels reasonable, thanks.

PR #1969, opened 2023-02-21
Add dotLottie error handling

mkj-is:
Previously discussed in https://github.com/airbnb/lottie-ios/discussions/1957.

- Adds proper error propagation to dotLottie loading.
  - Handles manifest loading errors.
  - Handles animation loading errors.
- Adds `LottieLogger` warnings to methods loading from dotLottie.
- Removes unnecessary `animationUrl` property from DotLottieAnimation.
- Unifies asset loading mechanism for dotLottie and standard .json files.

Propagating errors has actually made the error handling shorter and easier to debug.

### DIFF
--- a/Sources/Private/Model/DotLottie/DotLottieUtils.swift
+++ b/Sources/Private/Model/DotLottie/DotLottieUtils.swift
@@ -54,5 +54,4 @@ extension FileManager {
 public enum DotLottieError: Error {
   case invalidFileFormat
   case invalidData
-  case animationNotAvailable
 }
--- a/Sources/Private/Model/DotLottie/DotLottieUtils.swift
+++ b/Sources/Private/Model/DotLottie/DotLottieUtils.swift
@@ -52,7 +52,9 @@ extension FileManager {
 // MARK: - DotLottieError
 
 public enum DotLottieError: Error {
-  case urlLoadingFailed
+  /// URL response has no data.
+  case noDataLoaded
+  /// Asset with this name was not found in the provided bundle.
   case assetNotFound(name: String, bundle: Bundle?)
   /// Animation loading from asset is not supported on macOS 10.10.
   case loadingFromAssetNotSupported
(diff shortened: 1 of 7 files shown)


Sources/Private/Model/DotLottie/DotLottieUtils.swift
```diff
@@ -52,7 +52,10 @@ extension FileManager {
 // MARK: - DotLottieError
 
 public enum DotLottieError: Error {
-  case invalidFileFormat
```
calda, line 55: Since this is public I think we need to preserve these enum cases, otherwise this would be a breaking change. Could we deprecate the existing cases instead of removing them?

Sources/Private/Utility/Extensions/DataExtension.swift
```diff
@@ -14,14 +14,24 @@ import AppKit
 
 extension Data {
 
-  static func jsonData(from assetName: String, in bundle: Bundle) -> Data? {
+  init(assetName: String, in bundle: Bundle) throws {
     #if canImport(UIKit)
-    return NSDataAsset(name: assetName, bundle: bundle)?.data
+    if let asset = NSDataAsset(name: assetName, bundle: bundle) {
+      self = asset.data
+      return
```
calda, line 21: Is this necessary? It seems like it may be redundant

```suggestion
```

calda (approved):
Thanks! Great improvements, appreciate it.

Sources/Private/Model/DotLottie/DotLottieUtils.swift
calda: Haha oops accidentally merged the PR after I finished reviewing it. I posted a follow-up PR to add back these three public enum cases: https://github.com/airbnb/lottie-ios/pull/1970

mkj-is: Agree, it makes sense to keep them until the next minor/major version is released.

PR #1970, opened 2023-02-21
Add back public error values that were removed in #1969

calda:
#1969 removed some cases from the public `DotLottieError` type, which is potentially breaking change for consumers. This PR adds back those previous cases, but with deprecation warnings indicating that they are no longer used.


mkj-is (approved):

mkj-is:
Looks good!

erichoracek (approved):

PR #1978, opened 2023-03-04
Update ZIPFoundation to 0.9.16

### LINKED ISSUE #1945: dotLottie load fail

Lottie 4.1.2

my dotlottie File can play success in https://app.lottiefiles.com/preview?utm_medium=web&utm_source=navigation-preview&fileUrl=

but when I use in my app like this:

DotLottieFile.named("testdot") { result in
                    switch result {
                    case .success(let dotLottieFile):
                        self.tipsAniView = LottieAnimationView(dotLottie: dotLottieFile)
                        self.tipsAniView.loopMode = .loop
                        self.tipsAniView.play()
                    case .failure(let error):
                        print(error)
                    }
                }

error: failure	Foundation.Data.CompressionError	corruptedData
Error when decoding lottie "testdot": corruptedData

You can unzip  dotLottieFile.zip. get dotLottie

## Animation JSON

[dotLottieFile.zip](https://github.com/airbnb/lottie-ios/files/10683228/dotLottieFile.zip)

calda:
This PR updates our ZIPFoundation dependency to `0.9.16`. This fixes #1945, where a dotLottie file would fail to load with a `corruptedData` error due to an issue in the version of ZIPFoundation that we were previously using.

It's still not really ideal that we integrate ZIPFoundation via source, but it's still the best option. I went ahead and added a README.md to our ZIPFoundation folder that includes a bit of discussion plus instructions for how to update this in the future if necessary.


Sources/Private/Model/DotLottie/DotLottieManifest.swift
```diff
@@ -10,21 +10,10 @@ import Foundation
...
-
-  // MARK: Internal
-
   var animations: [DotLottieAnimation]
-  var version: String
-  var author: String
-  var generator: String
+  var version: String?
+  var author: String?
+  var generator: String?
```
calda, line 16: The animation in #1945 was also failing to load because the manifest didn't provide these properties. Seems fine to make them optional, especially since we don't use these anywhere.

Sources/Private/Model/DotLottie/ZipFoundation/README.md
```diff
@@ -0,0 +1,24 @@
+## ZipFoundation
```
calda, line 1: Added some more discussion about our ZIPFoundation dependency, plus instructions for if we have to update this again in the future.

Sources/Private/Model/DotLottie/ZipFoundation/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+This directory includes the source code of the ZIPFoundation library, from the following release:
+https://github.com/weichsel/ZIPFoundation/releases/tag/0.9.16
+
+Lottie is distributed via multiple package managers (SPM, Cocoapods, Carthage, and NPM),
+each with different packaging and compilation requirements. 
+
+Due to limitations of these package managers, we can't depend on / import 
+a separate ZIPFoundation module / library. Instead, we include the source
+directly within the Lottie library and compile everything as a single unit.
```
calda, line 11: 🫠

calda: At least updating to the most recent version was painless

Sources/Private/Model/DotLottie/ZipFoundation/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+From time to time we may need to update to a more recent version of ZIPFoundation.
+When doing this, follow these steps:
+
+ 1. Download the latest release from https://github.com/weichsel/ZIPFoundation
+    and replace the source code in this directory with the updated code.
+    
+ 2. Update the URL at the top of this file to indicate what release is being used.
+ 
+ 3. Change all of the `public` symbols defined in this module to instead be `internal`
```
erichoracek, line 23: Do we have a script that automates this? might be useful to write as a rake task

erichoracek (approved):
LGTM! Just a small suggestion

Sources/Private/Model/DotLottie/ZipFoundation/README.md
calda: We'll have to do this pretty infrequently (ideally never but we'll see), so I don't think we need a script for now. If we end up having to do this more regularly then a script would make sense

RayFor24:
Does version 4.1.3 not yet include this update?

calda:
Could you file an issue? Seems worth fixing

PR #1988, opened 2023-03-08
Fix issue where ** wildcard incorrectly required matching at least one path component

### LINKED ISSUE #1847: setValueProvider not working as before
### Discussed in https://github.com/airbnb/lottie-ios/discussions/1846

<div type='discussions-op-text'>

<sup>Originally posted by **ana-bonnet** December 12, 2022</sup>
Hello - hope you are all good.

We just update the SDK from 3.4.3 to 4.0.1 using cocoa pods. We are using a few simple animations that contains only one color, for example we have a circular loader and we have been updating the color with the code below, have been working great but since we updated the library it doesn't work anymore, the JSON hasn't change, so does any of this flow have been updated in the 4.0.1 version? Do I need to change anything?

`let animationView: LottieAnimationView = LottieAnimationView(name: "loader")`
`let keypath = AnimationKeypath(keys: ["**", "Stroke 1", "**", "Color"])`
`let colorProvider = ColorValueProvider(theme.tintColor.lottieColorValue)`  
`animationView.setValueProvider(colorProvider, keypath: keypath)`

Thanks a lot in advance.</div>

calda:
This PR fixes an issue where the Core Animation rendering engine incorrectly required ** wildcards to match at least one path component, which was inconsistent with the Main Thread rendering engine. Fixes #1847.


Tests/SnapshotConfiguration.swift
```diff
@@ -68,6 +68,10 @@ extension SnapshotConfiguration {
     "Issues/issue_1664": .customValueProviders([
       AnimationKeypath(keypath: "**.base_color.**.Color"): ColorValueProvider(.black),
     ]),
+    
+    "Issues/issue_1847": .customValueProviders([
+      AnimationKeypath(keypath: "**.Stroke 1.**.Color"): ColorValueProvider(.red),
```
calda, line 73: Before this PR, this value provider was unexpectedly only applied when using the Main thread rendering engine but not when using the Core Animation rendering engine.

Sources/Private/CoreAnimation/ValueProviderStore.swift
```diff
@@ -125,13 +125,19 @@ extension AnimationKeypath {
       + keypath.keys.joined(separator: "\\.") // match this keypath, escaping "." characters
       + "$" // match the end of the string
 
-    // ** wildcards match anything
-    //  - "**.Color" matches both "Layer 1.Color" and "Layer 1.Layer 2.Color"
-    regex = regex.replacingOccurrences(of: "**", with: ".+")
-
-    // * wildcards match any individual path component
+    let doubleWildcardMarker = "DOUBLE"
+    let singleWildcardMarker = "SINGLE"
```
erichoracek, line 129: what happens if someone has one of these tokens as a layer name—seems like we could accidentally treat it as a marker?

erichoracek (approved):
LGTM! Just one question

Sources/Private/CoreAnimation/ValueProviderStore.swift
calda: Aaaaah thank you, good thinking. Yeah that would be a problem. I'll change this to something less likely to cause collisions. Maybe a UUID string?

chrischang26:
Hi @calda , we noticed that the issue also happens with gradients.
Instead of not showing any keypaths with 7.4.0 and above, we see duplicated gradient keypaths, hence it's displaying black shadow where we apply custom gradient.

PR #1994, opened 2023-03-10
Fix issue where completion handler could be called unexpectedly when setting up animation

### LINKED ISSUE #1984: Automatic mode not working for some animations with unsupported functionality
This is the issue from #1946 reopened, as I didn't see any way to reopen the existing issue.

We started testing Lottie v4.1.2 to determine if it would be safe to update, and have run into a strange issue with a subset of animations.

When running the animation in mainThread mode, the animation works correctly. When running in the animation in automatic mode, we get the warning "The Core Animation rendering engine doesn't currently support applying trims to filled shapes (only stroked shapes)" which is expected at the moment. However, in automatic mode, the animation does not display at all, even though the code appears to be attempting to fall back to the mainThread rendering engine.

We have seen the fall back to mainThread rendering enginer work correctly for other animations, though. The common thread among animations that don't work is the listed feature around trim paths and filled shapes.

One example animation is attached.

Lottie 4.1.2

## Expected Behavior 
When using the automatic rendering engine mode and the animation has an unsupported feature, it should fall back to the mainThread rendering mode and properly display the animation.

## Actual Behavior
The unsupported feature is properly logged and the code appears to attempt to fall back to the mainThread rendering mode, but the animation never actually plays.

## Animation JSON
[fireworks2018.json.zip](https://github.com/airbnb/lottie-ios/files/10913612/fireworks2018.json.zip)

Here is a small project that replciates this issue:
[TestingLottie.zip](https://github.com/airbnb/lottie-ios/files/10913614/TestingLottie.zip)

calda:
This PR fixes an issue where `LottieAnimationView`'s animation completion handlers could be called unexpectedly when setting up an animation. Fixes #1984.

Specifically, this would happen when using `RenderingEngineOption.automatic` with an animation not supported by the Core Animation rendering engine. We would initialize the Core Animation animation layer, detect it was unsupported, deinitialize the Core Animation layer, and then initialize the Main Thread animation layer. Previously, when this flow deinitialized the Core Animation animation layer it also unexpectedly triggered the animation completion handler.

### DIFF
--- a/Tests/AnimationViewTests.swift
+++ b/Tests/AnimationViewTests.swift
@@ -7,6 +7,14 @@ import XCTest
 @MainActor
 final class AnimationViewTests: XCTestCase {
 
+  override func setUp() async throws {
+    LottieLogger.shared = .printToConsole
+  }
+
+  override func tearDown() {
+    LottieLogger.shared = LottieLogger()
+  }
+
   func testLoadJsonFile() {
     let animationView = LottieAnimationView(
       name: "LottieLogo1",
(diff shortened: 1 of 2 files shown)


Tests/AnimationViewTests.swift
```diff
@@ -74,67 +82,87 @@ final class AnimationViewTests: XCTestCase {
...
-          XCTAssertFalse(
-            animationView.isAnimationPlaying,
-            "Failed case \(test) on engine: \(engine.label)")
-
-          animationCompleteExpectation.fulfill()
-        }
-
-        DispatchQueue.main.asyncAfter(deadline: .now() + 0.05) {
-          animationPlayingExpectation.fulfill()
+    let animationUnsupportedByCoreAnimationRenderingEngine = LottieAnimation.named(
```
calda, line 90: Updated this existing test to include an additional case. This case fails without the change to `LottieAnimationView`.

Tests/AnimationViewTests.swift
```diff
@@ -74,67 +82,87 @@ final class AnimationViewTests: XCTestCase {
...
 
-          // Check that the animation is playing in the correct direction:
-          // After a brief delay we should be closer to the from frame than the to frame
-          let distanceFromStartFrame = abs((values.fromFrame ?? 0) - animationView.realtimeAnimationFrame)
-          let distanceFromEndFrame = abs(values.toFrame - animationView.realtimeAnimationFrame)
-          XCTAssertTrue(
-            distanceFromStartFrame < distanceFromEndFrame,
-            "Failed case \(test) on engine: \(engine.label)")
+    for animation in animations {
+      XCTAssertNotNil(animation)
```
calda, line 101: Everything after this is just reindenting after the additional `for` loop.

erichoracek (approved):

PR #2004, opened 2023-03-24
Added a functionality to get a list of the layers for an animation, s…

Bjorninn1:
…imilar to logHierarchyKeypaths but instead of just logging them they are returned and can for example be written directly to a file. To use: let layers = animationView.getHierarchyKeypaths().

### DIFF
--- a/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
+++ b/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
@@ -190,7 +190,7 @@ final class CoreAnimationLayer: BaseAnimationLayer {
   private let animation: LottieAnimation
   private let valueProviderStore: ValueProviderStore
   private let compatibilityTracker: CompatibilityTracker
-  private let logger: LottieLogger
+  private var logger: LottieLogger
 
   /// The current playback state of the animation that is displayed in this layer
   private var currentPlaybackState: PlaybackState? {
--- a/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
+++ b/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
@@ -448,12 +448,12 @@ extension CoreAnimationLayer: RootAnimationLayer {
   }
 
   func logHierarchyKeypaths() {
-    for keypath in allAnimationKeypaths() {
+    for keypath in allHierarchyKeypaths() {
       logger.info(keypath)
     }
   }
 
-  func allAnimationKeypaths() -> [String] {
+  func allHierarchyKeypaths() -> [String] {
     guard var configuration = pendingAnimationConfiguration?.animationConfiguration ?? currentAnimationConfiguration else {
       logger.info("Cannot log hierarchy keypaths until animation has been set up at least once")
       return []
--- a/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
+++ b/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
@@ -81,7 +81,13 @@ final class CoreAnimationLayer: BaseAnimationLayer {
   struct AnimationConfiguration: Equatable {
     var animationContext: AnimationContext
     var timingConfiguration: CAMediaTimingConfiguration
-    var logHierarchyKeypaths = false
+    var recordHierarchyKeypath: ((String) -> Void)?
+
+    static func ==(_ lhs: AnimationConfiguration, _ rhs: AnimationConfiguration) -> Bool {
+      lhs.animationContext == rhs.animationContext
+        && lhs.timingConfiguration == rhs.timingConfiguration
+        && ((lhs.recordHierarchyKeypath == nil) == (rhs.recordHierarchyKeypath == nil))
+    }
   }
 
   /// A closure that is called after this layer sets up its animation.
@@ -190,7 +196,7 @@ final class CoreAnimationLayer: BaseAnimationLayer {
   private let animation: LottieAnimation
   private let valueProviderStore: ValueProviderStore
   private let compatibilityTracker: CompatibilityTracker
-  private var logger: LottieLogger
+  private let logger: LottieLogger
 
   /// The current playback state of the animation that is displayed in this layer
   private var currentPlaybackState: PlaybackState? {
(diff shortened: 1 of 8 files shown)


Sources/Private/CoreAnimation/CoreAnimationLayer.swift
```diff
@@ -441,6 +441,11 @@ extension CoreAnimationLayer: RootAnimationLayer {
     // Unimplemented / unused
   }
 
+  func getHierarchyKeypaths() -> [String] {
+    logHierarchyKeypaths()
+    return []
```
calda, line 446: Can you implement this for the Core Animation rendering engine as well?

Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift
```diff
@@ -198,6 +198,13 @@ final class MainThreadAnimationLayer: CALayer, RootAnimationLayer {
     animationLayers.forEach { $0.displayWithFrame(frame: currentFrame, forceUpdates: true) }
   }
 
+  func getHierarchyKeypaths() -> [String] {
```
calda, line 201: An API like this would be more idiomatic:

```suggestion
  var allAnimationKeypaths: [String] {
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -710,6 +710,11 @@ open class LottieAnimationView: LottieAnimationViewBase {
     animationLayer?.getOriginalValue(for: keypath, atFrame: atFrame)
   }
 
+  /// Gets all child keypaths.
+  public func getHierarchyKeypaths() -> [String] {
```
calda, line 714: An API like this would be more idiomatic:

```suggestion
  public var allAnimationKeypaths: [String] {
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -710,6 +710,11 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
+    animationLayer != nil ? animationLayer!.getHierarchyKeypaths() : []
```
calda, line 715: No need to check for nil / force unwrap:

```suggestion
    animationLayer?.getHierarchyKeypaths() ?? []
```

Sources/Public/Animation/LottieAnimationView.swift
calda, line 714: What use case do you have in mind for this API?

Bjorninn1: I write the layers to a file. Having the layer names they can then be modified.

Sources/Private/CoreAnimation/CoreAnimationLayer.swift
Bjorninn1: Yes, will look into that.

Sources/Public/Animation/LottieAnimationView.swift
calda: What do you do with that file? Is there anything this lets you do that isn't possible today with `logHierarchyKeypaths()`?

Bjorninn1: It allows making dynamic changes. Using logHierarchyKeypaths() means this would have to be done manually.

Sources/Private/CoreAnimation/CoreAnimationLayer.swift
Bjorninn1: I see how displayIfNeeded() inside logHierarchyKeypaths in CoreAnimationLayer is magically causing the log to happen. Looks to me that customKeyframes is somehow being triggered. It doesn't seem to be straightforward to implement getHierarchyKeypaths for Core Animation rendering.

calda: Here's an approach we use elsewhere to get an array of keypath names just by calling `logHierarchyKeypaths()`: https://github.com/airbnb/lottie-ios/blob/master/Tests/AnimationKeypathTests.swift#L118 This works on both the Core Animation and Main Thread rendering engines.

Bjorninn1: Thanks, looked into it and I thought it would be best to create an extension to RootAnimationLayer to have the same implementation for the function and use the idea from hierarchyKeypaths from the code in your link.
This implementation causes a crash when returning in getHierarchyKeypaths.

added to protocol RootAnimationLayer:
`func getHierarchyKeypaths(getAnimationView: @escaping (_ newLogger: LottieLogger) -> LottieAnimationView) -> [String]
`

created extension in RootAnimationLayer.swift to have default implementation for the function:

```
extension RootAnimationLayer {
    func getHierarchyKeypaths(getAnimationView: @escaping (_ newLogger: LottieLogger) -> LottieAnimationView) -> [String] {
        var printedMessages = [String]()
        let logger = LottieLogger(info: { message in
            printedMessages.append(message())
        })
       let animationView = getAnimationView(logger)
       animationView.logHierarchyKeypaths()
       
      return Array(printedMessages[1...])    //Thread 1: Fatal error: Range requires lowerBound <= upperBound
    }
}

```

In LottieAnimationView.swift:

```
private func getAnimationView(newLogger: LottieLogger) -> LottieAnimationView {
    self.logger = newLogger
    return self
}
    
public var allAnimationKeypaths: [String] {
    animationLayer?.getHierarchyKeypaths(getAnimationView: getAnimationView) ?? []
}
```

calda: If you push a commit with that change I can take a look. It sounds like the problem is that `printedMessages` is empty.

calda: It's possible that `logHierarchyKeypaths()` is asynchronous, so hasn't finished running by the time you try to read `printedMessages`.

Bjorninn1: Yes that's right, the list is empty. I have pushed the changes. Everything is inside allAnimationKeypaths in LottieAnimationView. Don't see that logHierarchyKeypaths is async.

Bjorninn1: The newest commit solves everything, both for core and main thread engines. Really appreciate the support and feedback. Just let me know if something else needs to be changed.

calda (approved):
Thanks! I pushed some commits that make a few improvements. Appreciate the contribution.

PR #2008, opened 2023-03-28
Make maskToBounds configurable

johnny-duo:
Based on [this discussion](https://github.com/airbnb/lottie-ios/discussions/2007), this PR adds support for configuring maskToBounds to fix issues with old animations breaking.

### DIFF
--- a/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
+++ b/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
@@ -19,6 +19,7 @@ final class CoreAnimationLayer: BaseAnimationLayer {
     imageProvider: AnimationImageProvider,
     textProvider: AnimationTextProvider,
     fontProvider: AnimationFontProvider,
+    maskAnimationToBounds: Bool,
     compatibilityTrackerMode: CompatibilityTracker.Mode,
     logger: LottieLogger)
     throws
@@ -31,7 +32,7 @@ final class CoreAnimationLayer: BaseAnimationLayer {
     compatibilityTracker = CompatibilityTracker(mode: compatibilityTrackerMode, logger: logger)
     valueProviderStore = ValueProviderStore(logger: logger)
     super.init()
-    masksToBounds = true
+    masksToBounds = maskAnimationToBounds
     setup()
     try setupChildLayers()
   }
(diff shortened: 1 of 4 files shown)


Sources/Private/CoreAnimation/CoreAnimationLayer.swift
```diff
@@ -19,6 +19,7 @@ final class CoreAnimationLayer: BaseAnimationLayer {
     imageProvider: AnimationImageProvider,
     textProvider: AnimationTextProvider,
     fontProvider: AnimationFontProvider,
+    maskAnimationToBounds: Bool,
```
johnny-duo, line 22: It shouldn't actually be necessary to pass this in since the default value is the same as what we set it to in the init. Not sure what's preferred

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -454,6 +454,13 @@ open class LottieAnimationView: LottieAnimationViewBase {
     }
   }
 
+  /// Sets if sublayers of the animation are clipped to bounds
+    public var maskAnimationToBounds: Bool = true {
+        didSet {
+            animationLayer?.masksToBounds = maskAnimationToBounds
+        }
+    }
+
```
calda, line 463:
```suggestion
  /// Whether or not the animation is masked to the bounds. Defaults to true.
  public var maskAnimationToBounds: Bool = true {
    didSet {
      animationLayer?.masksToBounds = maskAnimationToBounds
    }
  }
```

calda (approved):

Sources/Private/CoreAnimation/CoreAnimationLayer.swift
calda: I would probably prefer a design like:


```swift
let layer = CoreAnimationLayer(...)
layer.masksToBounds = maskAnimationToBounds
```

since there's no need to inject a property into the initializer when we can just configure the same thing directly at the call site. But either is fine.

calda: What you have here may be better anyway so we don't forget to configure this if we ever add another `init` call to either of these layer type

johnny-duo: This makes more sense, I'll change it

johnny-duo:
I'll do a quick check first to see if I need to call anything else in the didSet

Sources/Private/CoreAnimation/CoreAnimationLayer.swift
johnny-duo: Oh oops didn't see your second message

johnny-duo:
Did a quick check that this will fix our issue 👍. Thanks!

johnny-duo:
@calda Looks like a performance test is failing? 
image

Not sure why this is happening since it shouldn't change any behaviour

calda:
Those are flaky sometimes. We can just run it again.

PR #2016, opened 2023-04-01
Fixed 2010: Added trimming whitespaces and newlines condition in dataString.

rohit9934:
Fixed 2010: Added condition to trim the String that contains new lines to avoid returning nil from data with a base64 encoded string.


calda (approved):
Thanks!

PR #2036, opened 2023-04-19
Split shared component between MainThread and CoreAnimation.

chuynadamas:
In this PR, I'm making changes to the Lottie library that will help us to optimize the performance of our animations. Since Lottie [4.1.1](https://github.com/airbnb/lottie-ios/releases/tag/4.1.1), the CoreAnimation renderer has been set as default which means that in the future all the animations will be rendered using it instead of the MainThread renderer.

To make the future migration an easy process, I'm separating the shared components in a dedicated folder named "SharedComponents" that currently are used for both renderers. This will also create a clear boundary between both implementations and decouple them.

As an additional benefit, we are also separating both renderers logic into independent modules that will create the flexibility to import one or both if needed. This will create a binary size benefit, making the Lottie library more lightweight and efficient.


calda:
Hi @chuynadamas, it looks like you're proposing a significant change to the project architecture. Could you please [start a discussion](https://github.com/airbnb/lottie-ios/discussions) about this? I don't currently think it would be a good idea / worth while to split the two rendering engines into separate modules, so we'd need to align on the value of that and the architecture direction before making these sorts of changes.

> the CoreAnimation renderer has been set as default which means that in the future all the animations will be rendered using it instead of the MainThread renderer.

fwiw this isn't quite right -- some parts of the Lottie animation spec (like time remapping) are only supported by the Main Thread rendering engine. We currently have no plans to support full parity between the two engines, since that would be extremely difficult or maybe impossible. The two engines will always need to live side by side using something like `RenderingEngineOption.autmatic` (use the Core Animation by default _if possible_ but _otherwise_ use the Main Thread rendering engine) to provide the best experience to consumers.

chuynadamas:
Hey @calda, thanks for the quick response, more than happy to start a discussion about this :)

PR #2041, opened 2023-04-24
Fix issue where Trim could be applied to too many shape items

### LINKED ISSUE #2030: Issues with coreAnimation rendering
Some of our animations fails to render correctly with codeAnimation rendering engine. Issues are fixed when using mainThread. Animation can be requested from me by email. See my profile for email info.

calda:
This PR fixes an issue where a `Trim` could be be applied to too many shape items.

In a group like `[Shape 1, Trim, Shape 2]`, the trim should only be applied to Shape 1 and not Shape 2. Previously, it was applied to both.

Fixes #2030.

| Before | After |
| ---- | ---- |
| Simulator Screenshot - iPhone 8 - 2023-04-24 at 16 37 58 | Simulator Screenshot - iPhone 8 - 2023-04-24 at 16 35 34 |


Tests/SnapshotTests.swift
```diff
@@ -121,7 +121,8 @@ class SnapshotTests: XCTestCase {
         assertSnapshot(
           matching: animationView,
           as: .imageOfPresentationLayer(
-            precision: SnapshotConfiguration.forSample(named: sampleAnimationName).precision),
+            precision: SnapshotConfiguration.forSample(named: sampleAnimationName).precision,
+            perceptualPrecision: 0.985),
```
calda, line 125: I was getting diff failures locally on master, probably because I'm using a different Xcode version than CI. I enabled `perceptualPrecision` (a new-ish feature in the snapshotting library) which should prevent that sort of issue from now on.

erichoracek (approved):

PR #2048, opened 2023-04-26
Use `XCTExpectFailure(strict: false)` for tests that are flaky in CI

calda:
This PR adopts `XCTExpectFailure(strict: false)` for our tests that are historically flaky in CI.

Test failures here will no longer cause CI to fail, but will still show an alert in Xcode and still show up in the logs:

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,6 +56,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/setup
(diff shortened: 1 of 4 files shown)


calda:
The changes in this PR fail to compile in Xcode 14.2 but do compile in Xcode 14.3, so I took this opportunity to update CI to run on macOS 13 using Xcode 14.3 by default.

erichoracek (approved):

PR #2053, opened 2023-04-27
Fix issue where trim offset could be calculated incorrectly

### LINKED ISSUE #1979: Animation behavior changed in new render engine, breaking loop

Tested on:
 - Lottie 3.3.0 (working as expected)
 - Lottie 3.5.0 (working as expected)
 - Lottie 4.0.0 (automatic rendering: not working, main thread: working as expected)
 - Lottie 4.1.3 (automatic rendering: not working, main thread: working as expected)

## Expected Behavior 
(matches lottiefiles.com preview)

https://user-images.githubusercontent.com/6655924/223274773-64972eb6-0d8b-479c-b861-85f7a9192a8f.mp4

## Actual Behavior
https://user-images.githubusercontent.com/6655924/223275193-b07038d0-6d42-4f8e-a9b8-b637d7812bfd.mp4

## Animation JSON
[spinnerCheckmark.json.zip](https://github.com/airbnb/lottie-ios/files/10903678/spinnerCheckmark.json.zip)

calda:
This PR fixes an issue where trim offsets could be calculated incorrectly, causing the animation to be rendered incorrectly. Fixes #1979.

The previous implementation from #1612 predated `Keyframes.combined` so was totally custom, and must have had a bug somewhere. Consolidating this to use the same interpolation code we use elsewhere also fixes some other minor issues in the regression suite where some stroke trims were rendered in a very-slightly different way between the two engines.

| Before | After |
| ---- | ---- |
|  |  |


Sources/Private/CoreAnimation/Animations/ShapeAnimation.swift
```diff
@@ -140,22 +140,19 @@ extension Trim {
...
-        strokeKeyframes: interpolatedStrokeEnd,
-        offsetKeyframes: interpolatedStrokeOffset))
+    // Apply the offset to the start / end values at each frame
+    let offsetStrokeKeyframes = Keyframes.combined(
+      strokeStart,
+      strokeEnd,
+      offset,
+      makeCombinedResult: { start, end, offset -> (start: LottieVector1D, end: LottieVector1D) in
+        let offsetStart = start.cgFloatValue + (offset.cgFloatValue / 360 * 100)
+        let offsetEnd = end.cgFloatValue + (offset.cgFloatValue / 360 * 100)
```
erichoracek, line 150: nit: maybe discuss why we have these values we're dividing and multiplying by

erichoracek (approved):
-150 net LoC and a bugfix, nice!

gsbernstein:
Thank you!

PR #2067, opened 2023-05-16
Fix issue where layers with negative X scale values could be pixelated

### LINKED ISSUE #2066: Pixellated Shape Layers
Hello! 

Parts of our animation is showing up pixelated on iOS.  It plays fine in Web browser.   Most are direct shape layers, not precomped.   Some objects are scaled past 100%, but nothing past 370%.  Scaling has been fine in the past, so this seems to be the bug?

This is a similar behavior to this past issue 1022: https://github.com/airbnb/lottie-ios/issues/1022 

Lottie 4.1.2

## Expected Behavior 
Expected Behavior

## Actual Behavior
Actual Behavior

## Animation JSON

JSON and AE files in here:
[D12_Bug Report.zip](https://github.com/airbnb/lottie-ios/files/11489333/D12_Bug.Report.zip)

calda:
This PR fixes an issue where layers with negative X scale values could be pixelated. Fixes #2066.

Core Animation doesn't support negative X scale values when animating `transform.scale.x` and `transform.scale.y` separately (https://openradar.appspot.com/FB9862872). 

Previously we worked around this by applying a 180º rotation around the Y axis when `scale.x` is negative, which has the same effect. This seems to affect the display quality of layers under some circumstances (no clue why), though. This also didn't work in snapshot tests for some reason, which was inconvenient.

Instead, we now handle negative X scale values by combining all of the transform components into a single set of keyframes, which we apply to `layer.transform`. This works perfectly and doesn't have any of the issues we encountered with the other approach.

| Before | After |
| ---- | ---- |
| Simulator Screenshot - iPhone 14 Pro - 2023-05-16 at 13 19 29 | Simulator Screenshot - iPhone 14 Pro - 2023-05-16 at 13 20 25 |


Sources/Private/CoreAnimation/Extensions/Keyframes+combined.swift
```diff
@@ -133,6 +133,46 @@ enum Keyframes {
       })
   }
 
+  /// Combines the given keyframe groups of `Keyframe<T>`s into a single keyframe group of of `Keyframe<[T]>`s
+  ///  - If all of the `KeyframeGroup`s have the exact same animation timing, the keyframes are merged
+  ///  - Otherwise, the keyframes are manually interpolated at each frame in the animation
+  static func combined<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, CombinedResult>(
```
calda, line 139: are variadic generics coming in Swift 5.9?????????

PR #2072, opened 2023-06-06
Fix Swift 5.9 warnings

keith:
There is now a new warning produced by these functions since the type and the function both declare a generic with the same name.

```
lottie-ios/Sources/Private/CoreAnimation/Extensions/Keyframes+combined.swift:265:32: warning: generic parameter 'T' shadows generic parameter from outer scope with the same name; this is an error in Swift 6
  func hasSameTimingParameters<T>(as other: KeyframeGroup<T>) -> Bool {
lottie-ios/Sources/Private/Model/Keyframes/KeyframeGroup.swift:18:27: note: 'T' previously declared here
final class KeyframeGroup<T> {
```


calda (approved):
Thanks for the fix!

PR #2073, opened 2023-06-06
Add a public interface to display Lottie Animations directly from a CALayer

eromanc:
https://origami.design/ wants to add support for Lottie Animations. However Origami works at CALayer level and because of that it cannot add directly a NSView/UIView. Internally Lottie also works with CALayers. This commit creates then a CALayer interface that mimics basically the same behavior/functionality of the LottieAnimationView and then LottieAnimationView also uses that layer internally.

### DIFF
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -79,23 +79,23 @@ public class LottieAnimationLayer: CALayer {
   }
 
   /// Called by CoreAnimation to create a shadow copy of this layer
-   /// More details: https://developer.apple.com/documentation/quartzcore/calayer/1410842-init
-   override init(layer: Any) {
-     guard let typedLayer = layer as? Self else {
-       fatalError("\(Self.self).init(layer:) incorrectly called with \(type(of: layer))")
-     }
+  /// More details: https://developer.apple.com/documentation/quartzcore/calayer/1410842-init
+  override init(layer: Any) {
+    guard let typedLayer = layer as? Self else {
+      fatalError("\(Self.self).init(layer:) incorrectly called with \(type(of: layer))")
+    }
 
-     animation = typedLayer.animation
-     imageProvider = typedLayer.imageProvider
-     textProvider = typedLayer.textProvider
-     fontProvider = typedLayer.fontProvider
-     logger = typedLayer.logger
-     screenScale = typedLayer.screenScale
-     configuration = typedLayer.configuration
-     super.init(layer: typedLayer)
-   }
+    animation = typedLayer.animation
+    imageProvider = typedLayer.imageProvider
+    textProvider = typedLayer.textProvider
+    fontProvider = typedLayer.fontProvider
+    logger = typedLayer.logger
+    screenScale = typedLayer.screenScale
+    configuration = typedLayer.configuration
+    super.init(layer: typedLayer)
+  }
 
-  required init?(coder: NSCoder) {
+  required init?(coder _: NSCoder) {
     fatalError("init(coder:) has not been implemented")
   }
 
@@ -536,10 +536,6 @@ public class LottieAnimationLayer: CALayer {
     get { animationLayer?.animationView }
   }
 
-  var hasAnimationContext: Bool {
-    animationContext != nil
-  }
-
   /// Sets the lottie file backing the animation layer. Setting this will clear the
   /// layer's contents, completion blocks and current state. The new animation will
   /// be loaded up and set to the beginning of its timeline.
@@ -774,6 +770,15 @@ public class LottieAnimationLayer: CALayer {
 
   var animationLayer: RootAnimationLayer? = nil
 
+  /// Context describing the animation that is currently playing in this `LottieAnimationView`
+  ///  - When non-nil, an animation is currently playing in this layer. Otherwise,
+  ///    the layer is paused on a specific frame.
+  fileprivate(set) var animationContext: AnimationContext?
+
+  var hasAnimationContext: Bool {
+    animationContext != nil
+  }
+
   /// Set animation name from Interface Builder
   var animationName: String? {
     didSet {
@@ -842,10 +847,6 @@ public class LottieAnimationLayer: CALayer {
 
   // MARK: Fileprivate
 
-  /// Context describing the animation that is currently playing in this `LottieAnimationView`
-  ///  - When non-nil, an animation is currently playing in this layer. Otherwise,
-  ///    the layer is paused on a specific frame.
-  fileprivate(set) var animationContext: AnimationContext?
   fileprivate var _activeAnimationName: String = LottieAnimationLayer.animationName
   fileprivate var animationID = 0
 
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -6,96 +6,16 @@
 import Foundation
 import QuartzCore
 
-// MARK: - LottieBackgroundBehavior
-
-/// Describes the behavior of an AnimationView when the app is moved to the background.
-public enum LottieBackgroundBehavior {
-  /// Stop the animation and reset it to the beginning of its current play time. The completion block is called.
-  case stop
-
-  /// Pause the animation in its current state. The completion block is called.
-  case pause
-
-  /// Pause the animation and restart it when the application moves to the foreground.
-  /// The completion block is stored and called when the animation completes.
-  ///  - This is the default when using the Main Thread rendering engine.
-  case pauseAndRestore
-
-  /// Stops the animation and sets it to the end of its current play time. The completion block is called.
-  case forceFinish
-
-  /// The animation continues playing in the background.
-  ///  - This is the default when using the Core Animation rendering engine.
-  ///    Playing an animation using the Core Animation engine doesn't come with any CPU overhead,
-  ///    so using `.continuePlaying` avoids the need to stop and then resume the animation
-  ///    (which does come with some CPU overhead).
-  ///  - This mode should not be used with the Main Thread rendering engine.
-  case continuePlaying
-
-  // MARK: Public
-
-  /// The default background behavior, based on the rendering engine being used to play the animation.
-  ///  - Playing an animation using the Main Thread rendering engine comes with CPU overhead,
-  ///    so the animation should be paused or stopped when the `LottieAnimationView` is not visible.
-  ///  - Playing an animation using the Core Animation rendering engine does not come with any
-  ///    CPU overhead, so these animations do not need to be paused in the background.
-  public static func `default`(for renderingEngine: RenderingEngine) -> LottieBackgroundBehavior {
-    switch renderingEngine {
-    case .mainThread:
-      return .pauseAndRestore
-    case .coreAnimation:
-      return .continuePlaying
-    }
-  }
-}
-
-// MARK: - LottieLoopMode
-
-/// Defines animation loop behavior
-public enum LottieLoopMode {
-  /// Animation is played once then stops.
-  case playOnce
-  /// Animation will loop from beginning to end until stopped.
-  case loop
-  /// Animation will play forward, then backwards and loop until stopped.
-  case autoReverse
-  /// Animation will loop from beginning to end up to defined amount of times.
-  case `repeat`(Float)
-  /// Animation will play forward, then backwards a defined amount of times.
-  case repeatBackwards(Float)
-}
-
-// MARK: Equatable
-
-extension LottieLoopMode: Equatable {
-  public static func == (lhs: LottieLoopMode, rhs: LottieLoopMode) -> Bool {
-    switch (lhs, rhs) {
-    case (.repeat(let lhsAmount), .repeat(let rhsAmount)),
-         (.repeatBackwards(let lhsAmount), .repeatBackwards(let rhsAmount)):
-      return lhsAmount == rhsAmount
-    case (.playOnce, .playOnce),
-         (.loop, .loop),
-         (.autoReverse, .autoReverse):
-      return true
-    default:
-      return false
-    }
-  }
-}
-
 // MARK: - LottieAnimationLayer
 
-/// This class will serve as an intermediate layer between a CoreAnimationLayer
-/// and a LottieAnimationView. The motivation behind creating this class is to
-/// have all the functionality of a LottieAnimationView on a CALayer-backed object.
+/// A CALayer subclass for rendering Lottie animations.
+/// All functionality is also available in a UIView as `LottieAnimationView`.
 
 public class LottieAnimationLayer: CALayer {
 
   // MARK: Lifecycle
 
-  // MARK: - Public (Initializers)
-
-  /// Initializes an AnimationView with an animation.
+  /// Initializes a LottieAnimationLayer with an animation.
   public init(
     animation: LottieAnimation?,
     imageProvider: AnimationImageProvider? = nil,
@@ -118,7 +38,7 @@ public class LottieAnimationLayer: CALayer {
     }
   }
 
-  /// Initializes an AnimationView with a .lottie file.
+  /// Initializes an LottieAnimationLayer with a .lottie file.
   public init(
     dotLottie: DotLottieFile?,
     animationId: String? = nil,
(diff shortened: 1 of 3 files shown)


Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -0,0 +1,1292 @@
...
+    }
+  }
+}
+
+// MARK: - LottieAnimationLayer
+
+/// This class will serve as an intermediate layer between a CoreAnimationLayer
+/// and a LottieAnimationView. The motivation behind creating this class is to
+/// have all the functionality of a LottieAnimationView on a CALayer-backed object.
+
```
calda, line 91:
```suggestion
/// A CALayer subclass for rendering Lottie animations.
/// All functionality is also available in a UIView as `LottieAnimationView`.
```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -0,0 +1,1292 @@
...
+public class LottieAnimationLayer: CALayer {
+
+  // MARK: Lifecycle
+
+  // MARK: - Public (Initializers)
+
```
calda, line 97: The Airbnb Swift Style Guide suggests not using `MARK`s other than Lifecycle / Public / Internal / Private etc

```suggestion
```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -0,0 +1,1292 @@
...
+    self.screenScale = 1
+    self.logger = logger
+    super.init()
+    makeAnimationLayer(usingEngine: configuration.renderingEngine)
+    if let animation = animation {
+      frame = animation.bounds
+    }
+  }
+
+  /// Initializes an AnimationView with a .lottie file.
```
calda, line 121:
```suggestion
  /// Initializes a LottieAnimationLayer with a .lottie file.
```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -0,0 +1,1292 @@
...
+    imageProvider = BundleImageProvider(bundle: Bundle.main, searchPath: nil)
+    textProvider = DefaultTextProvider()
+    fontProvider = DefaultFontProvider()
+    self.configuration = configuration
+    self.screenScale = 1
+    self.logger = logger
+    super.init()
+  }
+
+  required public override init(layer: Any) {
```
calda, line 161: This is for creating shadow `CALayer`s so should follow the patterns used elsewhere https://github.com/airbnb/lottie-ios/blob/b68c70572610afd14f013b5bd6473ac6f3fdaca2/Sources/Private/CoreAnimation/Layers/BaseCompositionLayer.swift#L29

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -126,56 +46,64 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
+    self.lottieAnimationLayer.screenScale = self.screenScale
+
+    self.lottieAnimationLayer.animationLoaded = {[weak self] _, animation in
+      guard let self = self else { return }
+      self.animationLoaded?(self, animation)
+      self.invalidateIntrinsicContentSize()
+      self.setNeedsLayout()
+    }
+
+    self.lottieAnimationLayer.animationLayerWillLoad = {[weak self] lottieAnimationLayer, _ in
```
calda, line 101:
```suggestion
    self.lottieAnimationLayer.animationLayerWillLoad = { [weak self] lottieAnimationLayer, _ in
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -126,56 +46,64 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
     commonInit()
+    self.viewLayer?.addSublayer(self.lottieAnimationLayer)
+    setNeedsLayout()
+  }
+
+  override func commonInit() {
+    super.commonInit()
+    self.lottieAnimationLayer.screenScale = self.screenScale
+
+    self.lottieAnimationLayer.animationLoaded = {[weak self] _, animation in
```
calda, line 94:
```suggestion
    self.lottieAnimationLayer.animationLoaded = { [weak self] _, animation in
```

eromanc:
Tested Running the examples on iOS and macOS also all tests passed

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -8,83 +8,6 @@
...
-      return true
-    default:
-      return false
-    }
-  }
-}
-
 // MARK: - LottieAnimationView
 
 @IBDesignable
```
calda, line 13:
```suggestion
/// A UIView subclass for rendering Lottie animations.
/// All functionality is also available in a CALayer as `LottieAnimationLayer`.
@IBDesignable
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -103,18 +26,15 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
+    self.lottieAnimationLayer = LottieAnimationLayer(animation: animation, imageProvider: imageProvider, textProvider: textProvider, fontProvider: fontProvider, configuration: configuration, logger: logger)
     self.logger = logger
     super.init(frame: .zero)
     commonInit()
-    makeAnimationLayer(usingEngine: configuration.renderingEngine)
     if let animation = animation {
       frame = animation.bounds
     }
+    self.viewLayer?.addSublayer(self.lottieAnimationLayer)
+    setNeedsLayout()
```
calda, line 37: Should these two items go in `commonInit`? It seems like we do this in every `init`

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -312,36 +185,27 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
+    self.lottieAnimationLayer.play(marker: marker, loopMode: loopMode, completion: completion)
   }
 
   /// Stops the animation and resets the view to its start frame.
   ///
   /// The completion closure will be called with `false`
   open func stop() {
-    removeCurrentAnimation()
-    currentFrame = 0
+    self.lottieAnimationLayer.stop()
```
calda, line 195: Airbnb's Swift Style Guide says to use implicit self where possible, so all of these should be:

```suggestion
    lottieAnimationLayer.stop()
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -312,36 +185,27 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
-    removeCurrentAnimation()
+    self.lottieAnimationLayer.pause()
   }
 
   // MARK: Public
 
-  /// The configuration that this `LottieAnimationView` uses when playing its animation
-  public let configuration: LottieConfiguration
+  // The backing CALayer for this animation view.
+  public var lottieAnimationLayer: LottieAnimationLayer;
```
calda, line 208: no need for a semicolon :)

```suggestion
  public var lottieAnimationLayer: LottieAnimationLayer
```

calda, line 208: Can this be `private` and immutable? 

```suggestion
  private let lottieAnimationLayer: LottieAnimationLayer
```

calda: I'd prefer to not increase the public API surface area unless really necessary

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -358,37 +222,22 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
-      makeAnimationLayer(usingEngine: configuration.renderingEngine)
-
-      if let animation = animation {
-        animationLoaded?(self, animation)
-      }
+    get {
+      return lottieAnimationLayer.animation
+    }
+    set (newValue) {
+      lottieAnimationLayer.animation = newValue
```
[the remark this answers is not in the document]
calda: When forwarding properties to an underlying type I generally prefer to format the `get` and `set` on a single line like:

```swift
get { lottieAnimationLayer.animation }
set { lottieAnimationLayer.animation = newValue }
```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -0,0 +1,1292 @@
...
+    animationLayer?.removeAnimation(forKey: activeAnimationName)
+    updateAnimationFrame(pauseFrame)
+    animationContext = nil
+  }
+
+  /// Context describing the animation that is currently playing in this `LottieAnimationView`
+  ///  - When non-nil, an animation is currently playing in this view. Otherwise,
+  ///    the view is paused on a specific frame.
+  fileprivate var animationContext: AnimationContext?
+  public var hasAnimationContext: Bool {
```
calda, line 937: Can this be `internal`? This should also have a doc comment.

```suggestion
  var hasAnimationContext: Bool {
```

calda: If this is just used to replace the `animationContext == nil` check in `LottieAnimationView`, could we just change `animationContext` to be internal with `fileprivate(set)`?

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -955,11 +756,11 @@ open class LottieAnimationView: LottieAnimationViewBase {
         position.y = bounds.maxY - animation.bounds.midY
         xform = CATransform3DIdentity
 
-      #if os(iOS) || os(tvOS)
+#if os(iOS) || os(tvOS)
```
calda, line 759:
```suggestion
      #if os(iOS) || os(tvOS)
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -955,11 +756,11 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
       @unknown default:
         logger.assertionFailure("unsupported contentMode: \(contentMode.rawValue)")
         xform = CATransform3DIdentity
-      #endif
+#endif
```
calda, line 763:
```suggestion
      #endif
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -993,11 +794,7 @@ open class LottieAnimationView: LottieAnimationViewBase {
 
       animationLayer.position = position
       animationLayer.transform = xform
-      #if os(OSX)
-      animationLayer.anchorPoint = layer?.anchorPoint ?? CGPoint.zero
-      #else
-      animationLayer.anchorPoint = layer.anchorPoint
-      #endif
+      animationLayer.anchorPoint = self.lottieAnimationLayer.anchorPoint
```
calda, line 797: Good simplification

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -1076,438 +857,27 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
-      }
-
-    case .coreAnimation, nil:
-      // The Core Animation engine automatically batches animation setup to happen
-      // in `CALayer.display()`, which won't be called until the layer is on-screen,
-      // so we don't need to defer animation setup at this layer.
-      break
+    let wasWaitingToPlayAnimation = waitingToPlayAnimation;
+    if waitingToPlayAnimation {
+      waitingToPlayAnimation = false;
```
calda, line 875:
```suggestion
      waitingToPlayAnimation = false
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -1021,37 +818,21 @@ open class LottieAnimationView: LottieAnimationViewBase {
 
     if shouldForceUpdates {
       animationLayer.forceDisplayUpdate()
+      self.lottieAnimationLayer.forceDisplayUpdate();
```
calda, line 821:
```suggestion
      self.lottieAnimationLayer.forceDisplayUpdate()
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -1076,438 +857,27 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
-        waitingToPlayAnimation = true
-        return
-      }
-
-    case .coreAnimation, nil:
-      // The Core Animation engine automatically batches animation setup to happen
-      // in `CALayer.display()`, which won't be called until the layer is on-screen,
-      // so we don't need to defer animation setup at this layer.
-      break
+    let wasWaitingToPlayAnimation = waitingToPlayAnimation;
```
calda, line 873:
```suggestion
    let wasWaitingToPlayAnimation = waitingToPlayAnimation
```

Sources/Public/Animation/LottieAnimationLayer.swift
calda: `CoreAnimationLayer` and `LottieAnimationView` are implementation details so don't need to be mentioned in the doc comment for this public type

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -0,0 +1,1292 @@
...
+      closure: completion)
+    removeCurrentAnimationIfNecessary()
+    addNewAnimationForContext(context)
+  }
+
+  /// Plays the animation from a progress (0-1) to a progress (0-1).
+  ///
+  /// - Parameter fromProgress: The start progress of the animation. If `nil` the animation will start at the current progress.
+  /// - Parameter toProgress: The end progress of the animation.
+  /// - Parameter loopMode: The loop behavior of the animation. If `nil` the view's `loopMode` property will be used.
```
calda, line 204:
```suggestion
  /// - Parameter loopMode: The loop behavior of the animation. If `nil` the layer's `loopMode` property will be used.
```

calda: We should consider replacing "view" with "layer" in all of the doc comments

amerelfallah (approved):

Sources/Public/Animation/LottieAnimationLayer.swift
eromanc: nice. good to know

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -0,0 +1,1292 @@
...
+    didSet {
+      if let animation = animation {
+        animationLoaded?(self, animation)
+      }
+    }
+  }
+
+  /// A closure called when the animation layer will be loaded.
+  /// Will inform the receiver the type of rendering engine that is used for the layer.
+  public var animationLayerWillLoad:((_ animationLayer: LottieAnimationLayer, _ renderingEngine: RenderingEngineOption) -> Void)?
```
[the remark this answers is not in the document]
eromanc: `makeAnimationLayer` used to call `invalidateIntrinsicContentSize` on the View, but since now this is on the layer I needed to pass the information that the layer was created so it could call `invalidateIntrinsicContentSize`

I noticed that on the automatic renderer engine, the creation of the layer can be called more than once, therefore the `invalidateIntrinsicContentSize` was a very important piece that I needed to communicate back to the view.

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -8,83 +8,6 @@
 import Foundation
 import QuartzCore
 
-// MARK: - LottieBackgroundBehavior
-
-/// Describes the behavior of an AnimationView when the app is moved to the background.
-public enum LottieBackgroundBehavior {
```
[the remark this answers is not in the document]
eromanc: sure

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -0,0 +1,1210 @@
...
+//
+
+import Foundation
+import QuartzCore
+
+// MARK: - LottieAnimationLayer
+
+/// A CALayer subclass for rendering Lottie animations.
+/// All functionality is also available in a UIView as `LottieAnimationView`.
+
```
calda, line 13: There shouldn't be a blank line between doc comments and their declaration -- I think this prevents the doc comment from being properly associated with the declaration

```suggestion
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -87,6 +87,9 @@ extension LottieLoopMode: Equatable {
 
 // MARK: - LottieAnimationView
 
+/// A UIView subclass for rendering Lottie animations.
+/// All functionality is also available in a CALayer as `LottieAnimationLayer`.
+
```
calda, line 92: Same here

```suggestion
```

PR #2074, opened 2023-06-09
Added additional support for multiple animations in DotLottie

eharrison:
As an effort to enhance the developer experience with DotLottie, I have added  the following functions to the animation view and layer, for convenience, I have also added the following method to load an animation using it's index, besides the existing function to load an animation with it's ID.
- loadAnimation(atIndex: Int, from dotLottieFile: DotLottieFile)

Made DotLottieConfiguration public and added the following functions inside DotLottieFile as public, to allow external customisation:

- animation(at index: Int) -> DotLottieFile.Animation?

Also made animations array public in DotLottieFile


Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -648,15 +648,54 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
-    animation = dotLottieAnimation.animation
+  /// Exclusive for DotLottie animations. This will be ignored if playing json animation.
+  /// Plays the next animation in the array.
+  /// If there are no next animations, will loop back to the first animation in the array.
+  /// If there is only one animation, will reset to the beginning of the current animation.
+  /// Setting this will clear the view's contents, completion blocks and current state.
+  /// The new animation will be loaded up and set to the beginning of its timeline.
+  /// The loopMode, animationSpeed and imageProvider will be set according
+  /// to lottie file settings
+  public func nextAnimation() {
```
calda, line 679: I'm not sure that `nextAnimation()` and `previousAnimation()` need to be public APIs on `LottieAnimationView`. Before adding new public APIs I think we need to demonstrate a clear bar of:
 1. usefulness / how often this is needed
 2. difficulty or impracticality of implementing the functionality yourself in client code

Currently I'm thinking:
 1. I think this is totally feasible to implement in client code that uses Lottie, assuming we expose the right information via the `DotLottieFile` info
 2. while this may be useful occasionally, I'm not sure it's an operation that would be used enough to warrant being included as public API

What would you think about removing these methods, in favor of creating public APIs on `DotLottieFile` that let you implement this yourself? (e.g. public APIs for getting the list of animations in the file).

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -648,15 +648,54 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
+  /// view's contents, completion blocks and current state. The new animation will
+  /// be loaded up and set to the beginning of its timeline.
+  /// The loopMode, animationSpeed and imageProvider will be set according
+  /// to lottie file settings
+  /// - Parameters:
+  ///   - animationIndex: Internal animation index to play. Optional
+  ///   Defaults to play first animation in file.
+  ///   - dotLottieFile: Lottie file to play
+  public func loadAnimation(
+    animationIndex: Int,
```
calda, line 664:
```suggestion
    atIndex: Int,
```

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -648,15 +648,54 @@ open class LottieAnimationView: LottieAnimationViewBase {
...
+  /// Sets the lottie file backing the animation view. Setting this will clear the
+  /// view's contents, completion blocks and current state. The new animation will
+  /// be loaded up and set to the beginning of its timeline.
+  /// The loopMode, animationSpeed and imageProvider will be set according
+  /// to lottie file settings
+  /// - Parameters:
+  ///   - animationIndex: Internal animation index to play. Optional
+  ///   Defaults to play first animation in file.
+  ///   - dotLottieFile: Lottie file to play
+  public func loadAnimation(
```
calda, line 663: This new method seems like a reasonable addition to me. As of #2073 you'll want to implement this method on `LottieAnimationLayer` instead of `LottieAnimationView`, and then add a method on `LottieAnimationView` that calls the method on the underlying CALayer.

Sources/Private/Model/DotLottie/DotLottieAnimationMode.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  DotLottieAnimationMode.swift
+//  Lottie
+//
+//  Created by Evandro Hoffmann on 09/06/23.
+//
+
+import Foundation
+
+enum DotLottieAnimationMode: String, Codable {
```
calda, line 10: Can we just put this in `DotLottieAnimation.swift` instead of a putting it in a new file?

Sources/Private/Model/DotLottie/DotLottieAnimation.swift
```diff
@@ -20,12 +20,15 @@ struct DotLottieAnimation: Codable {
...
+  var autoplay: Bool? = false
+
   /// mode - "bounce" | "normal"
-  var mode: String? = "normal"
+  var mode: DotLottieAnimationMode? = .normal
 
   /// Loop mode for animation
   var loopMode: LottieLoopMode {
-    mode == "bounce" ? .autoReverse : ((loop ?? false) ? .loop : .playOnce)
+    mode == .bounce ? .autoReverse : ((loop ?? false) ? .loop : .playOnce)
```
calda, line 31: Since we have an enum for this now, could we `switch mode` instead of using a ternary?

Sources/Private/Model/DotLottie/DotLottieAnimation.swift
```diff
@@ -20,12 +20,15 @@ struct DotLottieAnimation: Codable {
   /// 1 or -1
   var direction: Int? = 1
 
+  /// Auto plays animation
+  var autoplay: Bool? = false
```
calda, line 24: Do you know where I can find documentation about the json fields used in dotLottie animation manifests? That would be helpful

Sources/Public/Animation/LottieAnimationView.swift
eharrison: I agree with you, they don't make sense to most people. I'll change the PR to make necessary functions public and remove those methods

Sources/Public/Animation/LottieAnimationView.swift
eharrison: Great, I have moved the function

Sources/Private/Model/DotLottie/DotLottieAnimation.swift
```diff
@@ -21,11 +23,16 @@ struct DotLottieAnimation: Codable {
...
-  var mode: String? = "normal"
+  var mode: DotLottieAnimationMode? = .normal
 
   /// Loop mode for animation
   var loopMode: LottieLoopMode {
-    mode == "bounce" ? .autoReverse : ((loop ?? false) ? .loop : .playOnce)
+    switch mode {
+    case .bounce:
+      return .autoReverse
+    default:
```
calda, line 33: Can we switch exhaustively instead:

```suggestion
    case .normal, nil:
```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -787,6 +796,17 @@ public class LottieAnimationLayer: CALayer {
     }
   }
 
+  func loadAnimation(_ dotLottieAnimation: DotLottieFile.Animation) {
```
calda, line 799: Can this be private?
```suggestion
  private func loadAnimation(_ dotLottieAnimation: DotLottieFile.Animation) {
```

Sources/Public/DotLottie/DotLottieConfiguration.swift
```diff
@@ -0,0 +1,15 @@
+//
+// DotLottieSettings.swift
+// Lottie
+//
+// Created by Evandro Hoffmann on 19/10/22.
+//
+
+import Foundation
+
+public struct DotLottieConfiguration {
```
calda, line 10: Now that this is public, can you document the type and its properties?

Sources/Public/DotLottie/DotLottieConfiguration.swift
```diff
@@ -0,0 +1,15 @@
+//
+// DotLottieSettings.swift
+// Lottie
+//
+// Created by Evandro Hoffmann on 19/10/22.
+//
+
+import Foundation
```
calda, line 8: Is this import necessary?

```suggestion
```

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -25,14 +25,61 @@ public final class DotLottieFile {
...
-  struct Animation {
-    let animation: LottieAnimation
-    let configuration: DotLottieConfiguration
+  public struct Animation {
+    public let animation: LottieAnimation
+    public let configuration: DotLottieConfiguration
+  }
+
+  /// The `LottieAnimation` and `DotLottieConfiguration` for the given animation ID in this file
+  public func animation(for id: String? = nil) -> DotLottieFile.Animation? {
```
calda, line 37: In the `public` function, could we change `for id: String?` to `for id: String`? 

If we need a helper for handling `nil` IDs, can we keep that private?

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -25,14 +25,61 @@ public final class DotLottieFile {
...
+
+  /// The `LottieAnimation` and `DotLottieConfiguration` for the given animation index in this file
+  public func animation(at index: Int) -> DotLottieFile.Animation? {
+    guard index < animations.count else { return nil }
+    return animations[index]
+  }
+
+  /// Returns the next animation in the array. Returns the first animation in case it's the last index
+  /// The next `LottieAnimation` and `DotLottieConfiguration`
+  public func nextAnimation(after animationId: String) -> DotLottieFile.Animation? {
```
calda, line 53: I wonder if it would make sense to just expose the `[Animation]` array publicly, rather than requiring consumers to use this sort of iterator API. If `animations: [Animation]` was public, then I don't think we'd need these `nextAnimation` and `previousAnimation` methods either. Thoughts?

calda (approved):
Looks great, thanks @eharrison! You'll also need to run the linter (`bundle exec rake format:swift`) so CI will pass, then I can merge this

eharrison:
Amazing! thank you, Cal!

PR #2077, opened 2023-06-15
Add LottieView to support SwiftUI

vdotup:


calda:
Hi @vdotup. Thank you for this pull request.

For us to add a Lottie SwiftUI view to the library, I think it would need to more fully-featured and support most / all of the functionality available in `LottieAnimationView`. Some functionality that isn't included in this PR, but I think we would need before we could merge this, includes things like:
 1. controlling when the animation starts, as opposed to calling `play()` immediately / exactly once
 2. supporting displaying a static, fixed frame
 3. support for DotLottie animations
 4. support for `LottieConfiguration`
 5. support for text providers, image providers, etc

I'd also be interested in seeing a design that uses SwiftUI-style chaining (e.g. `LottieAnimationView(myAnimation).loopMode(.playOnce).playbackSpeed(1)`.

Ideally we'd also adopt the new view at least somewhere in the Lottie Example app, to make sure there's a way to test it within the repo.

calda:
I think we need to have a bit more discussion about the project direction before committing to merging and releasing this sort of change. If you're interested in collaborating on this, could you start a [discussion](https://github.com/airbnb/lottie-ios/discussions)? We can consider using this PR as a starting point once we're aligned on a direction.

I'm going to close this PR for now until we're aligned on how we want to move forward here.

PR #2082, opened 2023-06-21
Restructure `ValueProviderStore` to not accumulate multiple values for the same key

adamastern:
In our app we update value providers on animation views at various times throughout the view's lifecycle. This happens as the user changes their color theme preferences in our app.

Currently `ValueProviderStore` accumulates multiple providers for the same key, which leads to provider value resolution becoming more expensive over time.

This PR restructures `ValueProviderStore` so that it only keeps the most recent value provider for any given key. The updated class maintains the "last added" provider logic added in #1494.

### DIFF
--- a/Sources/Private/CoreAnimation/ValueProviderStore.swift
+++ b/Sources/Private/CoreAnimation/ValueProviderStore.swift
@@ -36,7 +36,9 @@ final class ValueProviderStore {
       properties. Supported properties are: \(supportedProperties.joined(separator: ", ")).
       """)
 
-    valueProviders.append((keypath: keypath, valueProvider: valueProvider))
+    orderedKeyPaths.remove(keypath)
+    orderedKeyPaths.insert(keypath, at: orderedKeyPaths.count)
+    valueProviders[keypath] = valueProvider
   }
 
   // Retrieves the custom value keyframes for the given property,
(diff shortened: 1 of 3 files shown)


Sources/Private/CoreAnimation/ValueProviderStore.swift
```diff
@@ -36,7 +36,9 @@ final class ValueProviderStore {
       properties. Supported properties are: \(supportedProperties.joined(separator: ", ")).
       """)
 
-    valueProviders.append((keypath: keypath, valueProvider: valueProvider))
+    orderedKeyPaths.remove(keypath)
+    orderedKeyPaths.insert(keypath, at: orderedKeyPaths.count)
```
calda, line 40: This is just an "append" right? Can we use a built-in method like https://developer.apple.com/documentation/foundation/nsmutableorderedset/1408009-add?

```suggestion
    orderedKeyPaths.add(keypath)
```

calda (approved):

Sources/Private/CoreAnimation/ValueProviderStore.swift
```diff
@@ -93,14 +94,13 @@ final class ValueProviderStore {
...
-
   private var valueProviders = [(keypath: AnimationKeypath, valueProvider: AnyValueProvider)]()
 
-  /// Retrieves the most-recently-registered Value Provider that matches the given keypat
+  /// Retrieves the most-recently-registered Value Provider that matches the given keypath.
   private func valueProvider(for keypath: AnimationKeypath) -> AnyValueProvider? {
     // Find the last keypath matching the given keypath,
     // so we return the value provider that was registered most-recently
-    valueProviders.last(where: { registeredKeypath, _ in
+    valueProviders.reversed().first(where: { registeredKeypath, _ in
```
calda, line 103: is `.reversed().first` different from `.last`?

Sources/Private/CoreAnimation/ValueProviderStore.swift
```diff
@@ -93,16 +95,19 @@ final class ValueProviderStore {
   // MARK: Private
 
   private let logger: LottieLogger
+  private var valueProviders = [AnimationKeypath: AnyValueProvider]()
+  private var orderedKeyPaths = NSMutableOrderedSet()
 
-  private var valueProviders = [(keypath: AnimationKeypath, valueProvider: AnyValueProvider)]()
```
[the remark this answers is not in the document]
adamastern: @calda thanks for the review! My thinking was that we could do faster lookup for existing keypaths `setValueProvider` with the set, but totally fine with keeping the existing data structure if that feels like an over-optimization. I reverted back to using an array in the latest commit, and now remove existing keys using `removeAll`. Let me know if that's what you had in mind!

Sources/Private/CoreAnimation/ValueProviderStore.swift
adamastern: No different, thanks! Reverted to using `last `.

calda (approved):
Thanks!

PR #2084, opened 2023-06-22
Add API for playing multiple markers sequentially

calda:
This PR adds an API for playing multiple markers sequentially.

I was testing this with one of the animations in the regression suite that happened to have markers. Here's a demo from calling:

```swift
animationView.play(markers: ["1", "1-reversed", "1", "2", "2-reversed", "2", "3", "3-reversed", "3"])
```


Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -243,6 +243,35 @@ public class LottieAnimationLayer: CALayer {
...
+  /// animation data and assigned a name. Multiple markers can be played sequentially
+  /// to create programmable animations.
+  ///
+  /// If a marker is not found, it will be skipped.
+  ///
+  /// If a marker doesn't have a duration value, it will play with a duration of 0
+  /// (effectively being skipped).
+  ///
+  /// - Parameter markers: The list of markers to play sequentially.
+  open func play(markers: [String]) {
```
thedrick, line 258: do we need to worry about the case where this could get called again even while the current sequence is still going? I would imagine if we called this again with a new set of markers, we should stop playing the previous set and start fresh with the new set

thedrick (approved):
LGTM after considering comment

Sources/Public/Animation/LottieAnimationLayer.swift
calda: Good thinking. I checked and this is handled automatically by the current implementation:
 1. when an animation is cancelled because another animation is started, the completion handler of the cancelled animation is called with `success: false`
 2. we only continue to the next marker segment when `success` is `true`
I'll add some comments to mention this behavior.

PR #2086, opened 2023-06-23
Add source for EpoxyCore library to support open-source SwiftUI view implementation

calda:
This PR adds the source of EpoxyCore 0.9.0 to this repo.

Airbnb has a Lottie `SwiftUI.View` implementation which is built using APIs in `EpoxyCore`. Making that dependency available in this library would let us open source this existing implementation.

Embedding the source of EpoxyCore into this repo feels reasonable and has existing precedent (we already integrate the ZipFoundation library in this way). EpoxyCore is only 3000 LOC, so does not meaningfully increase the size of this repo or the size of compiled binaries.

The only downside of this approach I can think of is that EpoxyCore doesn't currently support macOS. If the Lottie SwiftUI implementation uses EpoxyCore it will only be available on iOS and tvOS. I think this is probably an acceptable tradeoff, given the benefits of using the existing high-quality implementation. Thoughts?


calda:
This PR is ready for review @airbnb/lottie-ios-maintainers

.github/workflows/main.yml
```diff
@@ -136,4 +136,13 @@ jobs:
...
+        run: bundle exec rake lint:swift
+
+  embedded-libraries:
+    name: "Lint Embedded Libraries"
+    runs-on: macos-13
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/setup
+      - name: Lint Embedded Libraries
+        run: bundle exec rake lint:EmbeddedLibraries
```
calda, line 148: Added a new CI job that lints the `Sources/Private/EmbeddedLibraries` directory and verifies that it doesn't have any `public` symbols. Confirmed that it fails as expected if there's an unexpected `public` symbol:

Screenshot 2023-06-26 at 3 25 02 PM

Sources/Private/EmbeddedLibraries/EpoxyCore/README.md
```diff
@@ -0,0 +1,39 @@
+## EpoxyCore
```
calda, line 1: Added some discussion about why we include the source of `EpoxyCore`, plus instructions on how to update this in the future if necessary

thedrick (approved):
This LGTM @calda - the point about macOS does seem pretty important, especially given that SwiftUI's main conceit is that it can work across all platforms. I wonder how much work it would be to add support for macOS in the future

PR #2090, opened 2023-06-26
Add helper method to initialise DotLottie with data

kirti-swiggy:
This PRs adds a helper method for `DotLottieFile` to initialize a `DotLottieFile` instance with data (`Foundation.Data`). 

It follows [this](https://github.com/airbnb/lottie-ios/discussions/2087) discussion.

### DIFF
--- a/Sources/Public/DotLottie/DotLottieFileHelpers.swift
+++ b/Sources/Public/DotLottie/DotLottieFileHelpers.swift
@@ -78,6 +78,21 @@ extension DotLottieFile {
         return .failure(error)
       }
     }
+
+    /// Loads an DotLottie from a data synchronously. Returns a `DotLottieFile`. Optional.
+    ///
+    /// Please use the asynchronous methods whenever possible. This operation will block the Thread it is running in.
+    ///
+    /// - Parameters:
+    ///   - data: The data(`Foundation.Data`) object to load DotLottie from
+    ///   - filename: The name of the lottie file without the lottie extension. eg. "StarAnimation"
+    public static func loadedFrom(
+      data: Data,
+      filename: String)
+      throws -> DotLottieFile
+    {
+      return try DotLottieFile(data: data, filename: filename)
+    }
   }
 
   /// Loads a DotLottie model from a bundle by its name. Returns `nil` if a file is not found.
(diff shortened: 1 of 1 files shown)


Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -289,4 +289,21 @@ extension DotLottieFile {
...
   }
 
+  /// Loads an DotLottie from a data synchronously. Returns a `Result<DotLottieFile, Error>`
+  /// - Parameters:
+  ///   - data: The data(`Foundation.Data`) object to load DotLottie from
+  ///   - filename: The name of the lottie file without the lottie extension. eg. "StarAnimation"
+  public static func loadedFrom(
+    data: Data,
+    filename: String)
+    -> Result<DotLottieFile, Error>
```
calda, line 299: Rather than using `Result`, we should use `throws` and use just `DotLottieFile` as the return type

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -289,4 +289,21 @@ extension DotLottieFile {
...
+  {
+    do {
+      let dotLottieFile = try DotLottieFile(data: data, filename: filename)
```
calda, line 302: `DotLottieFile.init` performs a lot of work on the calling thread (it decompresses a zip file). This method should be asynchronous with a completion callback so the work isn't done on the main thread. We should also provide an `async` version that wraps the implementation that takes a completion handler (using `withCheckedThrowingContinuation`).

calda (commented):
Thank you! Before merging we'll want to make this method asynchronous, and use `throws` instead of `result`.

Sources/Public/DotLottie/DotLottieFileHelpers.swift
kirti-swiggy: I actually took reference from the other signatures where `Result` was used over throws. But, the feedback makes sense. Have made the changes.

Sources/Public/DotLottie/DotLottieFileHelpers.swift
kirti-swiggy: Done.

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -289,4 +289,65 @@ extension DotLottieFile {
     }
   }
 
+  /// Loads an DotLottie from a data synchronously. Returns a `Result<DotLottieFile, Error>`
+  /// Please use the asynchronous methods whenever possible. This operation will block the Thread it is running in.
+  ///
+  /// - Parameters:
+  ///   - data: The data(`Foundation.Data`) object to load DotLottie from
+  ///   - filename: The name of the lottie file without the lottie extension. eg. "StarAnimation"
+  public static func loadedFrom(
```
calda, line 298: If you want to include this helper, move it to the `SynchronouslyBlockingCurrentThread` namespace above

calda (approved):
Thanks! Let's move the synchronous method to the `SynchronouslyBlockingCurrentThread` namespace but otherwise LGTM

Sources/Public/DotLottie/DotLottieFileHelpers.swift
kirti-swiggy: Sure. Done.

PR #2092, opened 2023-06-28
Add SwiftUI LottieView

calda:
This PR open sources Airbnb's internal SwiftUI `LottieView` implementation. We'll use this as a starting point, and will add more functionality in follow-up PRs.


Sources/Public/Animation/LottieView.swift
```diff
@@ -0,0 +1,103 @@
+// Created by Bryn Bodayle on 1/20/22.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+import SwiftUI
+
+#if !os(macOS)
```
calda, line 6: At the moment we can't provide macOS support because EpoxyCore only supports UIKit.

I made a proof-of-concept change locally that shows we can add macOS support to EpoxyCore fairly easily, which lets us use this `LottieView` on macOS. I'll discuss those potential changes in follow-ups.

Screenshot 2023-06-28 at 9 53 30 AM

Example/tvOS/ContentView.swift
```diff
@@ -0,0 +1,17 @@
...
+struct ContentView: View {
+  var body: some View {
+    ZStack {
+      LottieView(animation: LottieAnimation.named("Samples/LottieLogo1")!)
+        .looping()
+        .resizable()
+    }
+    .ignoresSafeArea()
+    .frame(maxWidth: .infinity, maxHeight: .infinity)
+  }
```
calda, line 16: I deleted the previous tvOS UIKit example app, and created a new SwiftUI sample app. So much simpler!

brynbodayle (approved):

PR #2093, opened 2023-06-28
Fix breaking API changes from #2073

calda:
I noticed #2073 accidentally removed the `LottieAnimationView.configuration` property. This PR adds it back, as a passthrough to the underlying `LottieAnimationLayer.configuration`.


thedrick (approved):
LGTM @calda

PR #2098, opened 2023-07-04
Marks animation as optional in initializer.

fonkadelic:
This marks the animation of the newly introduced `LottieView` for SwiftUI as optional. 

The main motivation for this change was to align it with the initializer of the `LottieAnimationView`. Another reason is probably that a common way to create an animation is via `LottieAnimation.named` and this function returns an optional result.


calda (approved):
Thank you for this improvement! Agree that this is better.

PR #2100, opened 2023-07-05
VisionOS support

phiren:
- Changes related with providing support to visionOS


calda (approved):
Thanks!

Looks like there are some build failures in a few CI jobs. Those will need to be fixed before we can merge. You can see the individual commands being ran for those jobs in the `Rakefile`.

You'll also need to run the code formatter, which you can do with `bundle exec rake format:swift`

calda:
I'm going to close this PR for now since there are CI failures. Please feel free to re-open it if you try to fix the CI failures later!

PR #2104, opened 2023-07-10
Add LottieView support on macOS using EpoxyCore 0.10.0

calda:
This PR:
 - Updates our EpoxyCore dependency to 0.10.0, which adds macOS support: https://github.com/airbnb/epoxy-ios/pull/143
 - Updates the SwiftUI `LottieView` to be available on macOS
 - Removes the tvOS and macOS example apps in favor of a single multiplatform SwiftUI example app that runs on all platforms. Eventually this will replace the iOS Example app as well.

I also ported some functionality from the UIKit example app to the SwiftUI example app, which now works on macOS and tvOS:


Example/Example/AnimationListView.swift
```diff
@@ -0,0 +1,93 @@
...
+            AnimationListView(directory: "\(directory)/\(subdirectoryURL.lastPathComponent)")
+          }
+        }
+      }
+    }.navigationTitle(directory)
+  }
+
+}
+
+extension AnimationListView {
```
calda, line 46: Everything in this extension was copy/pasted from the UIKit example app

Example/Example/AnimationListView.swift
```diff
@@ -0,0 +1,93 @@
...
+import Lottie
+import SwiftUI
+
+// MARK: - AnimationListView
+
+struct AnimationListView: View {
+
+  let directory: String
+
+  var body: some View {
```
calda, line 13: Nice and simple -- this was like 200 LOC plus the entire EpoxyCollectionView / EpoxyLayoutGroups dependency before in the UIKit example app

Example/Example/AnimationListView.swift
```diff
@@ -0,0 +1,93 @@
...
+    List {
+      ForEach(items, id: \.self) { item in
+        NavigationLink(value: item) {
```
calda, line 16: over here googling simple things like "swiftui how to push on navigation stack"

calda: I have now learned that there are 5-6 different APIs for doing this and I don't know the differences. Classic SwiftUI.

calda:
ptal @brynbodayle

brynbodayle (approved):

PR #2108, opened 2023-07-13
Add SwiftUI APIs to get and set current progress / current frame

calda:
This PR adds SwiftUI `LottieView` APIs to get and set the `currentProgress` and `currentFrame` of the animation.

The `currentProgress` / `currentFrame` APIs let you do things like:

```swift
// The animation is paused at 50% progress
LottieView(animation: .named(animationName))
  .currentProgress(0.5)
```

```swift
// The animation always displays the frame of the slider
@State private var sliderValue: AnimationProgressTime = 0

// ...

LottieView(animation: .named(animationName))
  .currentProgress(sliderValue)

Slider(value: $sliderValue, in: 0...1)
```

This PR also includes APIs for accessing the `realtimeAnimationProgress` / `realtimeAnimationFrame` and assigning that value to a binding. You'd use this to update other pieces of UI to match the state of the animation, like a progress bar. This is a bit more niche, but since I was writing the functionality for the Example anyway I figured we may as well add it as a public API. 

For example, here's usage in the Example app where the slider displays the live-updating `realtimeAnimationProgress` until you manually start dragging the slider:

```swift
@State private var animationPlaying: Bool = true
@State private var sliderValue: AnimationProgressTime = 0

var body: some View {
  VStack {
    LottieView(animation: .named(animationName))
      .looping()
      .currentProgress(animationPlaying ? nil : sliderValue)
      .getRealtimeAnimationProgress(animationPlaying ? $sliderValue : nil)

    Slider(value: $sliderValue, in: 0...1, onEditingChanged: { editing in
      if animationPlaying, editing {
        animationPlaying = false
      }
    })
  }
}
```

### DIFF
--- a/Sources/Public/Animation/LottieView.swift
+++ b/Sources/Public/Animation/LottieView.swift
@@ -58,7 +58,7 @@ public struct LottieView: UIViewConfiguringSwiftUIView {
     return copy
   }
 
-  /// Returns a copy of this animation view that loops its animation whenever visible by playing
+  /// Returns a copy of this view that loops its animation whenever visible by playing
   /// whenever it is updated with a `loopMode` of `.loop` if not already playing.
   public func looping() -> Self {
     configure { view in
@@ -68,8 +68,7 @@ public struct LottieView: UIViewConfiguringSwiftUIView {
     }
   }
 
-  /// Returns a copy of this animation view with its `AnimationView` updated to have the provided
-  /// background behavior.
+  /// Returns a copy of this view updated to have the provided background behavior.
   public func backgroundBehavior(_ value: LottieBackgroundBehavior) -> Self {
     configure { view in
       view.backgroundBehavior = value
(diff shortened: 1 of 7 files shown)


calda:
ptal @brynbodayle

Sources/Public/Animation/LottieView.swift
```diff
@@ -162,6 +161,94 @@ public struct LottieView: UIViewConfiguringSwiftUIView {
...
+        view.currentTime = currentTime
+      }
+    }
+  }
+
+  /// Returns a view that updates the given binding each frame with the animation's `realtimeAnimationProgress`.
+  /// The `LottieView` is wrapped in a `TimelineView` with the `.animation` schedule.
+  ///  - This is a one-way binding. Its value is updated but never read.
+  ///  - If provided, the binding will be updated each frame with the `realtimeAnimationProgress`
+  ///    of the underlying `LottieAnimationView`. This is potentially expensive since it triggers
```
brynbodayle, line 216: Hmm, in the future we might want the ability to specify a level of granularity for how often you receive updates. Say every second or half second.

brynbodayle (approved):

Sources/Public/Animation/LottieView.swift
calda: That sounds like it could be useful. We can definitely add it later if a use case comes up.

PR #2109, opened 2023-07-13
Improve SwiftUI APIs for configuring value providers, image providers, etc

calda:
This PR improves the SwiftUI `LottieView` APIs for configuring:
 - value providers
 - image providers
 - font providers
 - text providers
 - accessibility labels
 - `LottieConfiguration`

All of these now have a builder-style method like `.imageProvider(MyImageProvider())`, and supported being updated at any time using equality checks.

For types like `AnimationImageProvider` that aren't `Equatable`, the builder methods use an `& Equatable` constraint. This PR adds an `Equatable` conformance to all of the first-party value provider implementations.

### DIFF
--- a/Sources/Public/Animation/LottieView.swift
+++ b/Sources/Public/Animation/LottieView.swift
@@ -11,20 +11,9 @@ public struct LottieView: UIViewConfiguringSwiftUIView {
 
   // MARK: Lifecycle
 
-  public init(
-    animation: LottieAnimation?,
-    imageProvider: AnimationImageProvider? = nil,
-    textProvider: AnimationTextProvider? = nil,
-    fontProvider: AnimationFontProvider? = nil,
-    configuration: LottieConfiguration = .shared,
-    accessibilityLabel: String? = nil)
-  {
+  /// Creates a `LottieView` that displays the given animation
+  public init(animation: LottieAnimation?) {
     self.animation = animation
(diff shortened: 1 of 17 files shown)


Sources/Public/Animation/LottieView.swift
```diff
@@ -99,17 +76,102 @@ public struct LottieView: UIViewConfiguringSwiftUIView {
...
+
+  /// Returns a copy of this view with its image provider updated to the given value.
+  /// The image provider must be `Equatable` to avoid unnecessary state updates / re-renders.
+  public func imageProvider<ImageProvider: AnimationImageProvider & Equatable>(_ imageProvider: ImageProvider) -> Self {
+    var copy = self
+    copy.imageProvider = imageProvider
+
+    copy = copy.configure { view in
+      if (view.imageProvider as? ImageProvider) != imageProvider {
+        view.imageProvider = imageProvider
```
calda, line 114: Pretty neat -- in the example app where we're using `.imageProvider(.exampleAppSampleImages)`, this line here actually never runs. The correct image provider is used when initializing the `LottieAnimationView` above in `LottieAnimationView.swiftUIView { ... }` and then this doesn't need to run again since the equality check is succeeding.

calda:
ptal @brynbodayle

brynbodayle (approved):

PR #2110, opened 2023-07-14
Play "reduced motion" marker if `UIAccessibility.isReduceMotionEnabled` is true

calda:
This PR adds support for respecting the system "reduction motion" option.

If a Lottie animation includes a marker named "reduced motion" (or "reducedMotion", "reduced_motion", etc) and `UIAccessibility.isReduceMotionEnabled` is true, that marker will be be played instead of any other animation that was configured.

Here's an example animation that includes a marker named "reduced motion":

| Reduced motion disabled | Reduced motion enabled |
| ---- | ----- |
|  |  |

### DIFF
--- a/Sources/Public/Configuration/DecodingStrategy.swift
+++ b/Sources/Public/Configuration/DecodingStrategy.swift
@@ -1,0 +1,15 @@
+// Created by Cal Stephens on 7/14/23.
+// Copyright © 2023 Airbnb Inc. All rights reserved.
+
+/// How animation files should be decoded
+public enum DecodingStrategy: Hashable {
+  /// Use Codable. This is was the default strategy introduced on Lottie 3, but should be rarely
+  /// used as it's slower than `dictionaryBased`. Kept here for any possible compatibility issues
+  /// that may come up, but consider it soft-deprecated.
+  case legacyCodable
+
+  /// Manually deserialize a dictionary into an Animation.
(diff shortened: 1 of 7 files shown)


Sources/Public/Configuration/DecodingStrategy.swift
```diff
@@ -0,0 +1,15 @@
+// Created by Cal Stephens on 7/14/23.
+// Copyright © 2023 Airbnb Inc. All rights reserved.
+
+/// How animation files should be decoded
+public enum DecodingStrategy: Hashable {
```
calda, line 5: Moved this to its own file

Sources/Public/Configuration/LottieConfiguration.swift
```diff
@@ -0,0 +1,47 @@
+// Created by Cal Stephens on 12/13/21.
+// Copyright © 2021 Airbnb Inc. All rights reserved.
+
+import QuartzCore
+
+/// Global configuration options for Lottie animations
+public struct LottieConfiguration: Hashable {
```
calda, line 7: Moved this to its own file and added the `reducedMotionOption` property

calda:
This change is ready for review. @thedrick, could you review this one?

thedrick (approved):
this is awesome @calda thanks for adding this feature!

PR #2111, opened 2023-07-18
Remove unused class

noahsmartin:
I noticed `NullHapticGenerator` was unused from a [Reaper](https://docs.emergetools.com/docs/reaper) report. It looks like since Lottie's minimum iOS version was increased to 11 we can remove all uses of this on iOS, and just leave it for other targets that don't support `UIImpactFeedbackGenerator`


calda (approved):
Thanks!

PR #2116, opened 2023-07-25
Fix default `ReducedMotionOption`

calda:
#2110 accidentally set the default `ReducedMotionOption` to `.reducedMotion`. I must have updated this while debugging and forgotten to change it back.

This PR updates the default to the expected `.systemReducedMotionToggle` value.


thedrick (approved):
LGTM @calda sorry I didn't catch this in my review last time

PR #2118, opened 2023-07-26
Support fetching multiple async animations

miguel-jimenez-0529:
This PR adds a new `Binding` to `LottieView` that serves as a reference to fetch more than one async animation

animations-loop


Sources/Public/Animation/View+ValueChanged.swift
```diff
@@ -0,0 +1,20 @@
...
+// Copyright © 2023 Airbnb Inc. All rights reserved.
+
+import Combine
+import SwiftUI
+
+@available(iOS 13.0, tvOS 13.0, macOS 10.15, *)
+extension View {
+  /// A backwards compatible wrapper for iOS 14 `onChange`
+  @ViewBuilder
+  func valueChanged<T: Equatable>(value: T, onChange: @escaping (T) -> Void) -> some View {
```
calda, line 11: Since this is internal, instead of putting this in `Sources/Public` could you move it to `Sources/Private/Utilities/Helpers`?

Sources/Public/Animation/LottieView.swift
```diff
@@ -67,17 +95,27 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
     self.loadAnimation = loadAnimation
     self.placeholder = placeholder
+    self.loadAnimationTrigger = loadAnimationTrigger
     _animationSource = State(initialValue: nil)
   }
 
   // MARK: Public
 
+  /// Token used when no `loadAnimationTrigger` is provided
+  public enum LoadReference {
```
calda, line 115: Instead of using `loadAnimationTrigger = .constant(LoadReference.constantSource)` to mean "call this only once", could we make it optional and use either `loadAnimationTrigger = nil` or `loadAnimationTrigger = .constant(nil)`? 

It would be nice to not add a new public value just for this functionality

Sources/Public/Animation/LottieView.swift
```diff
@@ -16,28 +15,44 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
 
   /// Creates a `LottieView` that asynchronously loads and displays the given `LottieAnimation`.
   /// The `loadAnimation` closure is called exactly once in `onAppear`.
-  public init(_ loadAnimation: @escaping () async throws -> LottieAnimation?) where Placeholder == EmptyView {
-    self.init(loadAnimation, placeholder: EmptyView.init)
+  /// - Parameters:
+  ///   - loadAnimationTrigger: `Binding` that  triggers a new call to `loadAnimation` when a new value is propagated
+  ///   - loadAnimation: Closure that provides an animation asynchronously
+  public init(
+    loadAnimationTrigger: Binding<AnyHashable> = .constant(LoadReference.constantSource),
```
calda, line 34: Instead of adding this an an argument to all of the `init` values, could we add a builder function for this (e.g. `func loadAnimationTrigger(_ binding: Binding<AnyHashable>) -> Self`)? 

This is optional and won't be used in all use cases. Anything optional seems better to provide using builder functions rather than in each individual `init`. 

Then we can just mention this in the doc comment of all of the `loadAnimation` initializers, like "By default the `loadAnimation` closure is called exactly once in `onAppear`. To trigger the closure to be called again at a later time, configure `.loadAnimationTrigger($reloadAnimation)` with a binding.

Sources/Public/Animation/LottieView.swift
```diff
@@ -344,10 +386,7 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
   private let placeholder: (() -> Placeholder)?
 
   private func loadAnimationIfNecessary() {
-    guard
-      let loadAnimation = loadAnimation,
-      animationSource == nil
```
calda, line 349: I think this change means we'll re-call the `loadAnimation` every time `onAppear` is called, which I don't think we want. Perhaps instead we should change the `valueChanged` block to:

```
.valueChanged(value: loadAnimationTrigger.wrappedValue) { _ in
  animation = nil
  loadAnimationIfNecessary()
}
```

This additionally makes it so updating the `loadAnimationTrigger` puts the view back in a loading state, which may or may not be desirable. I do think this makes sense though, since we've effectively changed the animation data source to no longer refer to the previous animation.

Example/Example/AnimationPreviewView.swift
```diff
@@ -54,12 +70,29 @@ struct AnimationPreviewView: View {
...
   @State private var animationPlaying = true
   @State private var sliderValue: AnimationProgressTime = 0
+  @State private var currentURLIndex: Int
+
+  private var fetchTrigger: Binding<AnyHashable> {
+    switch animationSource {
+    case .local:
+      return .constant(currentURLIndex)
+    case .remote:
+      return $currentURLIndex.mapGetter(transform: AnyHashable.init)
```
calda, line 93: I wonder if using `func loadAnimationTrigger<T: Hashable>(_ binding: Binding<Hashable>) -> Self)` instead of `AnyHashable` would let us avoid having to map this binding to an `AnyHashable` in the consumer code and instead let us pass in a `Binding<Int>`. That would be a nice callsite simplification.

calda (commented):
Pretty awesome, thanks for working on this! I left some comments on the API design and a few of the implementation details.

Sources/Public/Animation/LottieView.swift
miguel-jimenez-0529: @calda I followed your recommendation. Looks like we do show the placeholder view after switching animations. The GIF in the PR description is updated with how it looks like after the changes

miguel-jimenez-0529:
@calda this should be ready for another pass. Thanks for all the suggestions!

Example/Example/AnimationPreviewView.swift
```diff
@@ -33,9 +48,11 @@ struct AnimationPreviewView: View {
         try await lottieSource()
       } placeholder: {
         LoadingIndicator()
+          .frame(width: 50, height: 50)
       }
       .imageProvider(.exampleAppSampleImages)
       .resizable()
+      .loadAnimationTrigger($currentURLIndex)
```
calda, line 55: pretty awesome. this feels really intuitive to me.

Sources/Public/Animation/LottieView.swift
```diff
@@ -287,6 +304,18 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     }
   }
 
+  /// Returns a new instance of this view, which will invoke the provided `loadAnimation` closure
+  /// whenever the `binding` value is updated.
+  ///
+  /// - Note: This function requires a valid `loadAnimation` closure provided during view initialization.
+  /// - Warning: The existing animation will be removed before calling `loadAnimation`,
+  ///            potentially rendering the provided `Placeholder`.
```
calda, line 312: I think a "note" is more appropriate here than a "warning"

```suggestion
  /// - Note: The existing animation will be removed before calling `loadAnimation`,
  ///         which will cause the `Placeholder` to be displayed until the new animation finishes loading.
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -287,6 +304,18 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     }
   }
 
+  /// Returns a new instance of this view, which will invoke the provided `loadAnimation` closure
+  /// whenever the `binding` value is updated.
+  ///
+  /// - Note: This function requires a valid `loadAnimation` closure provided during view initialization.
```
calda, line 310:
```suggestion
  /// - Note: This function requires a valid `loadAnimation` closure provided during view initialization,
  ///         otherwise the `loadAnimationTrigger` will have no effect.
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -46,12 +52,18 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
   /// The `loadDotLottieFile` closure is called exactly once in `onAppear`.
-  public init(_ loadDotLottieFile: @escaping () async throws -> DotLottieFile?) where Placeholder == EmptyView {
-    self.init(loadDotLottieFile, placeholder: EmptyView.init)
+  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
+  public init(
+    _ loadDotLottieFile: @escaping () async throws -> DotLottieFile?) where Placeholder == EmptyView
+  {
+    self.init(
+      loadDotLottieFile,
+      placeholder: EmptyView.init)
```
calda, line 61: same here

```suggestion
    self.init(loadDotLottieFile, placeholder: EmptyView.init)
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -46,12 +52,18 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
   }
 
   /// Creates a `LottieView` that asynchronously loads and displays the given `DotLottieFile`.
   /// The `loadDotLottieFile` closure is called exactly once in `onAppear`.
+  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
```
calda, line 66: Suggestion that would apply to all of the inits with this sentence:

```suggestion
  /// If you wish to call `loadAnimation` again at a different time, you can use `.loadAnimationTrigger(...)`.
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -26,12 +25,19 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
 
   /// Creates a `LottieView` that asynchronously loads and displays the given `LottieAnimation`.
   /// The `loadAnimation` closure is called exactly once in `onAppear`.
-  public init(_ loadAnimation: @escaping () async throws -> LottieAnimation?) where Placeholder == EmptyView {
-    self.init(loadAnimation, placeholder: EmptyView.init)
+  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
+  public init(
+    _ loadAnimation: @escaping () async throws -> LottieAnimation?)
+    where Placeholder == EmptyView
+  {
```
calda, line 32: Same here

```suggestion
  public init(_ loadAnimation: @escaping () async throws -> LottieAnimation?) where Placeholder == EmptyView {
```

calda (approved):
Awesome! Really nice how simple the changes are in `LottieView` itself.

Sources/Public/Animation/LottieView.swift
```diff
@@ -46,12 +47,16 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
 
   /// Creates a `LottieView` that asynchronously loads and displays the given `DotLottieFile`.
   /// The `loadDotLottieFile` closure is called exactly once in `onAppear`.
-  public init(_ loadDotLottieFile: @escaping () async throws -> DotLottieFile?) where Placeholder == EmptyView {
+  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
+  public init(
+    _ loadDotLottieFile: @escaping () async throws -> DotLottieFile?) where Placeholder == EmptyView
+  {
```
miguel-jimenez-0529, line 53:
```suggestion
  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
  public init(_ loadDotLottieFile: @escaping () async throws -> DotLottieFile?) where Placeholder == EmptyView {
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -66,6 +69,7 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
 
   /// Creates a `LottieView` that asynchronously loads and displays the given `LottieAnimationSource`.
   /// The `loadAnimation` closure is called exactly once in `onAppear`.
+  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
```
calda, line 72:
```suggestion
  /// If you wish to call `loadAnimation` again at a different time, you can use `.loadAnimationTrigger(...)`.
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -46,12 +47,14 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
 
   /// Creates a `LottieView` that asynchronously loads and displays the given `DotLottieFile`.
   /// The `loadDotLottieFile` closure is called exactly once in `onAppear`.
+  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
```
calda, line 50:
```suggestion
  /// If you wish to call `loadAnimation` again at a different time, you can use `.loadAnimationTrigger(...)`.
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -26,12 +25,14 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
   /// Creates a `LottieView` that asynchronously loads and displays the given `LottieAnimation`.
   /// The `loadAnimation` closure is called exactly once in `onAppear`.
+  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
   public init(_ loadAnimation: @escaping () async throws -> LottieAnimation?) where Placeholder == EmptyView {
     self.init(loadAnimation, placeholder: EmptyView.init)
   }
 
   /// Creates a `LottieView` that asynchronously loads and displays the given `LottieAnimation`.
   /// The `loadAnimation` closure is called exactly once in `onAppear`.
+  /// If you wish to call `loadAnimation` at a different time, you can use the `loadAnimationTrigger` chain function.
```
calda, line 35:
```suggestion
  /// If you wish to call `loadAnimation` again at a different time, you can use `.loadAnimationTrigger(...)`.
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -26,12 +25,14 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
```
calda, line 28:
```suggestion
  /// If you wish to call `loadAnimation` again at a different time, you can use `.loadAnimationTrigger(...)`.
```

PR #2126, opened 2023-08-02
Support using SwiftUI LottieView with strict concurrency enabled

calda:
This PR add support for using the SwiftUI `LottieView` with strict concurrency enabled, and adopts strict concurrency in the example apps as a CI test case for this support.


rugoli (approved):

Example/Example/RemoteAnimationDemoView.swift
```diff
@@ -6,6 +6,7 @@ import SwiftUI
 
 // MARK: - AnimationListView
 
+@MainActor
 struct RemoteAnimationsDemoView: View {
```
miguel-jimenez-0529, line 10: Should't Views be main actors by default?

Sources/Public/Animation/LottieAnimation.swift
```diff
@@ -177,3 +177,7 @@ public final class LottieAnimation: Codable, DictionaryInitializable {
     })
   }()
 }
+
+// MARK: Sendable
+
+extension LottieAnimation: @unchecked Sendable { }
```
miguel-jimenez-0529, line 183: Should we add a comment about why this is `@unchecked`?

Sources/Public/DotLottie/DotLottieFile.swift
```diff
@@ -144,3 +144,7 @@ extension String {
     (self as NSString).deletingPathExtension
   }
 }
+
+// MARK: - DotLottieFile + Sendable
+
+extension DotLottieFile: @unchecked Sendable { }
```
miguel-jimenez-0529, line 150: dito

miguel-jimenez-0529 (approved):
Thanks! Really quick update 💯

Example/Example/RemoteAnimationDemoView.swift
calda: I thought so too, but you get a concurrency error without this. Maybe only `body` is implicitly main-actor but not the other properties?

Sources/Public/Animation/LottieAnimation.swift
calda: I switched this to use `Sendable` instead of `@unchecked Sendable`. This required marking all of the other data model types as `Sendable` as well, but that's fine since they're all immutable.

PR #2127, opened 2023-08-02
Async animation loading improvements

miguel-jimenez-0529:
This PR adds the following changes:
- Renames `loadAnimationTrigger` to `reloadAnimationTrigger`
- Replaces `reloadAnimationTrigger` binding value from `Hashable `to `Equatable`
- Fixes an issue where the animation is accidentally removed when calling `reloadAnimationTrigger` without a valid `loadAnimation` closure.
- Add documentation to `init(dotLottieFile:)` about not using the initializer with `SynchronouslyBlockingCurrentThread` APIs
- Add `assert` to `SynchronouslyBlockingCurrentThread` APIs when called from the main thread.


Sources/Private/Utility/Helpers/View+ValueChanged.swift
```diff
@@ -8,7 +8,7 @@ import SwiftUI
 extension View {
   /// A backwards compatible wrapper for iOS 14 `onChange`
   @ViewBuilder
-  func valueChanged<T: Equatable>(value: T, onChange: @escaping (T) -> Void) -> some View {
+  func valueChanged(value: any Equatable, onChange: @escaping (any Equatable) -> Void) -> some View {
```
calda, line 11: I think we'll need to use the old syntax here since Lottie still supports Swift 5.5. CI will probably fail with this written as-is.

```suggestion
  func valueChanged<T: Equatable>(value: T, onChange: @escaping (T) -> Void) -> some View {
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -17,22 +17,26 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     placeholder = nil
   }
 
-  /// Creates a `LottieView` that displays the given `DotLottieFile`
+  /// Initializes a `LottieView` with the provided `DotLottieFile` for display.
+  ///
+  /// - Important: Avoid using this initializer with the `SynchronouslyBlockingCurrentThread` APIs.
+  ///              If decompression of a `.lottie` file is necessary, prefer using the `.init(_ loadAnimation:)`
+  ///              initializer, which decompresses the file asynchronously.
```
calda, line 24: Including an example may be helpful here: 

```suggestion
  /// - Important: Avoid using this initializer with the `SynchronouslyBlockingCurrentThread` APIs.
  ///              If decompression of a `.lottie` file is necessary, prefer using the `.init(_ loadAnimation:)`
  ///              initializer, which takes an asynchronous closure:
  ///              
  ///                LottieView {
  ///                  try await DotLottieFile.asset(named: myDotLottieFile)
  ///                }
  ///
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -299,12 +317,16 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
-  /// - Note: The existing animation will be removed before calling `loadAnimation`,
-  ///         which will cause the `Placeholder` to be displayed until the new animation finishes loading.
-  public func loadAnimationTrigger<Value: Hashable>(_ binding: Binding<Value>) -> Self {
+  ///         otherwise `reloadAnimationTrigger` will have no effect.
+  /// - Parameters:
+  ///   - binding: The binding that triggers the reloading when its value changes.
+  ///   - showPlaceholder: When `true`, the current animation will be removed before invoking `loadAnimation`,
+  ///     displaying the `Placeholder` until the new animation loads.
+  ///     When `false`, the previous animation remains visible while the new one loads.
+  public func reloadAnimationTrigger<Value: Equatable>(_ binding: Binding<Value>, showPlaceholder: Bool = true) -> Self {
```
calda, line 326: does this need to be a `Binding`, or can it just take `Value` directly?

```suggestion
  public func reloadAnimationTrigger<Value: Equatable>(_ value: Value, showPlaceholder: Bool = true) -> Self {
```

calda: All we need to do is trigger a state update when `reloadAnimationTrigger` changes, right? That should still happen without using a binding

calda: maybe if we use `@State` instead of `Binding<any Equatable>`?

Sources/Public/Animation/LottieView.swift
```diff
@@ -380,4 +400,16 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+    if loadAnimation != nil, showPlaceholderWhileReloading {
+      animationSource = nil
+    }
+  }
+}
+
+// MARK: - DefaultLoadTrigger
+
+/// The default load animation trigger token
+private enum DefaultLoadTrigger: Hashable {
```
calda, line 413: Why is this necessary now if we didn't have it before?

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -20,6 +20,8 @@ extension DotLottieFile {
       dotLottieCache: DotLottieCacheProvider? = DotLottieCache.sharedCache)
       -> Result<DotLottieFile, Error>
     {
+      assert(!Thread.isMainThread, "Calling this API in the main thread will affect the performance of your App")
```
calda, line 23: This should use `LottieLogger.shared.assert`

```suggestion
      LottieLogger.shared.assert(!Thread.isMainThread, "`DotLottieFile.SynchronouslyBlockingCurrentThread` methods shouldn't be called on the main thread.")
```

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -55,6 +57,8 @@ extension DotLottieFile {
...
```
calda, line 60:
```suggestion
      LottieLogger.shared.assert(!Thread.isMainThread, "`DotLottieFile.SynchronouslyBlockingCurrentThread` methods shouldn't be called on the main thread.")
```

Sources/Public/DotLottie/DotLottieFileHelpers.swift
```diff
@@ -91,6 +95,7 @@ extension DotLottieFile {
       filename: String)
       -> Result<DotLottieFile, Error>
     {
+      assert(!Thread.isMainThread, "Calling this API in the main thread will affect the performance of your App")
```
calda, line 98:
```suggestion
```suggestion
      LottieLogger.shared.assert(!Thread.isMainThread, "`DotLottieFile.SynchronouslyBlockingCurrentThread` methods shouldn't be called on the main thread.")
```
```

File: Sources/Private/Utility/Helpers/View+ValueChanged.swift (hunk 1 above)
Comment (calda) [reply Sources/Private/Utility/Helpers/View+ValueChanged.swift:11]: Yeah, CI is failing so we can't use `any` like this

File: Sources/Public/Animation/LottieView.swift [hunk 8]
```diff
@@ -299,12 +317,16 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+  ///         otherwise `reloadAnimationTrigger` will have no effect.
+  /// - Parameters:
+  ///   - binding: The binding that triggers the reloading when its value changes.
+  ///   - showPlaceholder: When `true`, the current animation will be removed before invoking `loadAnimation`,
+  ///     displaying the `Placeholder` until the new animation loads.
+  ///     When `false`, the previous animation remains visible while the new one loads.
+  public func reloadAnimationTrigger<Value: Equatable>(_ binding: Binding<Value>, showPlaceholder: Bool = true) -> Self {
     var copy = self
-    copy.loadAnimationTrigger = binding.map(transform: AnyHashable.init)
+    copy.reloadAnimationTrigger = binding.map { $0 as any Equatable }
```
calda, line 328: We cant use `any Equatable` due to backwards compatibility requirements (CI is failing), so I think we need to switch back to requiring `Hashable` so we can use `AnyHashable` here. Fortunately most `Equatable` types are also `Hashable`.

Sources/Public/Animation/LottieView.swift
calda: We should be able to do the same `onChange { loadAnimationIfNecessary() }` logic if this used `@State` instead of a binding, right?

Sources/Public/Animation/LottieView.swift
```diff
@@ -380,4 +400,16 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     }
   }
 
+  private func clearAnimationIfNecessary() {
+    if loadAnimation != nil, showPlaceholderWhileReloading {
+      animationSource = nil
+    }
+  }
```
calda, line 407: I'm not sure about this specific method since it's tightly coupled to the `reloadAnimationTrigger` use case and it doesn't seem like we'd call it again for something else.

Maybe instead organization like this would make sense:

```suggestion
  private func reloadAnimationTriggerDidChange() {
    guard loadAnimation != nil else { return }
    
    if showPlaceholderWhileReloading {
      animationSource = nil
    }
    
    loadAnimationIfNecessary()
}
```

this feels more self-documenting to me, compared to `clearAnimationIfNecessary`

calda (approved):
Thanks! All good improvements.

Sources/Private/Utility/Helpers/View+ValueChanged.swift
miguel-jimenez-0529: Oh no, @calda we could introduce a new type such as [AnyEquatable]( https://gist.github.com/JadenGeller/f0d05a4699ddd477a2c1) and still being able to migrate off `AnyHashable`. wdyt?

calda: Sure, that seems reasonable to me. Better than adding an unnecessary `Hashable` constraint.

Sources/Public/Animation/LottieView.swift
miguel-jimenez-0529: I think the recommendation would be to pass a `Binding` when your view doesn't own the `@State`. https://swiftuipropertywrappers.com/

Sources/Public/Animation/LottieView.swift
miguel-jimenez-0529: I removed the `Optional` in the `fetchTrigger` binding so that the code could compile with the `any Equatable` changes. We use this as the default value for the binding now. Hopefully we can remove this if we remove the `any Equatable` changes.

miguel-jimenez-0529: I was able to remove this

Sources/Public/Animation/LottieView.swift
calda: @denandreychuk, so you have an suggestion for a `LottieView.reloadAnimationTrigger` API that doesn't require using a binding? Specifically important is that the `onChange` has to be inside `LottieView` itself rather than the parent view.

denandreychuk: I believe that your initial suggestion should work. It will trigger an `onChange` event as long as `value` is annotated with `@State`, `@Binding`, or `@Published`. However, this will not work if we have a regular property that is not marked with any of these property wrappers, which is fine.

```
public func reloadAnimationTrigger<Value: Equatable>(_ value: Value, showPlaceholder: Bool = true) -> Self
```

denandreychuk: > Specifically important is that the onChange has to be inside LottieView itself rather than the parent view.` 

It will.

```
final class TestViewModel: ObservableObject {
    
    private var timer: Timer?
    @Published var value: Bool = false
    
    func setup() {
        timer = Timer.scheduledTimer(withTimeInterval: 1, repeats: true) { [weak self] _ in
            self?.value.toggle()
        }
    }
}

struct ChildViewAKALottieView: View {
    let value: Bool
    
    var body: some View {
        Text("Hello World")
            .onChange(of: value) { value in
                print("value changed: \(value)")
            }
    }
}

struct TestView: View {
    @EnvironmentObject private var viewModel: TestViewModel
    
    var body: some View {
        ChildViewAKALottieView(value: viewModel.value)
            .onAppear {
                viewModel.setup()
            }
    }
}
```

calda: We were able to get this to work without a binding in https://github.com/airbnb/lottie-ios/pull/2127/commits/1e777ead9c5a566606cd7a4f90b402396f8d3708

calda: Interestingly, if you take that commit and change `private var reloadAnimationTrigger: AnyEquatable?` to `@State private var reloadAnimationTrigger: AnyEquatable?` then it no longer works correctly. Adding `@State` specifically breaks it.

denandreychuk: Thank you, @calda! Appreciate your time!

Regarding `@State`, it breaks everything cause it gets initialized only once. Then, even if you update the value in the parent, as long as the view is on-screen, it will maintain the same initialized value from the first call. However, you can override this behavior using `State(initialValue:)`, but in practice, this doesn't always work as expected too.

Issue #2123, opened 2023-08-01
Looping and currentProgress creates race condition


denandreychuk:
When we use looping with 'currentProgress' modifiers, and 'currentProgress' is set to 0, the animation plays for a few seconds before pausing.

Here's the logic I'm trying to implement:

- The animation progress is set to 0.
- The user presses the play button.
- The animation starts playing.
- When the user presses the stop button, the animation pauses and reverts back to 0.

Code in context of LazyVStack with a lot of animations:

>         LottieView {
>             try await DotLottieFile.named(someAnimation)
>         }
>         .looping()
>         .resizable()
>         .loadAnimationTrigger(.constant(isPlaying))
>         .currentProgress(isPlaying ? nil : 0)

Also, is there a specific reason to use Binding for 'loadAnimationTrigger' instead of any Equatable type?

Ideally, this is what is expected from the API:
- The animation respects the 'currentProgress' modifier. If it's set to a specific value, it should disregard the looping modifier.
- If we change 'currentProgress' from a value to nil and a looping modifier is present, the animation should revert back to looping.

calda:
Thanks for the report. Could you share a small sample project that demonstrates this issue? That would be really helpful. Then I'll take a look at fixing the issue.

calda:
I think I understand your issue though so I should be able to fix it without a sample project.

calda:
Here's a fix: https://github.com/airbnb/lottie-ios/pull/2128

PR #2128, opened 2023-08-03
Add SwiftUI APIs for controlling playback behavior using `LottiePlaybackMode`

calda:
This PR adds a new `LottiePlaybackMode` type, which we can use to add SwiftUI APIs for controlling the animation's playback behavior.

https://github.com/airbnb/lottie-ios/issues/2123 revealed an issue with the previous implementation approach using `configure { ... }`. Here's an example where the `LottieView` unexpectedly doesn't respect that the playback should be unexpectedly starts playing again. This was because `looping()` and `currentProgress(...)` were both adding a separate `configure` closure, which conflicted with eachother.

Instead, we should have a single property that is the source of truth for the animation playback configuration (`playbackMode`) which is set by both of these different APIs:

```swift
public func looping() -> Self {
  var copy = self
  copy.playbackMode = .fromProgress(0, toProgress: 1, loopMode: .loop)
  return copy
}

public func currentProgress(_ currentProgress: AnimationProgressTime?) -> Self {
  guard let currentProgress = currentProgress else { return self }
  var copy = self
  copy.playbackMode = .progress(currentProgress)
  return copy
}
```

This lets consumers express the current playback mode more fluently:

```swift
private var playbackMode: LottiePlaybackMode {
  if animationPlaying {
    return .fromProgress(playFromProgress, toProgress: playToProgress, loopMode: loopMode)
  } else {
    return .progress(sliderValue)
  }
}
```

This API also gives us an easy way to support playback modes other than just `looping`, so this PR also adds full parity with all of the playback options supported by `LottieAnimationView`. 

Fixes https://github.com/airbnb/lottie-ios/issues/2123.

### DIFF
--- a/Example/Example/AnimationPreviewView.swift
+++ b/Example/Example/AnimationPreviewView.swift
@@ -65,9 +65,9 @@ struct AnimationPreviewView: View {
             animationPlaying = false
           }
         })
-        #endif
 
         Spacer(minLength: 16)
+        #endif
 
         Button {
           animationPlaying.toggle()
@@ -116,6 +116,7 @@ struct AnimationPreviewView: View {
 
   @ViewBuilder
   private var optionsMenu: some View {
+    #if !os(tvOS)
     Menu {
       Menu {
         option("Automatic", keyPath: \.renderingEngine, value: .automatic)
@@ -155,6 +156,7 @@ struct AnimationPreviewView: View {
     } label: {
       Image(systemName: "gear")
     }
+    #endif
   }
 
   private func loadAnimation() async throws -> LottieAnimationSource? {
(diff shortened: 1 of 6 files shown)


Example/Example/AnimationPreviewView.swift
```diff
@@ -85,6 +101,61 @@ struct AnimationPreviewView: View {
...
+  private var playbackMode: LottiePlaybackMode {
+    if animationPlaying {
+      return .fromProgress(playFromProgress, toProgress: playToProgress, loopMode: loopMode)
+    } else {
+      return .progress(sliderValue)
+    }
+  }
+
+  @ViewBuilder
+  private var optionsMenu: some View {
```
miguel-jimenez-0529, line 118: So cool

miguel-jimenez-0529 (approved):
LGTM

PR #2134, opened 2023-08-10
Add `animationDidFinish` API to LottieView

calda:
This PR adds an `animationDidFinish` API to LottieView. 

When using `loopMode: .playOnce`, or another finite duration animation like `.repeat(3)`, this lets you know when the animation is complete. 

For example, in the example app you can see we previously didn't update the "pause" button to the "play" state when the animation ended. Now we do:

```swift
.animationDidFinish { completed in
  if completed {
    animationPlaying = false
  }
}
```

| Before | After |
| ---- | ----- |
|  |  |

### DIFF
--- a/Sources/Public/Animation/LottieView.swift
+++ b/Sources/Public/Animation/LottieView.swift
@@ -193,7 +193,10 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
   /// when an animation finishes playing.
   public func animationDidFinish(_ animationCompletionHandler: LottieCompletionBlock?) -> Self {
     var copy = self
-    copy.animationCompletionHandler = animationCompletionHandler
+    copy.animationCompletionHandler = { completed in
+      animationCompletionHandler?(completed)
+      copy.animationCompletionHandler?(completed)
+    }
     return copy
   }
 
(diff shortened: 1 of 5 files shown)


matthewcheok (approved):

Sources/Public/Animation/LottieView.swift
```diff
@@ -189,6 +189,14 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     return copy
   }
 
+  /// Returns a copy of this view with the given `LottieCompletionBlock` that is called
+  /// when an animation finishes playing.
+  public func animationDidFinish(_ animationCompletionHandler: LottieCompletionBlock?) -> Self {
+    var copy = self
+    copy.animationCompletionHandler = animationCompletionHandler
+    return copy
+  }
```
miguel-jimenez-0529, line 198: Should we chain them?
```suggestion
  public func animationDidFinish(_ animationCompletionHandler: LottieCompletionBlock?) -> Self {
    var copy = self
    copy.animationCompletionHandler = {
      animationCompletionHandler?()
      copy.animationCompletionHandler?()
    }
    return copy
  }
```

miguel-jimenez-0529 (approved):
N I C E

Sources/Public/Animation/LottieView.swift
calda: sure, seems reasonable

PR #2138, opened 2023-08-11
Add cross-platform SwiftUI wrapper for AnimatedSwitch

calda:
This PR adds a new `LottieSwitch` component, which is a cross-platform SwiftUI wrapper for `AnimatedSwitch`. This PR also adds macOS support to `AnimatedSwitch` and `AnimatedControl`, which previously was only available on iOS / tvOS.

Before officially releasing a new version with SwiftUI support, I think it makes sense to complete our migration away from the iOS-only UIKit example app and to the multiplatform SwiftUI example app (and then delete the iOS-only example app).

The only functionality in the iOS / UIKit app that doesn't exist in the multiplatform SwiftUI app is the controls demo that lets you test `AnimatedSwitch` and `AnimatedButton`. Will add SwiftUI `AnimatedButton` in a follow-up when I have a chance.

### DIFF
--- a/Example/Example/ControlsDemoView.swift
+++ b/Example/Example/ControlsDemoView.swift
@@ -1,0 +1,61 @@
+// Created by Cal Stephens on 8/11/23.
+// Copyright © 2023 Airbnb Inc. All rights reserved.
+
+import Lottie
+import SwiftUI
+
+// MARK: - ControlsDemoView
+
+struct ControlsDemoView: View {
+
+  var body: some View {
+    List {
+      LottieSwitchRow(
+        animationName: "Samples/Switch",
+        title: "Switch",
+        onTimeRange: 0.5...1.0,
+        offTimeRange: 0.0...0.5)
+
+      LottieSwitchRow(
+        animationName: "Samples/Switch",
+        title: "Switch (Custom Colors)",
+        onTimeRange: 0.5...1.0,
+        offTimeRange: 0.0...0.5,
+        colorValueProviders: [
+          "Checkmark Outlines.Group 1.Stroke 1.Color": [Keyframe(.black)],
+          "Checkmark Outlines 2.Group 1.Stroke 1.Color": [Keyframe(.black)],
+          "X Outlines.Group 1.Stroke 1.Color": [Keyframe(.black)],
+          "Switch Outline Outlines.Fill 1.Color": [
+            Keyframe(value: LottieColor.black, time: 0),
+            Keyframe(value: LottieColor(r: 0.76, g: 0.76, b: 0.76, a: 1), time: 75),
+            Keyframe(value: LottieColor.black, time: 150),
+          ],
+        ])
+
+      // TODO: Add SwiftUI support for AnimatedButton
+      // AnimatedButtonRow(
+      //   animationName: "Samples/TwitterHeartButton",
+      //   title: "Twitter Heart Button",
+      //   playRanges: [
+      //     .init(fromMarker: "touchDownStart", toMarker: "touchDownEnd", event: .touchDown),
+      //     .init(fromMarker: "touchDownEnd", toMarker: "touchUpCancel", event: .touchUpOutside),
+      //     .init(fromMarker: "touchDownEnd", toMarker: "touchUpEnd", event: .touchUpInside),
+      //   ]))
+
+      LottieSwitchRow(
+        animationName: "Samples/Issues/issue_1877",
+        title: "Issue #1877",
+        onTimeRange: nil, // use the default (0...1)
+        offTimeRange: nil, // use the default (1...0)
+        colorValueProviders: ["**.Color": [Keyframe(.black)]])
+    }
+    .navigationTitle("Controls Demo")
+  }
+
+}
+
+extension LottieColor {
+  static var black: LottieColor {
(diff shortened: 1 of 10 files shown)


Sources/Public/Controls/AnimatedControl.swift
```diff
@@ -100,6 +109,49 @@ open class AnimatedControl: UIControl {
...
 
+  #elseif canImport(AppKit)
+  open override func mouseDown(with _: NSEvent) {
+    guard let window = window else { return }
+
+    currentState = .highlighted
+    updateForState()
+
+    // AppKit mouse-tracking loop from:
+    // https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingMouseEvents/HandlingMouseEvents.html#//apple_ref/doc/uid/10000060i-CH6-SW4
```
calda, line 120: this is really how `NSControl` works????? so weird man

Example/Example/ControlsDemoView.swift
```diff
@@ -0,0 +1,61 @@
...
+import Lottie
+import SwiftUI
+
+// MARK: - ControlsDemoView
+
+struct ControlsDemoView: View {
+
+  var body: some View {
+    List {
+      LottieSwitchRow(
```
calda, line 13: These demo cases are copied exactly from the iOS example app

Example/Example/ControlsDemoView.swift
```diff
@@ -0,0 +1,61 @@
...
+        offTimeRange: nil, // use the default (1...0)
+        colorValueProviders: ["**.Color": [Keyframe(.black)]])
+    }
+    .navigationTitle("Controls Demo")
+  }
+
+}
+
+extension LottieColor {
+  static var black: LottieColor {
```
miguel-jimenez-0529, line 58: Do we have a way to create a `LottieColor` from a `Color` or `UIColor`?

Example/Example/ControlsDemoView.swift
```diff
@@ -0,0 +1,61 @@
...
+
+      // TODO: Add SwiftUI support for AnimatedButton
+      // AnimatedButtonRow(
+      //   animationName: "Samples/TwitterHeartButton",
+      //   title: "Twitter Heart Button",
+      //   playRanges: [
+      //     .init(fromMarker: "touchDownStart", toMarker: "touchDownEnd", event: .touchDown),
+      //     .init(fromMarker: "touchDownEnd", toMarker: "touchUpCancel", event: .touchUpOutside),
+      //     .init(fromMarker: "touchDownEnd", toMarker: "touchUpEnd", event: .touchUpInside),
+      //   ]))
```
miguel-jimenez-0529, line 43: Should we remove the comments until we have the demo working?

Sources/Public/Controls/AnimatedControl.swift
miguel-jimenez-0529: Weird API :/

Sources/Public/Controls/AnimatedControl.swift
```diff
@@ -100,6 +109,49 @@ open class AnimatedControl: UIControl {
...
+      let expectedState = (isInside && keepOn) ? LottieNSLottieControlState.highlighted : .normal
+      if currentState != expectedState {
+        currentState = expectedState
+        updateForState()
+      }
+    }
+  }
+
+  @objc
+  func handleMouseUp() {
```
miguel-jimenez-0529, line 150: Should this be `open`?

Sources/Public/Controls/AnimatedSwitch.swift
```diff
@@ -176,21 +218,27 @@ open class AnimatedSwitch: AnimatedControl {
...
+    didSet {
+      stateUpdated?(_isOn)
+    }
+  }
+
   // MARK: Private
 
   private func updateAccessibilityLabel() {
-    accessibilityValue = _isOn ? NSLocalizedString("On", comment: "On") : NSLocalizedString("Off", comment: "Off")
+    let value = _isOn ? NSLocalizedString("On", comment: "On") : NSLocalizedString("Off", comment: "Off")
```
miguel-jimenez-0529, line 232: Do we provide localizations for those strings? If not, should we inject the bundle so that someone can provide their localizations?

miguel-jimenez-0529 (approved):
Good stuff!

Example/Example/ControlsDemoView.swift
calda: Not at the moment. Would probably be a good idea though.

Example/Example/ControlsDemoView.swift
calda: Nah I'll uncomment this out in the next PR

Sources/Public/Controls/AnimatedControl.swift
calda: I think we can keep this `internal` for now, it's an implementation detail.

Sources/Public/Controls/AnimatedSwitch.swift
calda: Not sure, this was here before. I think we probably don't want to change this without really understanding how changes could affect consumers.

PR #2142, opened 2023-08-15
Add support for rendering drop shadows

### LINKED ISSUE #1169: Drop Shadows not rendering
- [] This issue isn't related to another open issue

## This issue is a:
- [] Crashing Bug
- [] Feature Request

Lottie 3.0

## What Platform are you on? 

- [X] MacOS
- [] iOS

## What Language are you in?
- [] Swift
- [] Objective-C

## Expected Behavior 

Screen Shot 2020-04-03 at 9 37 25 am
Screen Shot 2020-04-03 at 9 40 47 am

## Actual Behavior

Screen Shot 2020-04-03 at 9 42 15 am

## Animation JSON

Hey there, 
I have seen other similar issues been raised about the compatibility of drop shadows with Lottie. From what I have read, this is fine and is supported, as long as the Drop Shadow effect is used, and not the layer style. I have used this, but for some reason I am still not getting the drop shadow to render. Any ideas?

Appreciate any help :)

calda:
This PR adds support for rendering drop shadows using the Core Animation rendering engine. Fixes #1169.

There are two different JSON schema representations for drop shadows. The most common option is the "layer effects" schema, but there's also a second "layer style" schema. This PR adds support for both.


matthewcheok (approved):

Sources/Private/CoreAnimation/Animations/DropShadowAnimation.swift
```diff
@@ -0,0 +1,183 @@
...
+      // `DropShadowEffect.softness` doesn't seem to map cleanly to
+      // `CALayer.cornerRadius` (e.g. with a simple multiplier).
+      // Instead, this uses a custom quadratic regression eyeballed
+      // to match the expected appearance of the start / end of the
+      // `issue_1169_shadow_effect_animated.json` sample animation:
+      //  - `softness=10` roughly corresponds to `cornerRadius=2.5`
+      //  - `softness=50` roughly corresponds to `cornerRadius=6.25`
+      // This could probably be improved with more examples.
+      let x = softnessValue.cgFloatValue
+      let cornerRadiusMapping = (-0.003 * pow(x, 2)) + (0.281 * x)
```
calda, line 78: Worked with Remington to get this to match the preview in After Effects. For example:

| Start, iOS | End, iOS |
| ---- | ---- |
| Screenshot 2023-08-15 at 3 22 13 PM | Screenshot 2023-08-15 at 3 22 16 PM |

| Start, After Effects | End, After Effects |
| ---- | ---- |
| Screenshot 2023-08-15 at 11 01 30 AM | Screenshot 2023-08-15 at 11 01 36 AM |

This was the only mapping I could find that we thought looked right in both of these cases

calda:
ptal @thedrick

Sources/Private/CoreAnimation/Animations/LayerProperty.swift
```diff
@@ -159,6 +159,34 @@ extension LayerProperty {
...
+    .init(
+      caLayerKeypath: #keyPath(CALayer.shadowRadius),
+      defaultValue: 3.0,
+      customizableProperty: nil /* currently unsupported */ )
+  }
+
+  static var shadowOffset: LayerProperty<CGSize> {
+    .init(
+      caLayerKeypath: #keyPath(CALayer.shadowOffset),
+      defaultValue: CGSize(width: 0, height: -3.0),
```
thedrick, line 187: why does this and `shadowRadius` have non-zero default values?

thedrick (approved):
LGTM @calda

Sources/Private/CoreAnimation/Animations/LayerProperty.swift
calda: This is the default value of the `CALayer.shadowOffset` property ([documentation](https://developer.apple.com/documentation/quartzcore/calayer/1410970-shadowoffset):
> The default value of this property is `(0.0, -3.0)`.

 Knowing this value lets us apply certain optimizations in the code that generates the `CAKeyframe`s. (e.g. if the Lottie animation just sets the property to the default value, we can completely omit the keyframes and reduce the overhead of computing the keyframes)

thedrick: ah great, this might be a good comment to add in code here so others aren't confused in the future!

PR #2151, opened 2023-08-22
Remove iOS-only UIKit example app in favor of multiplatform SwiftUI example app

calda:
This PR removes the iOS-only UIKit example app in favor of the multiplatform SwiftUI example app, which implements all of the same functionality but is additional available on macOS and tvOS.


calda:
ptal @miguel-jimenez-0529

miguel-jimenez-0529 (approved):
👋

PR #2152, opened 2023-08-23
Add support for visionOS, drop support for Swift 5.5

calda:
This PR adds support for using Lottie on visionOS via Swift Package Manager. Here's a gif of the example app running in the Vision Pro simulator:

The code changes to Lottie itself were fairly trivial, only requiring additional conditional complication checks in ~20 places. The trickier part of this change is continuing to support all of our CI jobs in older versions of Xcode without the visionOS SDK. 

To update CI, this PR:
 - adds visionOS builds to the relevant Rakefile task
 - adds a `SKIP_VISION_OS=true` flag to disable the visionOS build when running the rake task
 - adds new Github Actions jobs using Xcode 15, which includes visionOS builds
 - adds `SKIP_VISION_OS=true` to the existing pre-Xcode-15 CI jobs


Package.swift
```diff
@@ -1,11 +1,15 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.9
```
calda, line 1: It doesn't seem possible to reference `.visionOS("1.0")` when using `swift-tools-version:5.5`, do we have to split this into separate `Package.swift` and `Package-swift-5.5.swift` files.

calda: With some more research I figured out you can use `.custom("visionOS", versionString: "1.0")` for this, which is available in 5.6+. This required dropping support for Swift 5.5, which I think it reasonable at this point.

Package.swift
```diff
@@ -1,11 +1,14 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.6
```
calda, line 1: Dropped support for Swift 5.5, discussion here: https://github.com/airbnb/lottie-ios/pull/2152#discussion_r1303341991

.github/workflows/main.yml
```diff
@@ -33,6 +32,22 @@ jobs:
         xcode:
         - '14.2' # Swift 5.7
         - '14.3' # Swift 5.8
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/setup
+        with:
+          xcode: ${{ matrix.xcode }}
+      - name: Build Package
+        run: SKIP_VISION_OS=true bundle exec rake build:package:all
```
calda, line 41: Added `SKIP_VISION_OS=true` to the existing Xcode 13 / Xcode 14 jobs

.github/workflows/main.yml
```diff
@@ -94,9 +109,31 @@ jobs:
           # XCFrameworks are forwards-compatible but not backwards-compatible.
           # The Xcode version we use for this job is that oldest Xcode version that
           # will be able to use these XCFrameworks and the lottie-spm package.
-          xcode: '13.2.1' # Swift 5.5.2
+          xcode: '13.4.1' # Swift 5.6.1
```
calda, line 112: Bumped the XCFramework to be built by Xcode 13.4.1, the last version with Swift 5.6 (since we dropped support for Swift 5.5)

.github/workflows/main.yml
```diff
@@ -94,9 +109,31 @@ jobs:
...
+          path: .build/archives
+
+  build-xcframework-macos-13:
+    name: "Build XCFramework"
+    runs-on: macos-13
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/setup
+        with:
+          xcode: '15.0' # Swift 5.9, first version that includes visionOS SDK.
```
calda, line 134: We additionally have to build an XCFramework using Xcode 15.0 so that there is an XCFramework that supports visionOS.

.github/workflows/main.yml
```diff
@@ -120,17 +157,54 @@ jobs:
...
+        xcode:
+        - '14.3' # Swift 5.8
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ./.github/actions/setup
+        with:
+          install-mint: true
+          xcode: ${{ matrix.xcode }}
+      - name: Test Swift Package Manager support
+        run: SKIP_VISION_OS=true bundle exec rake test:spm
```
calda, line 190: Added `SKIP_VISION_OS=true` to the existing Xcode 13 / Xcode 14 jobs

thedrick (approved):
Very exciting @calda LGTM!

calda:
It seems the GitHub Actions runner doesn't include the visionOS SDK yet. Getting this error:
> { platform:visionOS, id:dvtdevice-DVTiOSDevicePlaceholder-xros:placeholder, name:Any visionOS Device, error:visionOS 1.0 is not installed. To use with Xcode, first download and install the platform }

Filed an issue here: https://github.com/actions/runner-images/issues/8144. We'll need to wait for that to be resolved.

PR #1917, opened 2023-01-17
Add support for skew animations

calda:
This PR adds support for skew animations. Fixes #1915.

| Before | After |
| ----- | ----- |
| 1915 before | 1915 after |

PR #2157, opened 2023-08-24
Fix issue where skew keyframes would unexpectedly not animate

### LINKED ISSUE #2055: The lottie animation is loaded and displayed but with no animation.
The attached json file does not animate in the sample application https://github.com/airbnb/lottie-ios/tree/master/Example/iOS.

Thank you for any help.

Lottie 4.2.0 and 4.1.3 have this issue

## Expected Behavior 
The animation should animate. 

## Actual Behavior
The animation in the attached file is loaded and displayed correctly but no animation can be observed.
image

## Animation JSON
The json file is attached.

[v13.json.zip](https://github.com/airbnb/lottie-ios/files/11355954/v13.json.zip)

calda:
This PR fixes an issue where skew keyframes in a `ShapeTransform` would unexpectedly not animate. Fixes #2055.

#1917 originally added support for rendering skew values, but the test animation from that PR didn't actually animate the skew. It looks like Core Animation doesn't animate skew values properly, which is maybe not surprising considering the skew calculation is [entirely custom](https://github.com/airbnb/lottie-ios/blob/f864891db4d04cdf141b6b35f6467d2205c45197/Sources/Private/Utility/Primitives/VectorsExtensions.swift#L253).

To fix this we can just do what the Main Thread rendering engine does, and compute the `CATransform3D` ourselves for every frame instead of letting Core Animation try to interpolate it.

| Before | After |
| ---- | ---- |
|  |  |


thedrick (approved):
LGTM @calda

PR #2164, opened 2023-08-29
Fix issue where value providers could be ignored when falling back to main thread rendering engine

### LINKED ISSUE #2163: tvOS: the key values which set with `setValueProvider` disappearing while switching between `.automatic` and `.mainThread`

Lottie 4.2.0

## Expected Behavior 
The values set by `setValueProvider` are kept internally and transferred between Animation Layers

## Actual Behavior
1. The following code is not working on tvOS because Lottie created `CoreAnimationLayer` when assigning `animation` of the view
2. In `CoreAnimationLayer.display()`, it fails to `setupAnimation()` due `The Core Animation rendering engine doesn't currently support applying trims to filled shapes`
3. create `MainThreadAnimationLayer` for fallback handling
4. the `valueProviders` are not transferred to the new animation layer

```Swift
    let savingAnimationView = LottieAnimationView()
    savingAnimationView.animation = LottieAnimation.named("CloudUpdating")
    let colorProvider = ColorValueProvider(UIColor(verkada: .charcoal).lottieColorValue)
    let keyPath = AnimationKeypath(keypath: "**.Color")
    savingAnimationView.setValueProvider(colorProvider, keypath: keyPath)
    savingAnimationView.play()
    savingAnimationView.loopMode = .loop
    addArrangedSubview(savingAnimationView)
    savingAnimationView.size(CGSize(width: 40, height: 40))
```

I have added the code at the end of `fileprivate func makeAnimationLayer(usingEngine renderingEngine: RenderingEngineOption)` method of `LottieAnimationView`, but not sure it will not cause any issue

```Swift
    valueProviders.forEach({ valueProvider in
      animationLayer.setValueProvider(valueProvider.value, keypath: valueProvider.key)
    })
```

calda:
This PR fixes an issue where custom value providers could be ignored when using `RenderingEngineOption.automatic` on an animation that is incompatible with the Core Animation render engine and instead falls back to the main thread rendering engine. Fixes #2163.

This was because `setValueProvider` was only being called on the `CoreAnimationLayer`, which was later being discarded and replaced with a `MainThreadAnimationLayer`.

Here's an example using the `Typeface/G` animation, which is incompatible with the Core Animation rendering engine and falls back to the Main Thread rendering engine. The demo uses value providers to make the "G" be red instead of blue.

| Before (value providers ignored) | After (value providers applied correctly) |
| ---- | ---- |
|  Screenshot 2023-08-29 at 10 35 08 AM | Screenshot 2023-08-29 at 10 33 31 AM |


Example/Example/AnimationPreviewView.swift
```diff
@@ -61,6 +61,11 @@ struct AnimationPreviewView: View {
           animationPlaying = false
         }
       }
+      .configure { animationView in
+        DispatchQueue.main.async {
+          currentRenderingEngine = animationView.currentRenderingEngine
+        }
```
miguel-jimenez-0529, line 67: Could we add a comment about what this is for?

miguel-jimenez-0529 (approved):
Great!

Example/Example/AnimationPreviewView.swift
calda: which part? tracking the current rendering engine in an `@State` property?

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1096,6 +1096,10 @@ public class LottieAnimationLayer: CALayer {
     }
 
     animationLayer.lottieAnimationLayer = self
+
```
calda, line 1099:
```suggestion
```

calda, line 1099:
```suggestion

```

PR #2168, opened 2023-08-30
Add option make main thread rendering engine force display update on every frame

calda:
This PR adds a workaround for a bug in the main thread rendering engine where it's possible for layer properties to be rendered with a stale, out-of-date value.

In the animation below (with colors modified for emphasis), the animation plays correctly during the first loop. But after it completes the first loop and starts playing subsequent loops, the bottom row of circles will unexpectedly flash red.

I spent a bit of time looking in to the underlying issue. This seems to be a bug in how the main thread rendering engine decides whether or not a property (in this case, a `FillNode`) needs to be re-rendered. This is pretty complex though because it involves all of the parent layers (which in this case includes multiple parent precomp layers) and a much of mutable state that gets modified every frame. I didn't see any obvious ways to fix the issue. 

As a simple workaround, we can provide the ability to make the main thread rendering engine pass in `forceUpdates: true` on every frame. This guarantees that each property will be re-rendered each frame, avoiding the bug.

| Before | With `forceDisplayUpdateOnEachFrame` |
| ---- | ---- |
|  |  |

### DIFF
--- a/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
+++ b/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
@@ -94,8 +94,8 @@ final class CoreAnimationLayer: BaseAnimationLayer {
     }
   }
 
-  /// The parent `LottieAnimationView` that manages this layer
-  weak var animationView: LottieAnimationView?
+  /// The parent `LottieAnimationLayer` that manages this layer
+  weak var lottieAnimationLayer: LottieAnimationLayer?
 
   /// A closure that is called after this layer sets up its animation.
   /// If the animation setup was unsuccessful and encountered compatibility issues,
@@ -316,7 +316,7 @@ final class CoreAnimationLayer: BaseAnimationLayer {
     else { return }
 
     if isAnimationPlaying == true {
-      animationView?.updateInFlightAnimation()
+      lottieAnimationLayer?.updateInFlightAnimation()
     } else {
       let currentFrame = currentFrame
       removeAnimations()
--- a/Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift
+++ b/Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift
@@ -140,7 +140,7 @@ final class MainThreadAnimationLayer: CALayer, RootAnimationLayer {
       newFrame = floor(newFrame)
     }
     for animationLayer in animationLayers {
-      animationLayer.displayWithFrame(frame: newFrame, forceUpdates: false)
+      animationLayer.displayWithFrame(frame: newFrame, forceUpdates: forceDisplayUpdateOnEachFrame)
     }
   }
 
@@ -149,8 +149,16 @@ final class MainThreadAnimationLayer: CALayer, RootAnimationLayer {
   /// The animatable Current Frame Property
   @NSManaged var currentFrame: CGFloat
 
-  /// The parent `LottieAnimationView` that manages this layer
-  weak var animationView: LottieAnimationView?
+  /// The parent `LottieAnimationLayer` that manages this layer
+  weak var lottieAnimationLayer: LottieAnimationLayer?
+
+  /// Whether or not to use `forceDisplayUpdate()` when rendering each individual frame.
+  ///  - The main thread rendering engine implements optimizations to decrease the amount
+  ///    of properties that have to be re-rendered on each frame. There are some cases
+  ///    where this can result in bugs / incorrect behavior, so we allow it to be disabled.
+  ///  - Forcing a full render on every frame will decrease performance, and is not recommended
+  ///    except as a workaround to a bug in the main thread rendering engine.
+  var forceDisplayUpdateOnEachFrame = false
 
   var animationLayers: ContiguousArray<CompositionLayer>
 
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -427,9 +427,6 @@ public class LottieAnimationLayer: CALayer {
   /// Will inform the receiver the type of rendering engine that is used for the layer.
   public var animationLayerDidLoad:((_ animationLayer: LottieAnimationLayer, _ renderingEngine: RenderingEngineOption) -> Void)?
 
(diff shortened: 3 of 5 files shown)


Sources/Private/CoreAnimation/CoreAnimationLayer.swift
```diff
@@ -94,8 +94,8 @@ final class CoreAnimationLayer: BaseAnimationLayer {
     }
   }
 
-  /// The parent `LottieAnimationView` that manages this layer
-  weak var animationView: LottieAnimationView?
+  /// The parent `LottieAnimationLayer` that manages this layer
+  weak var lottieAnimationLayer: LottieAnimationLayer?
```
calda, line 98: While I was looking in to this I noticed and fixed an unrelated issue:

We were passing in a reference to the parent `LottieAnimationView` so this layer could delegate some callbacks to its parent.

When using `LottieAnimationLayer` CALayer subclass on its own, there won't be a `LottieAnimationView`. We should pass in the `LottieAnimationLayer` instead, since that's where all of the functionality is implemented anyway.

Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift
```diff
@@ -149,8 +149,16 @@ final class MainThreadAnimationLayer: CALayer, RootAnimationLayer {
...
+  /// The parent `LottieAnimationLayer` that manages this layer
+  weak var lottieAnimationLayer: LottieAnimationLayer?
+
+  /// Whether or not to use `forceDisplayUpdate()` when rendering each individual frame.
+  ///  - The main thread rendering engine implements optimizations to decrease the amount
+  ///    of properties that have to be re-rendered on each frame. There are some cases
+  ///    where this can result in bugs / incorrect behavior, so we allow it to be disabled.
+  ///  - Forcing a full render on every frame will decrease performance, and is not recommended
+  ///    except as a workaround to a bug in the main thread rendering engine.
+  var forceDisplayUpdateOnEachFrame = false
```
miguel-jimenez-0529, line 161: In this context, where does `forceDisplayUpdate()` gets called? From the code it seems we are simply passing this bool to `animationLayer.displayWithFrame(frame:, forceUpdates: )`

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -427,6 +427,9 @@ public class LottieAnimationLayer: CALayer {
   /// Will inform the receiver the type of rendering engine that is used for the layer.
   public var animationLayerDidLoad:((_ animationLayer: LottieAnimationLayer, _ renderingEngine: RenderingEngineOption) -> Void)?
 
+  /// The parent `LottieAnimationView` of this layer, if applicable.
+  public internal(set) weak var animationView: LottieAnimationView?
+
```
miguel-jimenez-0529, line 432: Why do we need this, asking to learn?

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -703,9 +706,19 @@ public class LottieAnimationLayer: CALayer {
...
+  ///  - The main thread rendering engine implements optimizations to decrease the amount
+  ///    of properties that have to be re-rendered on each frame. There are some cases
+  ///    where this can result in bugs / incorrect behavior, so we allow it to be disabled.
+  ///  - Forcing a full render on every frame will decrease performance, and is not recommended
+  ///    except as a workaround to a bug in the main thread rendering engine.
+  ///  - Has no effect when using the Core Animation rendering engine.
+  public var mainThreadRenderingEngineShouldForceDisplayUpdateOnEachFrame = false {
+    didSet {
+      (rootAnimationLayer as? MainThreadAnimationLayer)?.forceDisplayUpdateOnEachFrame
+        = mainThreadRenderingEngineShouldForceDisplayUpdateOnEachFrame
```
miguel-jimenez-0529, line 720:
```suggestion
        = newValue
```

miguel-jimenez-0529 (approved):
LGTM

Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift
calda: I described it that way because `forceDisplayUpdate()` is a public API but `forceUpdates: true` is not. 

The implementation of `forceDisplayUpdate()` is:

```swift
func forceDisplayUpdate() {
  for animationLayer in animationLayers {
    animationLayer.displayWithFrame(frame: currentFrame, forceUpdates: true)
  }
}
```

and the relevant code for `forceDisplayUpdateOnEachFrame` is:

```swift

File: Sources/Public/Animation/LottieAnimationLayer.swift (hunk 3 above)
Comment (calda) [reply Sources/Public/Animation/LottieAnimationLayer.swift:432]: There was an existing public API for this:

```swift
public var animationView: LottieAnimationView? {
    set { rootAnimationLayer?.animationView = newValue }
    get { rootAnimationLayer?.animationView }
}
```

I was thinking we can't remove an existing public API due to backwards compatibility, so I just changed it to a stored property.

calda: Technically I guess we could remove this, because there hasn't been an official release that includes `LottieAnimationLayer`, so we can still make breaking changes here. 

@eromanc, are you using this for anything? I imagine not since Oragami just uses CALayers.

If there are no consumers of this we could just remove it completely.

Sources/Public/Animation/LottieAnimationLayer.swift
calda: `didSet` doesn't provide a `newValue` property, only `set` and `willSet` provide that

Sources/Public/Animation/LottieAnimationLayer.swift
eromanc: let me confirm with you tomorrow @calda but I assume we are not using that.

eromanc: Confirmed. We are not using that property in Origami.

calda: Great! In that case I'll just remove this API.

PR #2171, opened 2023-09-07
add a switch to ignore internal image cache

JerryFans:
feature: add a switch to ignore internal image cache or not , in image provider protocol, default is false. (some scene some one want to custom the image cache in custom image provider such as replace some image in image  provider, if internal image cache enable, the replace event will fail.)

### DIFF
--- a/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
+++ b/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
@@ -21,7 +21,7 @@ private final class CachedImageProvider: AnimationImageProvider {
   // MARK: Public
 
   public func imageForAsset(asset: ImageAsset) -> CGImage? {
-    if imageProvider.ignoreInternalImageCache == false, let image = imageCache.object(forKey: asset.id as NSString) {
+    if imageProvider.cacheEligible, let image = imageCache.object(forKey: asset.id as NSString) {
       return image
     }
     if let image = imageProvider.imageForAsset(asset: asset) {
--- a/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
+++ b/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
@@ -21,7 +21,7 @@ private final class CachedImageProvider: AnimationImageProvider {
   // MARK: Public
 
   public func imageForAsset(asset: ImageAsset) -> CGImage? {
-    if let image = imageCache.object(forKey: asset.id as NSString) {
+    if imageProvider.ignoreInternalImageCache == false, let image = imageCache.object(forKey: asset.id as NSString) {
       return image
     }
     if let image = imageProvider.imageForAsset(asset: asset) {
--- a/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
+++ b/Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
@@ -21,7 +21,7 @@ private final class CachedImageProvider: AnimationImageProvider {
   // MARK: Public
 
   public func imageForAsset(asset: ImageAsset) -> CGImage? {
-    if imageProvider.cacheEligible, let image = imageCache.object(forKey: asset.id as NSString) {
+    if let image = imageCache.object(forKey: asset.id as NSString) {
       return image
     }
     if let image = imageProvider.imageForAsset(asset: asset) {
@@ -42,6 +42,7 @@ extension AnimationImageProvider {
   /// It wraps the current provider as image loader, and uses `NSCache` to cache the images for resue.
   /// The cache will be reset when the `animation` is reset.
   var cachedImageProvider: AnimationImageProvider {
-    CachedImageProvider(imageProvider: self)
+    guard cacheEligible else { return self }
+    return CachedImageProvider(imageProvider: self)
   }
 }
--- a/Sources/Public/ImageProvider/AnimationImageProvider.swift
+++ b/Sources/Public/ImageProvider/AnimationImageProvider.swift
@@ -8,6 +8,8 @@
 import CoreGraphics
 import Foundation
 
+// MARK: - AnimationImageProvider
+
 /// Image provider is a protocol that is used to supply images to `LottieAnimationView`.
 ///
 /// Some animations require a reference to an image. The image provider loads and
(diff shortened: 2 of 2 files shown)


Sources/Public/ImageProvider/AnimationImageProvider.swift
```diff
@@ -17,5 +17,15 @@ import Foundation
 /// Additionally custom Image Providers can be made to load images from a URL,
 /// or to Cache images.
 public protocol AnimationImageProvider {
+    
+  var ignoreInternalImageCache: Bool { get set }
+
```
calda, line 22: Can we invert this to something like `cacheImages` which defaults to `true`?

```suggestion
  /// Whether or not the resulting image of this image provider can be cached by Lottie. Defaults to true.
  /// If true, Lottie may internally cache the result of `imageForAsset`
  var cacheImages: Bool { get set }
```

calda (approved):
lgtm other than the comment I added

calda:
You'll need to run the code formatter to update your code to follow Airbnb's Swift Style Guide. Locally you can run `bundle exec rake format:swift` or `swift package format`.

Sources/Public/ImageProvider/AnimationImageProvider.swift
calda: Open to other naming suggestions as well. I'd just like to describe this in some way similar to "whether or not Lottie is permitted to cache the result of this image provider"

JerryFans: > Can we invert this to something like `cacheEligible` which defaults to `true`?

Of course, the reason why I set the default false is not to affect the old version, if you think it is better, I think it is OK.

JerryFans: > > Can we invert this to something like `cacheEligible` which defaults to `true`?
> 
> Of course, the reason why I set the default false is not to affect the old version, if you think it is better, I think it is OK.



> Open to other naming suggestions as well. I'd just like to describe this in some way similar to "whether or not Lottie is permitted to cache the result of this image provider"

Of course, if you need to default to true, cacheEligible description is more accurate.

JerryFans:
> 



> You'll need to run the code formatter to update your code to follow Airbnb's Swift Style Guide. Locally you can run `bundle exec rake format:swift` or `swift package format`.

Ok, I try it.

JerryFans:
Do I modify it as described above, then format it, and then commit it again?

calda:
Yeah

JerryFans:
I have committed again. Thank you.

Sources/Public/ImageProvider/AnimationImageProvider.swift
```diff
@@ -17,5 +19,17 @@ import Foundation
 /// Additionally custom Image Providers can be made to load images from a URL,
 /// or to Cache images.
 public protocol AnimationImageProvider {
+
+  /// Whether or not the resulting image of this image provider can be cached by Lottie. Defaults to true.
+  /// If true, Lottie may internally cache the result of `imageForAsset`
+  var cacheEligible: Bool { get set }
+
   func imageForAsset(asset: ImageAsset) -> CGImage?
```
calda, line 27: We may as well add a doc comment here for consistency:

```suggestion
  /// The image to display for the given `ImageAsset` defined in the `LottieAnimation` JSON file.
  func imageForAsset(asset: ImageAsset) -> CGImage?
```

Sources/Public/ImageProvider/AnimationImageProvider.swift
```diff
@@ -17,5 +19,18 @@ import Foundation
 /// Additionally custom Image Providers can be made to load images from a URL,
 /// or to Cache images.
 public protocol AnimationImageProvider {
+
+  /// Whether or not the resulting image of this image provider can be cached by Lottie. Defaults to true.
+  /// If true, Lottie may internally cache the result of `imageForAsset`
+  var cacheEligible: Bool { get set }
```
calda, line 25: I don't think we want this to be mutable

```suggestion
  var cacheEligible: Bool { get }
```

Sources/Public/ImageProvider/AnimationImageProvider.swift
```diff
@@ -17,5 +19,18 @@ import Foundation
...
+
+  /// The image to display for the given `ImageAsset` defined in the `LottieAnimation` JSON file.
   func imageForAsset(asset: ImageAsset) -> CGImage?
 }
+
+extension AnimationImageProvider {
+  public var cacheEligible: Bool {
+    get { true }
+    set { }
+  }
```
calda, line 35:
```suggestion
  public var cacheEligible: Bool {
    true
  }
```

Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
```diff
@@ -21,7 +21,7 @@ private final class CachedImageProvider: AnimationImageProvider {
   // MARK: Public
 
   public func imageForAsset(asset: ImageAsset) -> CGImage? {
-    if let image = imageCache.object(forKey: asset.id as NSString) {
+    if imageProvider.cacheEligible, let image = imageCache.object(forKey: asset.id as NSString) {
```
calda, line 24: Right now this is unnecessarily caching images in the `imageCache` which are never used.

Instead of creating a `CachedImageProvider` but just not using it, perhaps instead we should update the `var cachedImageProvider` implementation below:

```swift
extension AnimationImageProvider {
  /// Create a cache enabled image provider which will reuse the asset image with the same asset id
  /// It wraps the current provider as image loader, and uses `NSCache` to cache the images for resue.
  /// The cache will be reset when the `animation` is reset.
  /// If this image provider is not eligible for caching, simply returns self.
  var cachedImageProvider: AnimationImageProvider {
    guard cacheEligibile else { return self }
    return CachedImageProvider(imageProvider: self)
  }
}
```

JerryFans: But I think other built-in providers, like FilepathImageProvider. Having a memory cache is necessary, and we should leave the choice to the dev user to handle some special scenarios.

JerryFans:
I have submitted it again. Thank you for your review suggestion.

Sources/Private/MainThread/LayerContainers/Utility/CachedImageProvider.swift
calda: Not sure what you mean about `FilepathImageProvider`. In that case `imageProvider.cacheEligible` will be `true` so there will be no behavior change no matter what.

I think we shouldn't be writing images to the cache in the case where `imageProvider.cacheEligible` is `false`. Those cached images will not be used, but writing them to the cache consumes additional memory.

JerryFans: Yeah,I looked at the source code carefully, and indeed if cacheEligibile is not true , Only return our Custom AnimationImageProvider it's ok, your solution is feasible.

JerryFans:
> Thank you!

You're welcome . Thanks also for your review suggestion. Good Night.

PR #2175, opened 2023-09-08
Improve conversion from After Effects shadow softness value to `CALayer.shadowRadius` value

calda:
This PR fixes an issue where the previous math converting from an After Effects shadow softness value to a `CALayer.shadowRadius` was producing results that were very incorrect.

When rendering a drop shadow, we can't directly apply the softness value from After Effects to the `layer.shadowRadius` field. If we do this, it produces shadows that are much softer than they appear in after effects.

I had previously tried to hand-tune a sample animation to perfectly match after effects, which resulted in a weird quadratic equation. This turns out to be over-turned and not general, because it results in junk results for some other values (e.g. in this case below was converting `120` to `-11`).

Instead, we should just use a simple "divide by X" conversion. This should be good enough without being overfit to any specific test case. I worked with Remington to identify that dividing by 5 seems to work best in the most cases (and most importantly, this actual non-sample animation below).


calda:
ptal @thedrick

thedrick (approved):
LGTM @calda

PR #2177, opened 2023-09-11
Add contents gravity configuration

matthewcheok:
This PR adds an option for setting the `contentsGravity` inside of `AnimationImageProvider`


Sources/Public/ImageProvider/AnimationImageProvider.swift
```diff
@@ -26,10 +27,18 @@ public protocol AnimationImageProvider {
...
 
 extension AnimationImageProvider {
   public var cacheEligible: Bool {
     true
   }
+
+  /// The default value is `.resize`, similar to that of `CALayer`.
+  public func contentsGravity(for asset: ImageAsset) -> CALayerContentsGravity? {
+    nil
+  }
```
calda, line 43: Maybe it would be better for this to be non-optional, with the default defined like this:

```suggestion
  public func contentsGravity(for asset: ImageAsset) -> CALayerContentsGravity {
    .resize
  }
```

Sources/Public/ImageProvider/AnimationImageProvider.swift
```diff
@@ -26,10 +27,18 @@ public protocol AnimationImageProvider {
 
   /// The image to display for the given `ImageAsset` defined in the `LottieAnimation` JSON file.
   func imageForAsset(asset: ImageAsset) -> CGImage?
+
+  /// Specifies how the layer's contents are positioned or scaled within its bounds for a given asset.
```
calda, line 31:
```suggestion
  /// Specifies how the layer's contents are positioned or scaled within its bounds for a given asset.
  /// Defaults to `.resize`, which stretches the image to fill the layer.
```

Sources/Private/CoreAnimation/Layers/ImageLayer.swift
```diff
@@ -48,6 +48,7 @@ final class ImageLayer: BaseCompositionLayer {
 
     self.imageAsset = imageAsset
     contentsLayer.contents = image
+    contentsLayer.contentsGravity = context.imageProvider.contentsGravity(for: imageAsset) ?? .resize
```
calda, line 51: We should also make this change to the `ImageCompositionLayer` used by the Main Thread rendering engine

calda (approved):
Thank you!

Sources/Public/iOS/FilepathImageProvider.swift
```diff
@@ -18,16 +18,20 @@ public class FilepathImageProvider: AnimationImageProvider {
...
 
-  public init(filepath: URL) {
+  public init(filepath: URL, contentsGravity: CALayerContentsGravity = .resize) {
     self.filepath = filepath
+    self.contentsGravity = contentsGravity
   }
 
   // MARK: Public
 
+  public var contentsGravity: CALayerContentsGravity
```
calda, line 33: This is a class so we probably want this to be internal and immutable

```suggestion
  let contentsGravity: CALayerContentsGravity
```

Sources/Public/iOS/BundleImageProvider.swift
```diff
@@ -24,13 +24,16 @@ public class BundleImageProvider: AnimationImageProvider {
...
-  public init(bundle: Bundle, searchPath: String?) {
+  public init(bundle: Bundle, searchPath: String?, contentsGravity: CALayerContentsGravity = .resize) {
     self.bundle = bundle
     self.searchPath = searchPath
+    self.contentsGravity = contentsGravity
   }
 
   // MARK: Public
 
+  public var contentsGravity: CALayerContentsGravity
```
calda, line 35: same here

```suggestion
  let contentsGravity: CALayerContentsGravity
```

Issue #1949, opened 2023-02-10
CoreAnimation rendering engine changed the behavior of Text providers


JoeSzymanski:
In testing the new verison of Lottie, I ran into an issue where custom text providers were not working in automatic mode but were working in mainThread mode.

By adding a log message in DictionaryTextProvider.textFor(keypathName: String, sourceText: String), I noticed the following:

Automatic mode:
JRS - custom text for keypathName ENVELOPE-FRONT.sender_username, sourceText: sender_username
JRS - custom text for keypathName ENVELOPE-FRONT.From, sourceText: From:
JRS - custom text for keypathName ENVELOPE-BACK-TEXTBOX.LETTER-TEXTBOX.sender_username, sourceText: sender_username
JRS - custom text for keypathName ENVELOPE-BACK-TEXTBOX.LETTER-TEXTBOX.custom_text, sourceText: custom_text

MainThread mode:
JRS - custom text for keypathName sender_username, sourceText: sender_username
JRS - custom text for keypathName From, sourceText: From:
JRS - custom text for keypathName custom_text, sourceText: custom_text
JRS - custom text for keypathName sender_username, sourceText: sender_username
JRS - custom text for keypathName From, sourceText: From:
JRS - custom text for keypathName custom_text, sourceText: custom_text
JRS - custom text for keypathName sender_username, sourceText: sender_username

It looks like the new CoreAnimation enginer is using a different definition for the keypathName parameter, which means it's not possible to safely support both modes.

## Which Version of Lottie are you using?

Lottie 4.1.2

## Animation JSON
Working to confirm that I can share the JSON file. I will update once I have approval.

calda:
Yeah, the two rendering engines have different `AnimationKeypath` implementations which are different in some edge cases that we aren't handling correctly yet. Please share your animation json when you're able to!

JoeSzymanski:
The json for the animation is attached.

[love_letter.json.zip](https://github.com/airbnb/lottie-ios/files/10710730/love_letter.json.zip)

PR #2183, opened 2023-09-12
Update text provider API to use full AnimationKeypath values

### LINKED ISSUE #1588: AnimationTextProvider should provide the full keypath

## This issue is a:
- [ ] Non-Crashing Bug (Visual or otherwise)
- [ ] Crashing Bug
- [x] Feature Request

Lottie 3.0

## What Platform are you on? 

- [ ] MacOS
- [x] iOS

## What Language are you in?
- [x] Swift
- [ ] Objective-C

## Expected Behavior 

`AnimationTextProvider` should provide the full keypath. IMO we should remove/deprecate the existing method and add a new one that provides the full keypath instead.

## Actual Behavior

`AnimationTextProvider` calls `textFor(keypathName:sourceText:)` with only the last component.

calda:
This PR replaces the `AnimationTextProvider` API with a new `AnimationKeypathTextProvider` API that uses full `AnimationKeypath` values. 
 - Fixes #1588. 
 - Fixes https://github.com/airbnb/lottie-ios/issues/1949

`AnimationKeypath` represents a layer's full path in the animation. For example, a text layer could have an `AnimationKeypath` like `PARENT_A.PARENT_B.text_value`. 

The old `AnimationTextProvider` API was implemented inconsistently between the two rendering engines:
 - The Main Thread rendering engine always called the text provider with the last component of the text layer keypath (so just `text_value`).
 - The Core Animation rendering engine always called the text provider with the entire keypath (so `PARENT_A.PARENT_B.text_value`).

Before we even introduced the Core Animation rendering engine, #1588 had previously suggested updating the Main Thread rendering engine to always use the full keypath name. This seems like the best solution, but we just have to make the change in a backwards-compatible way.

This PR introduces a new `AnimationKeypathTextProvider` API, which takes the entire keypath and has the same behavior across both rendering engines:

```swift
/// Protocol for providing dynamic text to for a Lottie animation.
public protocol AnimationKeypathTextProvider: AnyObject {
  /// The text to display for the given `AnimationKeypath`.
  /// If `nil` is returned, continues using the existing default text value.
  func text(for keypath: AnimationKeypath, sourceText: String) -> String?
}
```

For backwards compatibility we preserve the existing `AnimationTextProvider` API, with the existing inconsistent behavior, but renamed to `LegacyAnimationTextProvider`:

```swift
/// Legacy protocol for providing dynamic text for a Lottie animation.
/// Instead prefer conforming to `AnimationKeypathTextProvider`.
@available(*, deprecated, message: """
  `AnimationKeypathTextProvider` has been deprecated and renamed to `LegacyAnimationTextProvider`. \
  Instead, conform to `AnimationKeypathTextProvider` instead or conform to `LegacyAnimationTextProvider` explicitly.
  """)
public typealias AnimationTextProvider = LegacyAnimationTextProvider

// MARK: - LegacyAnimationTextProvider

/// Legacy protocol for providing dynamic text for a Lottie animation.
/// Instead prefer conforming to `AnimationKeypathTextProvider`.
public protocol LegacyAnimationTextProvider {
  /// Legacy method to look up the text to display for the given keypath.
  /// Instead, prefer implementing `AnimationKeypathTextProvider.`
  /// The behavior of this method depends on the current rendering engine:
  ///  - The Core Animation rendering engine always calls this method
  ///    with the full keypath (e.g. `MY_LAYER.text_value`).
  ///  - The Main Thread rendering engine always calls this method
  ///    with the final component of the key path (e.g. just `text_value`).
  func textFor(keypathName: String, sourceText: String) -> String
}
```

In the next major version of Lottie (Lottie 5.0), we could completely delete `LegacyAnimationTextProvider`.

This PR includes pretty extensive tests capturing the behavior of `AnimationKeypathTextProvider`, `LegacyAnimationTextProvider`, and `DictionaryTextProvider` in both rendering engines.


Sources/Private/MainThread/LayerContainers/CompLayers/PreCompositionLayer.swift
```diff
@@ -69,6 +77,7 @@ final class PreCompositionLayer: CompositionLayer {
     childKeypaths.append(contentsOf: layers)
 
     layerImageProvider.addImageLayers(imageLayers)
+    layerTextProvider.addTextLayers(textLayers)
```
calda, line 80: Found and fixed an unrelated bug where the Main Thread rendering engine didn't correctly track and update text layers included in `PreCompositionLayer`s.

Sources/Private/MainThread/LayerContainers/CompLayers/TextCompositionLayer.swift
```diff
@@ -92,8 +98,18 @@ final class TextCompositionLayer: CompositionLayer {
...
-  var textProvider: AnimationTextProvider
+  var textProvider: AnimationKeypathTextProvider
   var fontProvider: AnimationFontProvider
+  weak var rootAnimationLayer: MainThreadAnimationLayer?
+
+  lazy var fullAnimationKeypath: AnimationKeypath = {
+    // Individual layers don't know their full keypaths, so we have to delegate
+    // to the `MainThreadAnimationLayer` to search the layer hierarchy and find
+    // the full keypath (which includes this layer's parent layers)
+    rootAnimationLayer?.keypath(for: self)
```
calda, line 109: This was tricky -- I suppose this is why the Main Thread rendering engine didn't previously work like this

Sources/Private/Utility/Extensions/AnimationKeypathExtension.swift
```diff
@@ -138,6 +144,30 @@ extension KeypathSearchable {
 
     return allKeypaths
   }
+
+  /// Computes the list of animation keypaths that descend from this layer
+  func layerKeypaths(for keyPath: AnimationKeypath? = nil) -> [CALayer: AnimationKeypath] {
```
calda, line 149: Adapted from the `allKeypaths` method above, but just different enough that it didn't seam feasible to consolidate the logic into a single helper

Sources/Public/DynamicProperties/AnimationKeypath.swift
```diff
@@ -44,6 +46,14 @@ public struct AnimationKeypath: Hashable, ExpressibleByStringLiteral {
     self.keys = keys
   }
 
-  var keys: [String]
+  // MARK: Public
+
+  /// The dot-separated key values that represent this keypath.
+  public internal(set) var keys: [String]
```
calda, line 52: Added some more public helpers to this type since it has a bigger role in the public API of `AnimationKeypathTextProvider` now.

miguel-jimenez-0529 (approved):
LGTM

alisamili:
在3.3.0,使用DictionaryTextProvider(replacements)，替换国际化的多语言key，在阿语下文本内容仍然展示的是英文能解决吗？

alisamili:
> let replacements = ["#textnodea": getStringWithKey("KEY_xxxxa"), "#textnodeb": getStringWithKey("KEY_xxxxb")] 在3.3.0中，使用DictionaryTextProvider(replacements)，替换国际化多语言键。在使用阿拉伯语的国家，文本内容仍然显示为英文。这个版本能解决这个问题吗？
> 
> 在3.3.0中，使用DictionaryTextProvider（替换），替换国际化的多语言键，在使用阿语的国家下，文本内容仍然显示的是英文，该版本能解决这个问题吗？

Issue #2161, opened 2023-08-28
[iOS] App is crashing if you use CompatibleAnimationView in xib/storyboard


Saratsin:
## Which Version of Lottie are you using?

Lottie 4.2.0 - iOS

## Steps to reproduce

1. Create a simple Xcode project with storyboard UI and Objective-C as a language
2. Add Lottie reference in any way (e.g. as xcframework reference)
3. Add `UIView` element the main xib file or storyboard, change its type to `CompatibleAnimationView`
4. Build and run project on simulator or device

## Expected Behavior 
App is launching and shows the main screen without a problems

## Actual Behavior
App is launching and crashes on the main screen:
```
Lottie/CompatibleAnimationView.swift:197: Fatal error: init(coder:) has not been implemented
2023-08-28 19:06:27.015025+0300 LottieTest[84664:923278] Lottie/CompatibleAnimationView.swift:197: Fatal error: init(coder:) has not been implemented
```

## Reason:

Actually it is described in the exception message. `CompatibleAnimationView` doesn't implement required `init` method

calda:
@Saratsin, are you interested in making a PR to fix this? That would be much appreciated! 

Perhaps this init needs to be made public? https://github.com/airbnb/lottie-ios/blob/master/Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift#L196C1-L198C4

If not, could you share a sample Xcode project that demonstrates the error?

Saratsin:
> @Saratsin, are you interested in making a PR to fix this? That would be much appreciated!
> 
> Perhaps this init needs to be made public? https://github.com/airbnb/lottie-ios/blob/master/Sources/Public/iOS/Compatibility/CompatibleAnimationView.swift#L196C1-L198C4
> 
> If not, could you share a sample Xcode project that demonstrates the error?

Sure, I can create a PR with fix for it. Do I have to update version to `4.2.1` in `podsec` and `package.json` files, or just add the fix?

calda:
If you share a sample xcode project to demonstrate the error you can just point to 4.2.0. If you want to fix the error and then validate that the sample project now works correctly you'd need to edit the code locally, in the cocoapod checkouts.

calda:
@Saratsin, are you able to share a sample project that demonstrates this issue?

Saratsin:
[...]

PR #2192, opened 2023-09-15
Initialization fix for xib files

Saratsin:
xib uses init method with NSCoder parameter.
Fix is pretty simple: just add initialization logic for this method instead of throwing fatal error


calda (approved):
Thanks!

calda:
Fixes https://github.com/airbnb/lottie-ios/issues/2161

Issue #2193, opened 2023-09-17
[iOS] Animation stretches out from the top left AnimationView corner since Lottie 4.3.0


aureliencolas:
## Issue
Since Lottie 4.3.0, when the animation starts, it is most of the time stretched from the top left corner of its AnimationView to the entier AnimationView (according to the AnimationView contentMode).
The issue described here is a regression introduced with Lottie 4.3.0 as it is not reproduced with Lottie 4.2.0 and earlier.

## Which Version of Lottie are you using?
Lottie 4.3.0 iOS SDK from SPM.

## Expected Behavior 
The animation fills the AnimationView right at the beginning of the animation. No additional stretching animation occurs.

With Lottie 4.2.0, this is the case:
https://github.com/airbnb/lottie-ios/assets/82930766/4dd31b8a-006e-4c1b-a8c1-e63b7cd86cd9

## Actual Behavior
Since Lottie 4.3.0, the animation sometimes starts by streching from the top left corner of its AnimationView to the entire AnimationView.

Here is the issue with Lottie 4.3.0:
https://github.com/airbnb/lottie-ios/assets/82930766/e969896c-97cc-4e98-9fa7-12e341e17249

## Animation JSON
I manage multiple apps and the issue appears with all of them. Here is one of these JSON:
[data-ipad_110.json.txt](https://github.com/airbnb/lottie-ios/files/12642930/data-ipad_110.json.txt)

## Implementation details
In case it is relevant, here is how the AnimationView is added to a UIViewController:
```swift
private let animationView = LottieAnimationView()

override func viewDidLoad() {
  super.viewDidLoad()

  animationView.translatesAutoresizingMaskIntoConstraints = false
  view.addSubview(animationView)
  animationView.constrain(to: view) // helper to constrain the for view edges to another view
  animationView.backgroundBehavior = .forceFinish
  animationView.backgroundColor = UIColor(named: "content_background1")
  animationView.contentMode = .scaleAspectFit
  animationView.isHidden = true
}
```
Then, when the animation has to start, here is how it is started:
```swift
animationView.isHidden = false
animationView.animation = animation
animationView.play(completion: completion)
```

I tried to initialise the AnimationView with UIView.performWithAnimation() but with no luck. Actually, the AnimationView itself is already constrain when it appears; it is the animation itself which stretches out, not the AnimationView.

calda:
Can you share a zip of sample Xcode project that demonstrates this issue? Thanks!

[...]

PR #2194, opened 2023-09-18
Fix layout animation regression introduced in 4.3.0

calda:
This PR fixes a regression introduced by #2183 where the frame of the animation layer could unexpectedly animate.

I don't have repro steps for this, but I worked with @aureliencolas to identify the change that introduced the regression: https://github.com/airbnb/lottie-ios/issues/2193#issuecomment-1722586996

#2183 added a `display()` call to `CoreAnimationLayer.forceDisplayUpdate()`, which appears to be the culprit here. We can revert that change and implement the behavior needed for those tests a different way.

Here's a gif from the issue showing the regression, which is fixed by this change:


calda:
ptal @thedrick

PR #2196, opened 2023-09-19
Reduce log volume from warning about After Effects expressions

calda:
This PR reduces the log volume created by our warnings about unsupported After Effects expressions.

https://github.com/airbnb/lottie-ios/discussions/2191 mentioned that these logs are a bit excessive. This has also come up within Airbnb, where this is creating quite a lot of noise in our CI jobs. 

Instead of printing this for each instance of AE expressions, we can instead just print it once per animation. We can also remove the code from the log message, since the code is potentially very long and doesn't actually contribute much to the log message.

https://github.com/airbnb/lottie-ios/pull/2196/commits/4f88cbefe3e214c1e6ab165d1c6aaba77da00169 sets up snapshot tests for our log output.

https://github.com/airbnb/lottie-ios/pull/2196/commits/a9628a3afad1acb588b4bf79932dc1dc9acde898 then actually changes the log output for After Effects expressions, and updates the log output snapshot. You can see much less is logged in the diff for that commit: 

image


miguel-jimenez-0529 (approved):
LGTM. It is so cool that we snapshot our logs 🎉

calda:
I love the point free snapshot testing library, especially for text output. Super easy!

PR #2198, opened 2023-09-25
Allow pausing at marker

matthewcheok:
This change adds support for pausing playback at markers.

Thought on restructuring the `PlaybackMode` enum? This is a bigger change which IMO could clean things up quite a bit (we could leave current APIs in-place marked as deprecated until the next major version bump):
```
enum LottiePlaybackMode {

  case paused(Paused)

  enum Paused {
    case currentFrame
    case progress(AnimationProgressTime)
    case frame(AnimationFrameTime)
    case timestamp(TimeInterval)
    case marker(_ name: String, position: MarkerPosition)
  }

  case play(PlayRange)

  enum PlayRange {
    case progress(_ from: AnimationProgressTime?, to: AnimationProgressTime, loopMode: LottieLoopMode)
    case frames(_ from: AnimationFrameTime?, to: AnimationFrameTime, loopMode: LottieLoopMode)
    case marker(_ name: String, loopMode: LottieLoopMode)
    case markers(_ names: [String])
  }

  @available(*, deprecated, message: "Will be removed in a future major release.")
  case progress(_ progress: AnimationProgressTime)

  @available(*, deprecated, message: "Will be removed in a future major release.")
  case frame(_ frame: AnimationFrameTime)

  @available(*, deprecated, message: "Will be removed in a future major release.")
  case time(_ time: TimeInterval)

  @available(*, deprecated, message: "Will be removed in a future major release.")
  case pause

  @available(*, deprecated, message: "Will be removed in a future major release.")
  case fromProgress(_ fromProgress: AnimationProgressTime?, toProgress: AnimationProgressTime, loopMode: LottieLoopMode)

  @available(*, deprecated, message: "Will be removed in a future major release.")
  case fromFrame(_ fromFrame: AnimationFrameTime?, toFrame: AnimationFrameTime, loopMode: LottieLoopMode)

  @available(*, deprecated, message: "Will be removed in a future major release.")
  case marker(_ marker: String, loopMode: LottieLoopMode)

  @available(*, deprecated, message: "Will be removed in a future major release.")
  case markers(_ markers: [String])
}
```

### DIFF
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -107,7 +107,7 @@ public class LottieAnimationLayer: CALayer {
     guard let animation = animation else { return }
 
     defer {
-      currentPlaybackMode = .play(.progress(to: 1, loopMode: loopMode))
+      currentPlaybackMode = .play(.fromProgress(nil, toProgress: 1, loopMode: loopMode))
     }
 
     if shouldOverrideWithReducedMotionAnimation {
@@ -139,7 +139,7 @@ public class LottieAnimationLayer: CALayer {
     guard let animation = animation else { return }
 
     defer {
-      currentPlaybackMode = .play(.progress(from: fromProgress, to: toProgress, loopMode: loopMode ?? self.loopMode))
+      currentPlaybackMode = .play(.fromProgress(fromProgress, toProgress: toProgress, loopMode: loopMode ?? self.loopMode))
     }
 
     if shouldOverrideWithReducedMotionAnimation {
@@ -172,7 +172,7 @@ public class LottieAnimationLayer: CALayer {
     completion: LottieCompletionBlock? = nil)
   {
     defer {
-      currentPlaybackMode = .play(.frames(from: fromFrame, to: toFrame, loopMode: loopMode ?? self.loopMode))
+      currentPlaybackMode = .play(.fromFrame(fromFrame, toFrame: toFrame, loopMode: loopMode ?? self.loopMode))
     }
 
     if shouldOverrideWithReducedMotionAnimation {
@@ -216,9 +216,9 @@ public class LottieAnimationLayer: CALayer {
     completion: LottieCompletionBlock? = nil)
   {
     defer {
-      currentPlaybackMode = .play(.markers(
-        from: fromMarker,
-        to: toMarker,
+      currentPlaybackMode = .play(.fromMarker(
+        fromMarker,
+        toMarker: toMarker,
         playEndMarkerFrame: playEndMarkerFrame,
         loopMode: loopMode ?? self.loopMode))
     }
@@ -311,7 +311,7 @@ public class LottieAnimationLayer: CALayer {
     guard !markers.isEmpty else { return }
 
     defer {
-      currentPlaybackMode = .play(.markerList(markers))
+      currentPlaybackMode = .play(.markers(markers))
     }
 
     if shouldOverrideWithReducedMotionAnimation {
@@ -371,7 +371,7 @@ public class LottieAnimationLayer: CALayer {
 
     case .frame(let animationFrameTime):
       currentFrame = animationFrameTime
-    case .timestamp(let timeInterval):
+    case .time(let timeInterval):
       currentTime = timeInterval
 
     case .marker(let name, let position):
@@ -399,6 +399,7 @@ public class LottieAnimationLayer: CALayer {
   }
 
   /// Applies the given `LottiePlaybackMode` to this layer.
+  /// - Parameter playbackMode: The playback mode to apply
   /// - Parameter completion: A closure that is called after
   ///   an animation triggered by this method completes.
   open func setPlaybackMode(
@@ -419,45 +420,49 @@ public class LottieAnimationLayer: CALayer {
       pause(at: .frame(frame))
 
     case .time(let time):
-      pause(at: .timestamp(time))
+      pause(at: .time(time))
 
     case .pause:
       pause(at: .currentFrame)
 
     case .fromProgress(let from, let to, let loopMode):
-      play(.progress(from: from, to: to, loopMode: loopMode), completion: completion)
+      play(.fromProgress(from, toProgress: to, loopMode: loopMode), completion: completion)
 
     case .fromFrame(let from, let to, let loopMode):
-      play(.frames(from: from, to: to, loopMode: loopMode), completion: completion)
+      play(.fromFrame(from, toFrame: to, loopMode: loopMode), completion: completion)
 
     case .fromMarker(let from, let to, let playEndMarkerFrame, let loopMode):
-      play(.markers(from: from, to: to, playEndMarkerFrame: playEndMarkerFrame, loopMode: loopMode), completion: completion)
+      play(.fromMarker(from, toMarker: to, playEndMarkerFrame: playEndMarkerFrame, loopMode: loopMode), completion: completion)
 
     case .marker(let name, let loopMode):
       play(.marker(name, loopMode: loopMode), completion: completion)
 
     case .markers(let names):
-      play(.markerList(names), completion: completion)
+      play(.markers(names), completion: completion)
     }
   }
 
-  open func play(_ mode: LottiePlaybackMode.PlaybackMode, completion: LottieCompletionBlock? = nil) {
-    switch mode {
-    case .progress(from: let from, to: let to, loopMode: let loopMode):
+  /// Applies the given `LottiePlaybackMode` to this layer.
+  /// - Parameter playbackMode: The playback mode to apply
+  /// - Parameter completion: A closure that is called after
+  ///   an animation triggered by this method completes.
+  open func play(_ playbackMode: LottiePlaybackMode.PlaybackMode, completion: LottieCompletionBlock? = nil) {
+    switch playbackMode {
+    case .fromProgress(let from, let to, let loopMode):
       play(
         fromProgress: from,
         toProgress: to,
         loopMode: loopMode,
         completion: completion)
 
-    case .frames(from: let from, to: let to, loopMode: let loopMode):
+    case .fromFrame(let from, let to, let loopMode):
       play(
         fromFrame: from,
         toFrame: to,
         loopMode: loopMode,
         completion: completion)
 
-    case .markers(let from, let to, let playEndMarkerFrame, let loopMode):
+    case .fromMarker(let from, let to, let playEndMarkerFrame, let loopMode):
       play(
         fromMarker: from,
         toMarker: to,
@@ -468,7 +473,7 @@ public class LottieAnimationLayer: CALayer {
     case .marker(let name, loopMode: let loopMode):
       play(marker: name, loopMode: loopMode, completion: completion)
 
-    case .markerList(let names):
+    case .markers(let names):
       play(markers: names, completion: completion)
     }
   }
@@ -681,7 +686,7 @@ public class LottieAnimationLayer: CALayer {
     set {
       if let animation = animation {
         currentFrame = animation.frameTime(forTime: newValue)
-        currentPlaybackMode = .paused(at: .timestamp(newValue))
+        currentPlaybackMode = .paused(at: .time(newValue))
       } else {
         currentFrame = 0
       }
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -345,6 +345,25 @@ public class LottieAnimationLayer: CALayer {
     })
   }
 
+  /// Pauses the animation at a given marker and position
+  open func pause(at marker: String, position: LottiePlaybackMode.MarkerPosition)
+  {
+    guard let from = animation?.markerMap?[marker] else {
+      return
+    }
+
+    defer {
+      currentPlaybackMode = .pausedAtMarker(marker, position: position)
+    }
+
+    switch position {
+    case .start:
+      currentTime = from.frameTime
+    case .end:
+      currentTime = from.frameTime + from.durationFrameTime
+    }
+  }
+
   /// Stops the animation and resets the layer to its start frame.
   ///
   /// The completion closure will be called with `false`
@@ -382,6 +401,9 @@ public class LottieAnimationLayer: CALayer {
 
     case .pause:
       pause()
+
+    case .pausedAtMarker(let marker, let position):
+      pause(at: marker, position: position)
 
     case .fromProgress(let fromProgress, let toProgress, let loopMode):
       play(
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -107,7 +107,7 @@ public class LottieAnimationLayer: CALayer {
     guard let animation = animation else { return }
 
     defer {
-      currentPlaybackMode = .play(.fromProgress(nil, toProgress: 1, loopMode: loopMode))
+      currentPlaybackMode = .playing(.fromProgress(nil, toProgress: 1, loopMode: loopMode))
     }
 
     if shouldOverrideWithReducedMotionAnimation {
@@ -139,7 +139,7 @@ public class LottieAnimationLayer: CALayer {
     guard let animation = animation else { return }
 
     defer {
-      currentPlaybackMode = .play(.fromProgress(fromProgress, toProgress: toProgress, loopMode: loopMode ?? self.loopMode))
+      currentPlaybackMode = .playing(.fromProgress(fromProgress, toProgress: toProgress, loopMode: loopMode ?? self.loopMode))
     }
 
     if shouldOverrideWithReducedMotionAnimation {
@@ -172,7 +172,7 @@ public class LottieAnimationLayer: CALayer {
     completion: LottieCompletionBlock? = nil)
   {
     defer {
-      currentPlaybackMode = .play(.fromFrame(fromFrame, toFrame: toFrame, loopMode: loopMode ?? self.loopMode))
+      currentPlaybackMode = .playing(.fromFrame(fromFrame, toFrame: toFrame, loopMode: loopMode ?? self.loopMode))
     }
 
     if shouldOverrideWithReducedMotionAnimation {
@@ -216,7 +216,7 @@ public class LottieAnimationLayer: CALayer {
     completion: LottieCompletionBlock? = nil)
   {
     defer {
-      currentPlaybackMode = .play(.fromMarker(
+      currentPlaybackMode = .playing(.fromMarker(
         fromMarker,
         toMarker: toMarker,
         playEndMarkerFrame: playEndMarkerFrame,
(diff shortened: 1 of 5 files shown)


Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -345,6 +345,25 @@ public class LottieAnimationLayer: CALayer {
     })
   }
 
+  /// Pauses the animation at a given marker and position
+  open func pause(at marker: String, position: LottiePlaybackMode.MarkerPosition)
+  {
```
calda, line 350:
```suggestion
  open func pause(at marker: String, position: LottiePlaybackMode.MarkerPosition) {
```

Sources/Public/Animation/LottiePlaybackMode.swift
```diff
@@ -7,6 +7,13 @@ import Foundation
 
 /// Configuration for how a Lottie animation should be played
 public enum LottiePlaybackMode: Hashable {
+
+  /// The position within a marker.
+  public enum MarkerPosition: Hashable {
```
calda, line 12: Can this be a top-level `LottieMarkerPosition` type below the `LottiePlaybackMode` declaration?

Sources/Public/Animation/LottiePlaybackMode.swift
```diff
@@ -20,6 +27,9 @@ public enum LottiePlaybackMode: Hashable {
   /// Any existing animation will be paused at the current frame.
   case pause
 
+  /// Pauses the animation at a given marker and position
+  case pausedAtMarker(_ marker: String, position: MarkerPosition)
```
calda, line 31: I'm not totally sure about the `pausedAt` prefix considering we don't use that on the other specific-frame cases. Perhaps instead we call this `startFrameOfMarker(_:)` and `endFrameOfMarker(_:)`?

calda: > Thought on restructuring the PlaybackMode enum? This is a bigger change which IMO could clean things up quite a bit (we could leave current APIs in-place marked as deprecated until the next major version bump):

I see the appeal of this sort of change from a semantic perspective. 

I'm especially interested how this API looks at the call-site. For example, I generally prefer:

```swift
.play(.fromProgress(0.5, toProgress: 1)

.play(.frame(currentFrame))
```

over nested forms like:

```swift
.play(.play(.fromProgress(0.5, toProgress: 1))

.play(.pause(.frame(currentFrame)))
```

calda: I could imagine splitting `LottiePlaybackMode` into two separate types so you use `.play(...)` for the modes that trigger playback and `.pause(...)` for the modes that don't. I'd be interested in seeing how that affects the syntax for cases like `AnimationPreviewView` where you dynamically switch between playing and paused using an `@State var`.

.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.1
```
calda, line 1: Not sure what this is for. Maybe we should add it to the .gitignore?

Sources/Public/Animation/LottiePlaybackMode.swift
matthewcheok: Are you referring to the `CALayer` methods?

@calda I'd imagine we'd add the following in order to unnest the APIs - what do you think?
```
func pause(_ paused: LottiePlaybackMode.Paused)
func play(_ playRange: LottiePlaybackMode.PlayRange)
```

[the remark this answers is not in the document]
calda: maybe it would be better to just have separate `startOfMarker(_:)` and `endOfMarker(_:)` cases

calda: Yeah I don't think I mind this change. I acually kind of like the extra nesting here:

Before:

```swift
  private var playbackMode: LottiePlaybackMode {
    if animationPlaying {
      return .fromProgress(playFromProgress, toProgress: playToProgress, loopMode: loopMode)
    } else {
      return .progress(sliderValue)
    }
  }
```

After:

```swift
  private var playbackMode: LottiePlaybackMode {
    if animationPlaying {
      return .play(.fromProgress(playFromProgress, toProgress: playToProgress, loopMode: loopMode))
    } else {
      return .paused(at: .progress(sliderValue))
    }
  }
```

matthewcheok: We already have `LottieView.playbackMode(_ playbackMode: LottiePlaybackMode)` which takes the root enum. Is this what you might be looking for?

matthewcheok:
@calda Updated the PR, refactoring `LottiePlaybackMode`. I wasn't sure if we should leave the `play` methods on `LottieAnimationLayer` - I think there's no harm keeping them because it helps to break up the `play(_:)` method but I don't think we can change the to `private` with `@Available`.

Sources/Public/Animation/LottiePlaybackMode.swift
```diff
@@ -3,115 +3,155 @@
...
+    /// Pauses the animation at a given marker and position
+    case marker(_ name: String, position: LottieMarkerPosition = .start)
+  }
+
+  public enum PlaybackMode: Hashable {
+    /// Plays the animation from a progress (0-1) to a progress (0-1).
+    /// - Parameter from: The start progress of the animation. If `nil` the animation will start at the current progress.
+    /// - Parameter to: The end progress of the animation.
+    /// - Parameter loopMode: The loop behavior of the animation.
+    case progress(from: AnimationProgressTime? = nil, to: AnimationProgressTime, loopMode: LottieLoopMode)
```
calda, line 79: Can we leave the names of these as-is? I prefer `.play(.fromProgress(0, toProgress: 1))` over `.play(.progress(from: 1, to: 2))`. This reads more fluently and also matches the naming of the UIKit APIs on `LottieAnimationView`

Sources/Public/Animation/LottiePlaybackMode.swift
```diff
@@ -3,115 +3,155 @@
...
+    /// If a marker is not found, it will be skipped.
+    ///
+    /// If a marker doesn't have a duration value, it will play with a duration of 0
+    /// (effectively being skipped).
+    ///
+    /// If another animation is played (by calling any `play` method) while this
+    /// marker sequence is playing, the marker sequence will be cancelled.
+    ///
+    /// - Parameter names: The list of markers to play sequentially.
+    case markerList(_ names: [String])
```
calda, line 128: Can we keep this as-is?

```suggestion
    case markers(_ names: [String])
```

Sources/Public/Animation/LottiePlaybackMode.swift
calda:
```suggestion
    case fromProgess(_ fromProgress: AnimationProgressTime? = nil, toProgress: AnimationProgressTime, loopMode: LottieLoopMode)
```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -357,62 +357,119 @@ public class LottieAnimationLayer: CALayer {
...
+
+    case .marker(let name, let loopMode):
+      play(.marker(name, loopMode: loopMode), completion: completion)
+
+    case .markers(let names):
+      play(.markerList(names), completion: completion)
+    }
+  }
+
+  open func play(_ mode: LottiePlaybackMode.PlaybackMode, completion: LottieCompletionBlock? = nil) {
```
calda, line 444: Can you add a doc comment for this method

Sources/Public/Animation/LottiePlaybackMode.swift
```diff
@@ -3,115 +3,155 @@
 
 import Foundation
 
+// MARK: - LottieMarkerPosition
+
+/// The position within a marker.
+public enum LottieMarkerPosition: Hashable {
```
calda, line 9: Could you move this to the bottom of the file below `LottiePlaybackMode`? Ideally `LottiePlaybackMode` would be the first type in this file

Sources/Public/Animation/LottieView.swift
```diff
@@ -175,27 +175,36 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
   }
 
-  /// Returns a copy of this view playing with the given `LottiePlaybackMode`
+  @available(*, deprecated, renamed: "playbackMode(_:)", message: "Will be removed in a future major release.")
   public func play(_ playbackMode: LottiePlaybackMode) -> Self {
     self.playbackMode(playbackMode)
   }
 
+  /// Returns a copy of this view playing with the given playback mode
+  public func playing(_ playbackMode: LottiePlaybackMode.PlaybackMode = .progress(to: 1, loopMode: .playOnce)) -> Self {
```
calda, line 199: Before we had `play()`, `play(loopMode:)` and `play(_:)` (which took a `LottiePlaybackMode`)

Can we keep all three of those methods, without using a default parameter? That lets us tailor the doc comment to exactly describe what each method does (e.g. that `playing()` causing the animation to play once from the start to the end)

Sources/Public/Animation/LottieView.swift
```diff
@@ -175,27 +175,36 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+    self.playbackMode(.play(playbackMode))
+  }
+
+  /// Returns a copy of this view paused with the given state
+  public func paused(at state: LottiePlaybackMode.PausedState = .currentFrame) -> Self {
```
calda, line 204: May be nice to have separate `func paused()` and `func paused(at:)` so we can tailor the doc comment for the default case

calda (commented):
Thanks for these improvements! Left some suggestions on the new naming / organization

Sources/Public/Animation/LottiePlaybackMode.swift
```diff
@@ -7,111 +7,163 @@ import Foundation
 
 /// Configuration for how a Lottie animation should be played
 public enum LottiePlaybackMode: Hashable {
-  /// The animation is paused at the given progress value,
-  /// a value between 0.0 (0% progress) and 1.0 (100% progress).
+
+  case paused(at: PausedState)
+
+  case play(_ mode: PlaybackMode)
```
calda, line 13: should this be `case playing` to match `LottieView.playing`?

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -357,62 +357,124 @@ public class LottieAnimationLayer: CALayer {
...
+
+  /// Applies the given `LottiePlaybackMode` to this layer.
+  /// - Parameter playbackMode: The playback mode to apply
+  /// - Parameter completion: A closure that is called after
+  ///   an animation triggered by this method completes.
+  open func setPlaybackMode(
+    _ playbackMode: LottiePlaybackMode,
+    completion: LottieCompletionBlock? = nil)
   {
     switch playbackMode {
```
calda, line 409: Instead could this method be simplified to:

```swift
switch playbackMode {
case .playing(let playingMode):
  play(playbackMode, animationCompletionHandler: animationCompletionHandler)
case .paused(let pausedState):
  pause(at: pausedState, animationCompletionHandler: animationCompletionHandler)
}
```

Sources/Public/Animation/LottiePlaybackMode.swift
```diff
@@ -7,111 +7,163 @@ import Foundation
 
 /// Configuration for how a Lottie animation should be played
 public enum LottiePlaybackMode: Hashable {
-  /// The animation is paused at the given progress value,
-  /// a value between 0.0 (0% progress) and 1.0 (100% progress).
+
+  case paused(at: PausedState)
```
calda, line 11:
```suggestion
  /// The animation is paused at the given state (e.g. paused at a specific frame)
  case paused(at: PausedState)
```

Sources/Public/Animation/LottiePlaybackMode.swift
calda, line 13:
```suggestion
  /// The animation is playing using the given playback mode (e.g. looping from the start to the end)
  case playing(_ mode: PlaybackMode)
```

Sources/Public/Animation/LottiePlaybackMode.swift
```diff
@@ -7,111 +7,163 @@ import Foundation
...
 extension LottiePlaybackMode {
-  /// Plays the animation from the current progress to a progress value (0-1).
-  /// - Parameter toProgress: The end progress of the animation.
-  /// - Parameter loopMode: The loop behavior of the animation.
+  public static var paused: Self {
+    .paused(at: .currentFrame)
+  }
+
+  @available(*, deprecated, renamed: "LottiePlaybackMode.play(_:)", message: "Will be removed in a future major release.")
   public static func toProgress(_ toProgress: AnimationProgressTime, loopMode: LottieLoopMode) -> LottiePlaybackMode {
```
calda, line 143: Can we keep non-deprecated `LottiePlaybackMode.PlaybackMode.toProgress`, `.toFrame`, and `.toMarker` helpers? It looks like those were deprecated without a replacement

Sources/Public/Animation/LottieView.swift
```diff
@@ -175,27 +175,56 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     return copy
   }
 
-  // Returns a copy of this view playing once from the current frame to the end frame
+  @available(*, deprecated, renamed: "playing(_:)", message: "Will be removed in a future major release.")
```
calda, line 178:
```suggestion
  @available(*, deprecated, renamed: "playing()", message: "Will be removed in a future major release.")
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -175,27 +175,56 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
   public func play(_ playbackMode: LottiePlaybackMode) -> Self {
     self.playbackMode(playbackMode)
   }
 
+  /// Returns a copy of this view playing with the given playback mode
+  public func playing(
+    _ playbackMode: LottiePlaybackMode
+      .PlaybackMode = .fromProgress(nil, toProgress: 1, loopMode: .playOnce))
+    -> Self
+  {
```
calda, line 203:
```suggestion
  public func playing(
    _ playbackMode: LottiePlaybackMode.PlaybackMode = .fromProgress(nil, toProgress: 1, loopMode: .playOnce))
    -> Self
  {
```

calda: SwiftFormat is funny sometimes

Sources/Public/Animation/LottieView.swift
```diff
@@ -175,27 +175,56 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+    self.playbackMode(.play(playbackMode))
+  }
+
+  /// Returns a copy of this view playing from the current frame to the end frame,
+  /// with the given `LottiePlaybackMode`.
+  public func playing(loopMode: LottieLoopMode = .playOnce) -> Self {
```
calda, line 209: I think as-is`, just `playing()` would be ambiguous (is it referring to the `playing` method that takes a defaulted `LottiePlaybackMode` or a defaulted `LottieLoopMode`? Instead of using default values on these can we define a standalone `func playing()`?

calda, line 209: I think as-is `playing()` would be ambiguous if this method had a default argument

```suggestion
  public func playing(loopMode: LottieLoopMode) -> Self {
```

Sources/Public/Animation/LottieView.swift
```diff
@@ -175,27 +175,56 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
-  /// Returns a copy of this view playing with the given `LottiePlaybackMode`
+  @available(*, deprecated, renamed: "playbackMode(_:)", message: "Will be removed in a future major release.")
   public func play(_ playbackMode: LottiePlaybackMode) -> Self {
     self.playbackMode(playbackMode)
   }
 
+  /// Returns a copy of this view playing with the given playback mode
+  public func playing(
+    _ playbackMode: LottiePlaybackMode
+      .PlaybackMode = .fromProgress(nil, toProgress: 1, loopMode: .playOnce))
```
calda, line 201: I think as-is `playing()` would be ambiguous if this method had a default argument

```suggestion
      .PlaybackMode
```

Sources/Public/Animation/LottieView.swift
calda: I think we also don't need the default: https://github.com/airbnb/lottie-ios/pull/2198/files#r1336439481

Sources/Public/Animation/LottieView.swift
```diff
@@ -175,27 +175,56 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+  /// Returns a copy of this view paused with the given state
+  public func paused(at state: LottiePlaybackMode.PausedState = .currentFrame) -> Self {
+    playbackMode(.paused(at: state))
+  }
+
+  /// Returns a copy of this view paused with the current frame
+  public func paused() -> Self {
+    playbackMode(.paused(at: .currentFrame))
+  }
+
```
calda, line 227: I think this one isn't necessary, because:
 1. `LottieView(.myAnimation).paused()` would be redundant
 2. I can't think of a use case where you would use this for any sort of dynamic behavior, unless you can think of one

```suggestion
```

calda: If you can think of a use case for this then we can keep it

Sources/Public/Animation/LottieView.swift
```diff
@@ -175,27 +175,56 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+    playbackMode(.play(.fromProgress(nil, toProgress: 1, loopMode: loopMode)))
+  }
+
+  // Returns a copy of this view playing once from the current frame to the end frame
+  public func playing() -> Self {
+    playbackMode(.play(.fromProgress(nil, toProgress: 1, loopMode: .playOnce)))
+  }
+
+  /// Returns a copy of this view paused with the given state
+  public func paused(at state: LottiePlaybackMode.PausedState = .currentFrame) -> Self {
```
calda, line 219: Same here -- I think having this be defaulted would make just `paused()` ambigious

```suggestion
  public func paused(at state: LottiePlaybackMode.PausedState) -> Self {
```

calda (approved):
Thanks! I think this is a good improvement. Appreciate you iterating on this.

Sources/Public/Animation/LottieAnimationLayer.swift
matthewcheok: Not unless we fully remove all the other cases...

Sources/Public/Animation/LottiePlaybackMode.swift
matthewcheok: @calda Should these be moved to `LottiePlaybackMode.PlaybackMode`?

Sources/Public/Animation/LottieAnimationLayer.swift
calda: Ahh right

Sources/Public/Animation/LottiePlaybackMode.swift
calda: Yeah

calda (approved):
Looks great, thank you!

PR #2202, opened 2023-09-28
Fix unexpected crossfade animation when updating Lottie animation or rendering engine

### LINKED ISSUE #2200: Motion trails since 4.3.0
Hi all 👋 

We've noticed lagging/trails/tracing on the dots in this animation when they move together after full pin entry. 
The animation is two separate Lotties cycling from one to the other.
The problem *becomes noticeable here: ecdadec88b057e484b6a11a01454e7af343e3203
Our configuration of the view is unaltered.

4.3.2

## Expected Behavior 

https://github.com/airbnb/lottie-ios/assets/104841860/8a5c7949-5bda-42dc-aedd-54444da0ecb1

## Actual Behavior

https://github.com/airbnb/lottie-ios/assets/104841860/aecccbf3-a805-42de-b4e9-85eec29d3799

## Animation JSON

[stageThreeBottomToTop.json.zip](https://github.com/airbnb/lottie-ios/files/12737759/stageThreeBottomToTop.json.zip)

calda:
This PR fixes an issue where updating the animation displayed by a `LottieAnimationView`, or its current rendering engine, would unexpectedly trigger a short crossfade animation. 

This was a default implicit animation added by Core Animation, which we can disable by disabling it in `actions`.

This was a regression introduced by the architecture change that created the `LottieAnimationLayer` `CALayer` subclass (#2073). I think the `CALayer` of a `UIView` has all of these implicit actions disabled by default, so moving this code from a `UIView` subclass to a `CALayer` subclass changed the behavior of inserting a sublayer.

Fixes #2200:

| Before | After |
| ----- | ----- |
|  |  |

This also fixes an issue I had noticed in the example app where the view would animate strangely when updating the rendering engine:

### After

fyi @eromanc in case you find this interesting / helpful

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ To install Lottie using [Swift Package Manager](https://github.com/apple/swift-p
 or you can add the following dependency to your `Package.swift`:
 
 ```swift
-.package(url: "https://github.com/airbnb/lottie-spm.git", from: "4.3.2")
+.package(url: "https://github.com/airbnb/lottie-spm.git", from: "4.3.3")
 ```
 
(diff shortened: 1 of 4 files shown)


calda:
ptal @thedrick

thedrick (approved):
LGTM @calda sorry I thought I had already stamped this!

calda:
nppppp thanks for the review

PR #2208, opened 2023-10-10
Fix parsing regression in 4.3.0 from addition of parsing layer effects

### LINKED ISSUE #2206: Unable to download lottie from network source

Lottie 4.3.2

## Expected Behavior 
The lottie should load up and play when fetched over a network

## Actual Behavior
The lottie doesn't show up, seems it's not downloading

## Animation JSON
Here is the lottie url- https://www.jsonkeeper.com/b/QFW5

[lottie-downloader.zip](https://github.com/airbnb/lottie-ios/files/12828120/lottie-downloader.zip)

calda:
This PR fixes a parsing regression where an animation that was parsed successfully by 4.2.0 fails to parse in 4.3.0. This was because we now parse layer effects, which previously weren't parsed so could be invalid.

The fix is to be more generous with how we handle invalid layer effects. Instead of failing to parse the entire animation, we can just ignore the specific invalid effect. 

Fixes #2206.

| Before | After |
| ---- | ----- |
| Screenshot 2023-10-10 at 10 56 59 AM | Screenshot 2023-10-10 at 10 56 28 AM |


calda:
ptal @thedrick

thedrick (approved):
LGTM @calda

PR #2214, opened 2023-10-18
Remove old animation layer when creating a new animation layer

junjielu:
This PR fixes an issue in 4.3.0 where old animation layers were still rendering images after removing animation. This is because the old animation layer is still working in memory because it is not released.

The fix is to release the old animation layer when removing it from the superlayer.


calda (approved):
Thanks!

Issue #2215, opened 2023-10-18
UIScreen.main will be deprecated in a future version of iOS


hyun99999:
there are UIScreen.main that will be deprecated.

```swift
//  LottieAnimationViewBase.swift
  var screenScale: CGFloat {
    #if os(iOS) || os(tvOS)
    UIScreen.main.scale
    #else // if os(visionOS)
    // We intentionally don't check `#if os(visionOS)`, because that emits
    // a warning when building on Xcode 14 and earlier.
    1.0
    #endif
  }
```

PR #2216, opened 2023-10-18
Replace `UIScreen.main` to get display scale on iOS 13.0 and later (#2215)

hyun99999:
`UIScree.main` will be deprecated in a future version of iOS.
I used `window.windowScene.screen` to get display's scale.

developer doc > [mian](https://developer.apple.com/documentation/uikit/uiscreen/1617815-main) already deprecated. It causes serious errors in the near future.

So have to change another way. And `window.windowScene.screen` is the one way of getting display's scale. That is available from `iOS 13.0`. This is why control flow has `iOS 13.0` condition.

Resolved: #2215

### DIFF
--- a/Sources/Public/iOS/LottieAnimationViewBase.swift
+++ b/Sources/Public/iOS/LottieAnimationViewBase.swift
@@ -40,9 +40,9 @@ open class LottieAnimationViewBase: UIView {
   var screenScale: CGFloat {
     #if os(iOS) || os(tvOS)
     if #available(iOS 13.0, tvOS 13.0, *) {
-        return window?.windowScene?.screen.scale ?? 1.0
+      return window?.windowScene?.screen.scale ?? 1.0
     } else {
-        return UIScreen.main.scale
+      return UIScreen.main.scale
     }
     #else // if os(visionOS)
     // We intentionally don't check `#if os(visionOS)`, because that emits
--- a/Sources/Public/iOS/LottieAnimationViewBase.swift
+++ b/Sources/Public/iOS/LottieAnimationViewBase.swift
@@ -94,8 +94,8 @@ open class LottieAnimationViewBase: UIView {
     // Implemented by subclasses.
   }
 
-  // MARK: - Private
-  
+  // MARK: Private
+
   private var _screenScale: CGFloat = .zero
 }
 #endif
--- a/Sources/Public/iOS/LottieAnimationViewBase.swift
+++ b/Sources/Public/iOS/LottieAnimationViewBase.swift
@@ -40,7 +40,7 @@ open class LottieAnimationViewBase: UIView {
   var screenScale: CGFloat {
     #if os(iOS) || os(tvOS)
     if #available(iOS 13.0, tvOS 13.0, *) {
-        return window?.windowScene?.screen.scale ?? .zero
+        return window?.windowScene?.screen.scale ?? 1.0
     } else {
         return UIScreen.main.scale
     }
--- a/Sources/Public/iOS/LottieAnimationViewBase.swift
+++ b/Sources/Public/iOS/LottieAnimationViewBase.swift
@@ -24,11 +24,6 @@ open class LottieAnimationViewBase: UIView {
   public override func didMoveToWindow() {
     super.didMoveToWindow()
     animationMovedToWindow()
-    #if os(iOS) || os(tvOS)
-    if #available(iOS 13.0, tvOS 13.0, *) {
-      screenScale = window?.windowScene?.screen.scale ?? 1.0
-    }
-    #endif
   }
 
   public override func layoutSubviews() {
@@ -43,22 +38,18 @@ open class LottieAnimationViewBase: UIView {
   }
 
   var screenScale: CGFloat {
-    get {
-      #if os(iOS) || os(tvOS)
-      if #available(iOS 13.0, tvOS 13.0, *) {
-        return window?.windowScene?.screen.scale ?? 1.0
-      } else {
-        return UIScreen.main.scale
-      }
-      #else // if os(visionOS)
-      // We intentionally don't check `#if os(visionOS)`, because that emits
-      // a warning when building on Xcode 14 and earlier.
-      1.0
-      #endif
+    #if os(iOS) || os(tvOS)
+    if #available(iOS 13.0, tvOS 13.0, *) {
+        let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
+        return windowScene?.screen.scale ?? 1.0
+    } else {
+      return UIScreen.main.scale
     }
-    set {
-      _screenScale = newValue
-    }
+    #else // if os(visionOS)
+    // We intentionally don't check `#if os(visionOS)`, because that emits
+    // a warning when building on Xcode 14 and earlier.
+    1.0
+    #endif
   }
 
   func layoutAnimation() {
@@ -93,9 +84,5 @@ open class LottieAnimationViewBase: UIView {
   func animationWillEnterForeground() {
     // Implemented by subclasses.
   }
-
-  // MARK: Private
-
-  private var _screenScale: CGFloat = .zero
 }
 #endif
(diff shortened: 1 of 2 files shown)


Sources/Public/iOS/LottieAnimationViewBase.swift
```diff
@@ -39,7 +39,11 @@ open class LottieAnimationViewBase: UIView {
 
   var screenScale: CGFloat {
     #if os(iOS) || os(tvOS)
-    UIScreen.main.scale
+    if #available(iOS 13.0, *) {
+        return window?.windowScene?.screen.scale ?? .zero
```
calda, line 43: Is there a way to provide a reasonable fallback, similar to `UIScreen.main.scale`, rather than just `.zero`? Returning `.zero` for some reason would result in a very broken experience.

Sources/Public/iOS/LottieAnimationViewBase.swift
```diff
@@ -39,7 +39,11 @@ open class LottieAnimationViewBase: UIView {
 
   var screenScale: CGFloat {
     #if os(iOS) || os(tvOS)
-    UIScreen.main.scale
+    if #available(iOS 13.0, *) {
```
calda, line 42: We should also check for tvOS 13+

```suggestion
    if #available(iOS 13.0, tvOS 13.0) {
```

hyun99999: That's what I missed! Thank you🙂

Sources/Public/iOS/LottieAnimationViewBase.swift
hyun99999: `UIScreen.main.scale` has several values taht each device display.

>  For Retina displays, the scale factor may be 3.0 or 2.0 and one point can represented by nine or four pixels, respectively. For standard-resolution displays, the scale factor is 1.0 and one point equals one pixel
> https://developer.apple.com/documentation/uikit/uiscreen/1617836-scale

Not all devices had a Retina display for the iPhone. This is expected complex control flow.

If there noting special happens, a view will have `window.windowScene`. However, it seems reasonable to set it to `1.0` to ensure minimal operation through the scale factor.

hyun99999: - Related commit history : be10a85

calda (approved):

Sources/Public/iOS/LottieAnimationViewBase.swift
```diff
@@ -39,7 +39,11 @@ open class LottieAnimationViewBase: UIView {
 
   var screenScale: CGFloat {
     #if os(iOS) || os(tvOS)
-    UIScreen.main.scale
+    if #available(iOS 13.0, tvOS 13.0, *) {
+        return window?.windowScene?.screen.scale ?? 1.0
```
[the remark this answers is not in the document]
hyun99999: I agree with you.

Sources/Public/iOS/LottieAnimationViewBase.swift
```diff
@@ -39,7 +39,11 @@ open class LottieAnimationViewBase: UIView {
...
+    } else {
+        return UIScreen.main.scale
+    }
```
calda, line 46: The linter check is failing because these blocks are indented with 4 spaces instead of 2. You can run the linter yourself with `bundle exec rake format:swift`.

```suggestion
    if #available(iOS 13.0, tvOS 13.0, *) {
      return window?.windowScene?.screen.scale ?? 1.0
    } else {
      return UIScreen.main.scale
    }
```

hyun99999 (commented):
Thank you for your review so far.

calda:
This change has reduced the render quality of text in the snapshot tests regression suite. Here's an example:

| Before | After |
| image | image |

That will need to be fixed before we can merge this, since there could be other consumers using a render configuration similar to our snapshot test suite.

hyun99999:
> Currently `screenScale` is only accessed in `commonInit()`, which is called during `init()`. At this point `view.window` is always `nil`, so `screenScale` will always be `1.0`.
> 
> If we're going to access `view.window`, then we need to handle the case where the `window` changes, e.g. by rerunning any logic that depends on the `window` or `screenScale` in `didMoveToWindow`.

Implements `screenScale` setter to re-run action that depend on the `screenScale` in `didMoveToWindow` 
Related comiit is 5cf3555.

hyun99999:
hmm.. window property has `nil`, until a view is added to window.
Would it be difficult to change `Build Setting > Require Only App-Extension-Safe API`value to `NO`?
 That can use `UIApplication.shared`. And this is can access to `scale` property.
@calda

calda:
I don't think the change you made fixed the problem, because `LottieAnimationView` itself is accessing `screenScale` in `init` (before the view has a window) and then never accesses it again. We'll need to make it so `LottieAnimationView` itself reacts to `screenScale` updates.

calda:
> Require Only App-Extension-Safe APIvalue to NO?

Hmmmmm, this probably would be a breaking change for some number of consumers. If there a way for us to #if def the check, so we use the simpler code _unless_ in an app extension?

calda:
What would the code look like using `UIApplication.shared`?

hyun99999:
> What would the code look like using `UIApplication.shared`?

Like this!

```swift
let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
windowScene?.screen.scale
```

hyun99999:
> > Require Only App-Extension-Safe APIvalue to NO?
> 
> Hmmmmm, this probably would be a breaking change for some number of consumers. Is there a way for us to #if def the check, so we use the simpler code _unless_ in an app extension?

I don't understand '#if def the check'. Perdon?

calda:
I was thinking something like this: https://stackoverflow.com/questions/25048026/how-to-detect-if-code-is-running-in-main-app-or-app-extension-target but it sounds like this won't be possible for us since (1) we don't control the custom build flags, and (2) we distribute pre-compiled binaries.

hyun99999:
> I was thinking something like this: https://stackoverflow.com/questions/25048026/how-to-detect-if-code-is-running-in-main-app-or-app-extension-target but it sounds like this won't be possible for us since (1) we don't control the custom build flags, and (2) we distribute pre-compiled binaries. So I don't think we can use `UIApplication.shared` unfortunately.

Recently i access screenScale in `didMoveToWindow()` but it is not apply layer screen scale in `commonInit()`. I agree your opinion.
Then i'll have to look for it that `LottieAnimationView` updates `screenScale` itself. After view is added that it has window.

Sources/Public/iOS/LottieAnimationViewBase.swift
```diff
@@ -39,7 +39,11 @@ open class LottieAnimationViewBase: UIView {
 
   var screenScale: CGFloat {
     #if os(iOS) || os(tvOS)
-    UIScreen.main.scale
+    if #available(iOS 13.0, tvOS 13.0, *) {
+      return UITraitCollection.current.displayScale
```
calda, line 43: From the documentation:

> A value of 1.0 indicates a non-Retina display and a value of 2.0 indicates a Retina display. The default display scale for a trait collection is 0.0 (indicating unspecified).

Can you manually check for the case where `displayScale` is zero, and make sure we always return at least `1`?

```suggestion
      return max(UITraitCollection.current.displayScale, 1)
```

Also, could you verify that `UITraitCollection.current.displayScale` returns `3.0` on a modern 3x Retina display? The documentation doesn't mention this explicitly, but may simply be outdated.

hyun99999: I already have checked that returns 3.0 in 3x Retina display.

CC.https://www.ios-resolution.com

hyun99999: I agree to outdated doc.
Will the document be updated, if visionOS beta is officially adopted?
I'm wondering like you!

calda (approved):

calda:
Thanks!

hyun99999:
Thank you for your sincere review! 🙌

PR #2218, opened 2023-10-18
Add configuration setting to remove animated bounds changes

thedrick:
## Changes
This adds a new configuration option to disable animating position and transform changes alongside the current context. This fixes a bug in the Airbnb app where before this change the animation's transform would be animated causing a noticeable visual glitch in some screens.

## Please review
@calda

### DIFF
--- a/Sources/Public/Animation/LottieAnimationView.swift
+++ b/Sources/Public/Animation/LottieAnimationView.swift
@@ -368,6 +368,13 @@ open class LottieAnimationView: LottieAnimationViewBase {
 
   // MARK: Public
 
+  /// Whether or not transform and position changes of the view should animate alongside
+  /// any existing animation context.
+  ///  - Defaults to `true` which will grab the current animation context and animate position and
+  ///    transform changes matching the current context's curve and duration.
+  ///    `false` will cause transform and position changes to happen unanimated
+  public var animateLayoutChangesWithCurrentCoreAnimationContext = true
+
   /// The configuration that this `LottieAnimationView` uses when playing its animation
   public var configuration: LottieConfiguration {
     get { lottieAnimationLayer.configuration }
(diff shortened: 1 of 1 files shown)


Sources/Public/Configuration/LottieConfiguration.swift
```diff
@@ -44,4 +46,11 @@ public struct LottieConfiguration: Hashable {
   /// The color space to be used for rendering
   ///  - Defaults to `CGColorSpaceCreateDeviceRGB()`
   public var colorSpace: CGColorSpace
+
+  /// Whether or not transform and position changes of the view should animate alongside
+  /// any existing animation context.
+  ///  - Defaults to `true` which will grab the current animation context and animate position and
+  ///    transform changes matching the current context's curve and duration.
+  ///    `false` will cause transform and position changes to happen unanimated
+  public var animatesLayoutWithinCurrentContext: Bool
```
calda, line 55: Can this be an instance property on `LottieAnimationView` instead of a property in `LottieConfiguration`?

calda: `LottieConfiguration` is mostly for properties that we need to be able to configure globally for the whole app. I imagine in this case we'd rather configure this flag for the specific `LottieAnimationView` that is encountering this bug.

calda, line 55: What do you think about a name like this instead? I like the idea of using the word "inherit", and explicitly mentioning that this is coming from the set of Core Animation animations.

```suggestion
  public var layoutChangesInheritCurrentCoreAnimationContext: Bool
```

calda: `animateLayoutChangesWithCurrentCoreAnimationContext` makes sense too. "Current Context" by itself seems a bit ambiguous.

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -103,7 +103,8 @@ open class LottieAnimationView: LottieAnimationViewBase {
     textProvider: AnimationKeypathTextProvider = DefaultTextProvider(),
     fontProvider: AnimationFontProvider = DefaultFontProvider(),
     configuration: LottieConfiguration = .shared,
-    logger: LottieLogger = .shared)
+    logger: LottieLogger = .shared,
+    animateLayoutChangesWithCurrentCoreAnimationContext: Bool = true)
```
calda, line 107: Rather than adding this as an argument to `init`, can we just expose the `public var` for it? A good example of this pattern is the `mainThreadRenderingEngineShouldForceDisplayUpdateOnEachFrame` property added in https://github.com/airbnb/lottie-ios/pull/2168

calda (approved):
lgtm thanks!

thedrick:
Thanks @calda !

PR #2219, opened 2023-10-19
Change  pod dependencies `SwiftUI` and `Combine` to `weak` to ensure compatibility with iOS 12

florianrhein:
Since [SwiftUI was added](https://github.com/airbnb/lottie-ios/commit/5189094838e7fadfec2bd1466aaddabcbb73742c) to lottie-ios, it no longer runs on iOS 12 devices when embedded through CocoaPods. The runtime error encountered on iOS 12:

```
dyld: Library not loaded: /System/Library/Frameworks/SwiftUI.framework/SwiftUI
Referenced from: /private/var/containers/Bundle/Application/[...]/Frameworks/Lottie.framework/Lottie
Reason: image not found
```

The reason for this is, that CocoaPods automatically adds strong dependencies to any used frameworks, unless otherwise specified.

With this PR, we specifically add SwiftUI and Combine as `weak` dependencies, meaning that projects embedding lottie-ios as pod are not required to embed SwiftUI and Combine.
Additionally, all files using SwiftUI are prefixed with a `canImport(SwiftUI)`.

### DIFF
--- a/Sources/Private/EmbeddedLibraries/EpoxyCore/SwiftUI/EpoxySwiftUIHostingController.swift
+++ b/Sources/Private/EmbeddedLibraries/EpoxyCore/SwiftUI/EpoxySwiftUIHostingController.swift
@@ -1,9 +1,9 @@
 // Created by eric_horacek on 10/8/21.
 // Copyright © 2021 Airbnb Inc. All rights reserved.
 
+#if canImport(SwiftUI) && !os(macOS)
 import SwiftUI
 
-#if !os(macOS)
 // MARK: - EpoxySwiftUIUIHostingController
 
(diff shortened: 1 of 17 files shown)


calda:
Thanks for the fix, this seems reasonable mostly reasonable to me. I wonder how much longer we need to support iOS 12 and earlier -- at some point it will make sense to drop support for those old versions, which will avoid the need for this.

Looks like the build is macOS build is currently failing, so that will need to be fixed.

calda:
Thanks!

Can you think of a way to test that this works as expected within having to build and run an app on an iOS 12 simulator? I'd like to set up a test case for this in CI if possible.

calda (approved):

florianrhein:
No, I'm afraid including the pod on a device or simulator actually running iOS 12 is the only way to test this...

PR #2221, opened 2023-10-20
Fix issue where Repeater would be ignored if not at top level

### LINKED ISSUE #2220: Animation misses parts on iOS only

Lottie 4.3.3

## Expected Behavior 
Lottie JSON files to animate same on iOS as it does on web.

## Actual Behavior
The animations seem to be missing some parts (an "eye" in this case).

## Animation JSON
**Listing two examples**
[chatbot-falling-asleep.json](https://github.com/airbnb/lottie-ios/files/13045542/chatbot-falling-asleep.json)
[chatbot-sorry.json](https://github.com/airbnb/lottie-ios/files/13045543/chatbot-sorry.json)

calda:
This PR fixes three issues with `Repeater` support in the Core Animation rendering engine:
 1. We would only handle `Repeater`s if they were at the top level of a `ShapeLayer`, and didn't handle `Repeater`s nested within some other `Group`
 2. `Repeater`s were incorrectly applied to all `ShapeItem`s in a group
 3. We only handled a single `Repeater` within a group

Fixes #2220.

In this example, the shape group is set up like this:
 - Blue eyes
 - Repeater
 - Pink eyes

The repeater is supposed to only be applied to the blue eyes layer.

| Before | After |
| ---- | ----- |
| Screenshot 2023-10-20 at 9 04 46 AM | Screenshot 2023-10-20 at 8 59 46 AM |


thedrick (approved):
LGTM @calda

PR #2246, opened 2023-12-04
Remove a few unused imports

hanton:
This PR removes unused Swift imports, enhancing code clarity and easing maintenance. Although their direct effect on compilation time is minimal, in extensive projects with numerous dependencies, these unused imports might slightly increase overhead. The removal ensures a cleaner, more efficient codebase and clearer, more manageable dependencies.

PR #2253, opened 2023-12-10
Remove a few unused imports

hanton:
Follow up with #2246 to fix CI failure related to Xcode 14.1 / Swift 5.6


hanton:
cc: @calda

calda (approved):
Thanks

hanton:
Thank you for highlighting this matter. I apologize for any inconvenience caused! I have a theory about what might be causing this issue, but as I was unable to replicate it in Xcode 15.1, I'm uncertain if the changes I've made will resolve it. Could you please provide further details on how to repro the issue? Specifically, is there a need to use a particular version of Xcode, or are specific Schemas required?

calda:
Interestingly I can't seem to reproduce this in Xcode, but I can reproduce it when I run this command from our `Rakefile` on the command line:

```
$ bundle exec rake build:xcframework
```

Issue #2209, opened 2023-10-12
Unable to reach the end of the animation when setting `.currentProgress` with `.coreAnimation` engine


m-yeroshenko:
Hi everyone! 

First of all, thanks for such a great library for dealing with animations.
I am sorry to say that I'm having an issue when using the latest version `4.3.3` (actually, it is also present on `4.3.0`).

So, I have a few animation JSONs and have to control their progress with the `SwiftUI.Slider` view.
When using the `.mainThread` rendering engine everything is fine, but once I switch to the `.coreAnimation` I cannot reach the animation's latest "frame". 

I would appreciate any help with this.

## Which Version of Lottie are you using?

Lottie 4.3.3

## `.mainThread` behavior

https://github.com/airbnb/lottie-ios/assets/140485731/cd1dcb73-a5bf-49ee-8c88-bfa0c407c47f

## `.coreAnimation` behavior

https://github.com/airbnb/lottie-ios/assets/140485731/f12ceaf1-f7ad-4260-8488-99d150496abf

## Animation JSON

[animation.json](https://github.com/airbnb/lottie-ios/files/12878308/animation.json)

## Code to reproduce the behavior

```swift
import Lottie
import SwiftUI

struct SwiftUIView: View {
  @State
  private var currentProgress: Double = .zero

  let jsonURL = Bundle.main.url(
    forResource: "animation",
    withExtension: "json"
  )

  var body: some View {
    if let jsonURL {
      VStack {
        LottieView {
          await LottieAnimation.loadedFrom(url: jsonURL)
        }
        .configuration(.init(renderingEngine: .mainThread))
        // .configuration(.init(renderingEngine: .coreAnimation))
        .resizable()
        .currentProgress(currentProgress)
        Slider(value: $currentProgress, in: 0 ... 1.01)
      }
    }
  }
}

#Preview {
  SwiftUIView()
}
```

polurezov8:
Hello everyone!

I'm highlighting this critical issue with the `.coreAnimation` rendering engine. 
It's affecting our use of the Lottie library significantly, especially since we cannot access the animation's last frame.

If there's any known workaround or if the team is aware of this, it would be great to hear back. This issue is impacting our application's performance.

Thanks in advance for looking into it!

PR #2254, opened 2023-12-11
Fix issue where Core Animation rendering engine couldn't display last frame of animation when paused

calda:
This PR fixes an issue where the Core Animation rendering engine couldn't display the last frame of the animation -- setting the animation progress to `1.0` would unexpectedly display the second-to-last frame. Fixes https://github.com/airbnb/lottie-ios/issues/2209.

| Before (Core Animation) | After (Core Animation) | Main Thread rendering engine |
| ----- | ---- | ----- |
| Simulator Screenshot - iPhone SE (3rd generation) - 2023-12-11 at 14 27 05 | Simulator Screenshot - iPhone SE (3rd generation) - 2023-12-11 at 14 26 58 | Simulator Screenshot - iPhone SE (3rd generation) - 2023-12-11 at 14 27 18 |

We now handle this by extending the duration of the `CAAnimation`s by one frame when the animation is paused.

-----

Some interesting discoveries I made along the way:

When playing the animation, we don't actually display the final frame of the animation. lottie-web and both iOS rendering engines behave this way. For example, when looping the animation transitions from the second-to-last frame directly to the first frame in the next loop.

When the animation is _paused_, the Main Thread rendering engine always renders the last frame of the animation. 

There is an interesting edge case here that I don't fully understand yet. Several animations, like `LottieLogo1.json`, have a totally-blank last frame. The main thread rendering engine and lottie-web handle this differently:
 - The main thread rendering engine handles this by extending the visibility of layers by one frame. This creates some bugs, like the "1" and "9/10" overlapping in the above screenshots.
 - lottie-web / lottiefiles.com appear to handle this by intelligently(?) excluding the last frame from playback. I don't understand how this works, and have asked about it [here](https://lottieanimation.slack.com/archives/C46KHBU11/p1702333515404339).

For now we'll have it so the Core Animation rendering engine matches the behavior of the Main Thread rendering engine.


Sources/Private/CoreAnimation/Animations/VisibilityAnimation.swift
```diff
@@ -28,7 +28,11 @@ extension CALayer {
     animation.keyTimes = [
       NSNumber(value: 0.0),
       NSNumber(value: max(Double(context.progressTime(for: inFrame)), 0)),
-      NSNumber(value: min(Double(context.progressTime(for: outFrame)), 1)),
+      // Anything visible during the last frame should stay visible until the absolute end of the animation.
+      //  - This matches the behavior of the main thread rendering engine.
+      context.timeRemapping(outFrame) == context.animation.endFrame
+        ? NSNumber(value: Double(1.0))
```
calda, line 34: Without this, some animations (like `LottieLogo1.json`) would unexpectedly render a blank frame as the last frame. 

This matches the behavior of the main thread rendering engine as far as I can tell, but causes bugs. e.g. this change is why the "1" and "9/10" overlap in this screenshot:



For now let's prioritize parity with the main thread rendering engine.

thedrick (approved):
LGTM @calda

PR #2260, opened 2023-12-14
Adopt policy on minimum supported Swift / Xcode version, update minimum versions to Swift 5.7 / Xcode 14.1

calda:
In https://github.com/airbnb/lottie-ios/discussions/2197 I proposed a new policy for what minimum Swift / Xcode version Lottie should support. 

> Historically Lottie has supported many past Swift versions. For example, Lottie 4.2.0 supported Swift 5.5+, which was originally released in Xcode 13.0 (Sept 2021) and replaced by Swift 5.6 in Xcode 13.3 (March 2022).
> 
> We don't really have a holistic policy on which Swift versions we should support. I've just been dropping support for older Swift versions once it became inconvenient for some specific reason. For example, we dropped support for Swift 5.5 in Lottie 4.3.0 because it made it more difficult to add support for visionOS.
>
> I'm not entirely sure what the use cases of these older versions of Swift are. I imagine the most typical use case of Lottie is in apps published on the App Store. For authors of apps published on the App Store, Apple imposes a pretty strict minimum Xcode / Swift version requirement. As of [April 2023](https://developer.apple.com/news/?id=jd9wcyov) apps must be built with Xcode 14.1 / Swift 5.7.1 or later, and starting in [April 2024](https://developer.apple.com/ios/submit/) apps must be built with Xcode 15.0 / Swift 5.9 or later.
>
> It seems reasonable for the minimum Swift version supported by Lottie to match the minimum Xcode / Swift version required by the App Store. Following that policy, Lottie would currently support Swift 5.7+ and would then support Swift 5.9+ starting after April 2024.
>
> Does anyone have a use case that would be negatively impacted by that sort of policy? Remember that if your project uses Swift 5.6 but we require Swift 5.7 in Lottie 4.4.0, you could continue using Lottie 4.3.0 in your project.

We didn't hear any negative feedback, so it seems reasonable to adopt this as our "official" policy.

This PR adds this policy to the README, and updates the minimum supported versions to Swift 5.7.1 / Xcode 14.1 (the current minimum version supported by the App Store).

### DIFF
--- a/Sources/Private/CoreAnimation/Animations/CustomPathAnimation.swift
+++ b/Sources/Private/CoreAnimation/Animations/CustomPathAnimation.swift
@@ -65,7 +65,7 @@ struct BezierPathKeyframe {
     -> KeyframeGroup<BezierPathKeyframe>
   {
     guard
-      let cornerRadius = cornerRadius,
+      let cornerRadius,
       cornerRadius.keyframes.contains(where: { $0.value.cgFloatValue > 0 })
     else {
       return path.map { path in
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,7 @@ on:
   pull_request:
     branches: [ master ]
 
+jobs:
   build-package-macos-13:
     name: "Build Package"
     runs-on: macos-13
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,23 +6,6 @@ on:
   pull_request:
     branches: [ master ]
 
-jobs:
-  build-package-macos-12:
-    name: "Build Package"
-    runs-on: macos-12
-    strategy:
-      fail-fast: false
-      matrix:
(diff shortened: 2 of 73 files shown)


Sources/Private/CoreAnimation/Animations/CustomPathAnimation.swift
```diff
@@ -65,7 +65,7 @@ struct BezierPathKeyframe {
     -> KeyframeGroup<BezierPathKeyframe>
   {
     guard
-      let cornerRadius = cornerRadius,
```
calda, line 68: wheeee

calda: the linter automatically applied Swift 5.7 features after updating `Package.swift` to `swift-tools-version: 5.7`.

thedrick (approved):
LGTM @calda

PR #2263, opened 2023-12-16
Fix build warnings

hanton:
Address the following build warnings that were introduced in #2253. Verified the issue is solved by running `$ bundle exec rake build:xcframework`. 

```
⚠️  /Users/calstephens/Documents/lottie/Sources/Public/Animation/LottieAnimation.swift:100:26: 'AnimationFrameTime' aliases 'CoreFoundation.CGFloat' and cannot be used here because 'CoreFoundation' was not imported by this file; this is an error in Swift 6
  public let startFrame: AnimationFrameTime
                         ^
⚠️  /Users/calstephens/Documents/lottie/Sources/Public/Animation/LottieAnimation.swift:103:24: 'AnimationFrameTime' aliases 'CoreFoundation.CGFloat' and cannot be used here because 'CoreFoundation' was not imported by this file; this is an error in Swift 6
  public let endFrame: AnimationFrameTime
                       ^
⚠️  /Users/calstephens/Documents/lottie/Sources/Public/DynamicProperties/AnyValueProvider.swift:25:25: 'AnimationFrameTime' aliases 'CoreFoundation.CGFloat' and cannot be used here because 'CoreFoundation' was not imported by this file; this is an error in Swift 6
  func hasUpdate(frame: AnimationFrameTime) -> Bool
                        ^
⚠️  /Users/calstephens/Documents/lottie/Sources/Public/DynamicProperties/AnyValueProvider.swift:31:28: 'AnimationFrameTime' aliases 'CoreFoundation.CGFloat' and cannot be used here because 'CoreFoundation' was not imported by this file; this is an error in Swift 6
  public func value(frame: AnimationFrameTime) -> Any {
                           ^
⚠️  /Users/calstephens/Documents/lottie/Sources/Public/DynamicProperties/AnyValueProvider.swift:82:17: 'AnimationFrameTime' aliases 'CoreFoundation.CGFloat' and cannot be used here because 'CoreFoundation' was not imported by this file; this is an error in Swift 6
  case closure((AnimationFrameTime) -> T)
                ^
⚠️  /Users/calstephens/Documents/lottie/Sources/Public/DynamicProperties/AnyValueProvider.swift:110:49: 'AnimationFrameTime' aliases 'CoreFoundation.CGFloat' and cannot be used here because 'CoreFoundation' was not imported by this file; this is an error in Swift 6
  case keyframes([Keyframe<Any>], interpolate: (AnimationFrameTime) -> Any)
                                                ^
[...]

PR #2264, opened 2023-12-16
Continue XCFramework job if certificate install step fails

calda:
This PR updates the job that builds `Lottie.xcframework` to continue if the certificate install step fails. Should fix CI for https://github.com/airbnb/lottie-ios/pull/2263.


calda:
Confirmed this fixed the issue:

image

calda:
merging to unblock #2263

PR #2252, opened 2023-12-08
Add privacy manifest

calda:
This PR adds a privacy manfiest (`PrivacyInfo.xcprivacy`), which will be required for privacy-impacting SDKs ([including Lottie](https://developer.apple.com/support/third-party-SDK-requirements/)) starting in spring 2024. Fixes #2213.

We can include the manifest as a resource in our dependency definitions and Xcode project schemes. I verified that `Lottie.framework` and `Lottie.xcframework` build products now include the privacy manifest file.

We will additionally need to code sign `Lottie.xcframework`, but that will come later in a follow-up.

PR #2268, opened 2023-12-21
Fix privacy manifest location for SPM

mjburghard:
#2252 introduced `PrivacyInfo.xcprivacy` in the repositories root directory.
While testing the integration of Lottie with SPM, I noticed that the manifest is not included in the app's bundle. 
After comparing it with other packages that work, I was able to narrow it down to the location in the repository, or more preciously the `..` to point to the parent directory of `Sources` in the copy rule.

Moving `PrivacyInfo.xcprivacy` into `Sources` solves the issue.


calda (approved):
Thank you! Very helpful. Could you also update the link in README.md?

PR #2271, opened 2024-01-03
Do not create `DotLottieImageProvider` instance if there's no image files

hardworker:
`LottieFiles dotLottie Exporter 0.1` may produce `.lottie` files, where an animation contains `WebP` images not in `images` folder, but in `assets` part of `.json`.

This may lead to an issue while using `LottieAnimationLayer`'s:
```swift
public init(
    dotLottie: DotLottieFile?,
    animationId: String? = nil,
    textProvider: AnimationKeypathTextProvider = DefaultTextProvider(),
    fontProvider: AnimationFontProvider = DefaultFontProvider(),
    configuration: LottieConfiguration = .shared,
    logger: LottieLogger = .shared)
```

It takes `dotLottie?.imageProvider` which is not `nil` and cannot decode base64-encoded images inside `.json` files. However, if it's `nil` it fallbacks to `BundleImageProvider` and everything works as expected. So the main idea of my changes is to avoid `DotLottieImageProvider` creation if there're no images to provide.


calda:
Can you share an animation file that depends on this behavior to render correctly? That would be a great addition to the snapshot testing suite

calda (approved):

hardworker:
@calda Done!

calda:
Thanks!

PR #2275, opened 2024-01-04
[lottie/objectiveC] Add support for visionOS

joannaquu:
Added support for visionOS. 

Currently all of the iOS code builds for visionOS except `_compContainer.rasterizationScale = self.window.screen.scale;`, which uses `UIScreen` (unavailable on visionOS). If-def it and use a default value of 2.

### DIFF
--- a/lottie-ios/Classes/Private/LOTAnimationView.m
+++ b/lottie-ios/Classes/Private/LOTAnimationView.m
@@ -691,7 +691,11 @@ static NSString * const kCompContainerAnimationKey = @"play";
 }
 
 - (void)didMoveToWindow {
+#if TARGET_OS_IOS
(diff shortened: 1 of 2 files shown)


lottie-ios/Classes/Private/LOTAnimationView.m
```diff
@@ -691,7 +691,11 @@ - (void)willMoveToWindow:(UIWindow *)newWindow {
 }
 
 - (void)didMoveToWindow {
+#if TARGET_OS_IOS
```
calda, line 694: We need to make sure this also handles macOS and tvOS right? Can you just use `TARGET_OS_VISION` and `!TARGET_OS_VISION`?

calda (approved):
Thanks!

PR #2277, opened 2024-01-05
Fix parsing issue, add support for DotLottieConfiguration in SwiftUI LottieView

### LINKED ISSUE #2228: Passing DotLottie configuration to LottieView

Lottie 4.3.0

## Expected Behavior 
Loading a dotLottie file into a LottieView would use the loopMode and playback speed defined in the dotLottie.

## Actual Behavior
There is no way to configure the LottieView with the dotLottie without reimplementing loading the dotLottie

## Example

```swift
LottieView {
	try await DotLottieFile.asset(named: name, bundle: bundle)
}
```

Once the dotlottie file has actually been loaded asynchronously on line 2, its not possible to pass the `DotLottieConfiguration` accessible in `.configuration` to the functions that modify the LottieView.

I'm happy to try making a PR but I'm not sure what the best solution would be, I was thinking either a separate initializer that will load and use the configuration, or having a `.playbackMode(.configured)` or something. The problem with the latter is it could be applied to a LottieView with a loaded JSON file which has no configuration.

### LINKED ISSUE #2269: Error when decoding animation "trails_3_s": invalidInput
Hello, I have used the latest version of SDK (4.3.4), but there is an error when playing the animation. I have provided the animation file for you to test and see if it can help me solve this playback problem. Thank you
[trails_3_s.zip](https://github.com/airbnb/lottie-ios/files/13764588/trails_3_s.zip)

calda:
This PR fixes two issues:
 1. Fixes #2269
 2. Fixes #2228

For #2269, we were failing to parse the animation. It's missing some fields that iOS requires, but that web doesn't require. We can relax the parsing a bit to allow this animation to parse successfully.

When testing the animation after fixing the parsing issue, I ran in to #2228. I bundled that animation and its image assets in a dotLottie file, but it was completely blank because the `DotLottieConfiguration.imageProvider` was being ignored.

| Before #2269 | Before fixing #2228 | After |
| ----- | ----- | ----- |
|  |  |  |


Sources/Private/CoreAnimation/Layers/LayerModel+makeAnimationLayer.swift
```diff
@@ -51,6 +51,9 @@ extension LayerModel {
     case (.null, _):
       return TransformLayer(layerModel: self)
 
+    case (.unknown, _):
+      return nil
```
calda, line 55: This fixes an issue where a layer with an unknown type would log a compatibility error and fall back to the main thread rendering engine, even though the main thread engine wouldn't be able to handle it either

Sources/Private/Model/Assets/AssetLibrary.swift
```diff
@@ -18,14 +18,17 @@ final class AssetLibrary: Codable, AnyInitializable, Sendable {
...
+    {
+      if
+        keyContainer.contains(.layers),
+        let precompAsset = try? container.decode(PrecompAsset.self)
+      {
         decodedAssets[precompAsset.id] = precompAsset
         precompAssets[precompAsset.id] = precompAsset
-      } else {
-        let imageAsset = try container.decode(ImageAsset.self)
+      } else if let imageAsset = try? container.decode(ImageAsset.self) {
```
calda, line 31: If an image asset fails to parse for some reason, we can recover gracefully instead of throwing an error and rejecting the animation

Sources/Private/Model/DictionaryInitializable.swift
```diff
@@ -8,7 +8,7 @@
 // MARK: - InitializableError
 
 enum InitializableError: Error {
-  case invalidInput
+  case invalidInput(file: StaticString = #file, line: UInt = #line)
```
calda, line 11: Adding in file and line context makes it easier to debug where these errors are coming from

Sources/Private/Model/Layers/LayerModel.swift
```diff
@@ -93,7 +93,7 @@ class LayerModel: Codable, DictionaryInitializable {
     inFrame = try container.decode(Double.self, forKey: .inFrame)
     outFrame = try container.decode(Double.self, forKey: .outFrame)
     startTime = try container.decode(Double.self, forKey: .startTime)
-    transform = try container.decode(Transform.self, forKey: .transform)
+    transform = try container.decodeIfPresent(Transform.self, forKey: .transform) ?? .default
```
calda, line 96: All of the individual properties in a `Transform` are optional and have default values, so we can easily make the `Transform` itself optional with a default

Sources/Public/Animation/LottieView.swift
```diff
@@ -438,6 +436,20 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     }
   }
 
+  /// Returns a copy of this view with the `DotLottieConfigurationComponents`
+  /// updated to the given value.
+  ///  - Defaults to `[.imageProvider]`
+  ///  - If a component is specified here, that value in the `DotLottieConfiguration`
+  ///    of an active dotLottie animation will override any value provided via other methods.
+  public func dotLottieConfigurationComponents(
```
calda, line 444: Since the `DotLottieConfiguration` _overrides_ any value you pass in via `.imageProvider(...)`, `.loopMode(...)`, or `.speed(...)`, it seems like a good idea to make this behavior configurable. Otherwise it would be impossible to use a custom loop mode or speed when playing some dotLottie animations.

Sources/Private/CoreAnimation/Layers/LayerModel+makeAnimationLayer.swift
miguel-jimenez-0529: Should we add a this comment to the code?

Sources/Private/Model/Assets/AssetLibrary.swift
```diff
@@ -18,14 +18,17 @@ final class AssetLibrary: Codable, AnyInitializable, Sendable {
     var imageAssets = [String : ImageAsset]()
     var precompAssets = [String : PrecompAsset]()
 
-    while !container.isAtEnd {
-      let keyContainer = try containerForKeys.nestedContainer(keyedBy: PrecompAsset.CodingKeys.self)
-      if keyContainer.contains(.layers) {
-        let precompAsset = try container.decode(PrecompAsset.self)
+    while
+      !container.isAtEnd,
+      let keyContainer = try? containerForKeys.nestedContainer(keyedBy: PrecompAsset.CodingKeys.self)
```
miguel-jimenez-0529, line 23: Should we add resilient decoder https://github.com/airbnb/ResilientDecoding in case we need the error at another time?

Sources/Private/Utility/LottieAnimationSource.swift
```diff
@@ -20,8 +20,18 @@ extension LottieAnimationSource {
...
+  }
+
+  /// The `DotLottieFile.Animation`, if this is a dotLottie animation
+  var dotLottieAnimation: DotLottieFile.Animation? {
+    switch self {
+    case .lottieAnimation:
+      return nil
     case .dotLottieFile(let dotLottieFile):
-      return dotLottieFile.animation()?.animation
+      return dotLottieFile.animation()
```
miguel-jimenez-0529, line 34: 👍

miguel-jimenez-0529 (approved):
LGTM

Sources/Private/Model/Assets/AssetLibrary.swift
calda: We don't use the `Codable` decoding implementation by default (the dictionary-based implementation is the default in Lottie 4.0+ since it's a lot faster), so no need to invest much in the `Codable` implementation

PR #2284, opened 2024-01-08
Include dSYMs in xcframework build

Cyberbeni:
resolves #2282

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -68,32 +68,32 @@ namespace :build do
     sh 'rm -rf .build/archives'
 
     # Build the framework for each supported platform, including simulators
-    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (iOS)" -destination generic/platform=iOS -archivePath ".build/archives/Lottie_iOS" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES ENABLE_BITCODE=NO')
-    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (iOS)" -destination "generic/platform=iOS Simulator" -archivePath ".build/archives/Lottie_iOS_Simulator" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES ENABLE_BITCODE=NO')
-    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (iOS)" -destination "generic/platform=macOS,variant=Mac Catalyst" -archivePath ".build/archives/Lottie_Mac_Catalyst" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES ENABLE_BITCODE=NO')
-    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (macOS)" -destination generic/platform=macOS -archivePath ".build/archives/Lottie_macOS" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES ENABLE_BITCODE=NO')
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -68,32 +68,32 @@ namespace :build do
...
-    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (macOS)" -destination generic/platform=macOS -archivePath ".build/archives/Lottie_macOS" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES ENABLE_BITCODE=NO')
-    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (tvOS)" -destination generic/platform=tvOS -archivePath ".build/archives/Lottie_tvOS" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES ENABLE_BITCODE=NO')
-    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (tvOS)" -destination "generic/platform=tvOS Simulator" -archivePath ".build/archives/Lottie_tvOS_Simulator" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES ENABLE_BITCODE=NO')
+    xcodebuild('archive -workspace Lottie.xcworkspace -scheme "Lottie (iOS)" -destination generic/platform=iOS -archivePath ".build/archives/Lottie_iOS" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES ENABLE_BITCODE=NO SWIFT_SERIALIZE_DEBUGGING_OPTIONS=NO')
```
calda, line 71: What does `SWIFT_SERIALIZE_DEBUGGING_OPTIONS=NO` do exactly?

Rakefile
```diff
@@ -68,32 +68,32 @@ namespace :build do
...
     # Combine all of the platforms into a single XCFramework
     xcframeworkInvocation = [
       '-create-xcframework',
-      '-framework .build/archives/Lottie_iOS.xcarchive/Products/Library/Frameworks/Lottie.framework',
-      '-framework .build/archives/Lottie_iOS_Simulator.xcarchive/Products/Library/Frameworks/Lottie.framework',
-      '-framework .build/archives/Lottie_Mac_Catalyst.xcarchive/Products/Library/Frameworks/Lottie.framework',
-      '-framework .build/archives/Lottie_tvOS.xcarchive/Products/Library/Frameworks/Lottie.framework',
-      '-framework .build/archives/Lottie_tvOS_Simulator.xcarchive/Products/Library/Frameworks/Lottie.framework',
-      '-framework .build/archives/Lottie_macOS.xcarchive/Products/Library/Frameworks/Lottie.framework',
+      '-archive .build/archives/Lottie_iOS.xcarchive -framework Lottie.framework',
```
calda, line 86: Based on your comment [here](https://github.com/airbnb/lottie-ios/discussions/2282#discussioncomment-8054291), using `-archive` and `-framework` separately makes it so the dSYMs from the archive are included?

Rakefile
Cyberbeni: It avoids breaking lldb when used on a machine that doesn't have the derived data files created during the build: https://github.com/facebook/facebook-ios-sdk/issues/1628#issuecomment-771462320

Rakefile
Cyberbeni: Yes, this will include anything that belongs to the specific framework inside the archive.

calda (approved):

Rakefile
Cyberbeni: This is mentioned in a WWDC video as the official source, I didn't see it in the docs: https://developer.apple.com/videos/play/wwdc2022/110370?time=1096

Rakefile
Cyberbeni: This is also the "recommended" way according to the documentation: https://developer.apple.com/documentation/xcode/creating-a-multi-platform-binary-framework-bundle#Generate-the-XCFramework-bundle

calda:
Thank you @Cyberbeni!

PR #2285, opened 2024-01-08
Fix issue where manually interpolated keyframes could animate incorrectly

### LINKED ISSUE #2226: Animation is not correct after updating to 4.3.x
After updating to 4.3.0 or higher, we noticed that our startup animation handles flip transitions incorrectly. Running `git bisect` against the example project leads us to believe [this PR](https://github.com/airbnb/lottie-ios/pull/2067) introduced the issue.

We are able to work around this issue by using main thread as the rendering engine.

Lottie 4.3.x

## Expected Behavior 
In 4.2.0 we see the animation rotate the B character up, then flip then rotate down.

| | |
|-|-|
|4 2 0 A  | 4 2 0 B |

## Actual Behavior
In 4.3.x the x position of the B character changes during the flip transition

| | |
|-|-|
| 4 3 3 A| 4 3 3 B|

## Animation JSON
 [0splash_logo_animation.json](https://github.com/airbnb/lottie-ios/files/13242566/0splash_logo_animation.json)

calda:
In some cases we manually interpolate keyframes rather than letting Core Animation perform the interpolation. To do this we create a CAAnimation keyframe for each frame in the animation. This lets us work around issues where Core Animation interpolation behaves in unexpected ways, or doesn't support certain functionality.

Previously, we allowed Core Animation to additionally interpolate between these manual frames. For example, on frame 19.5 Core Animation would itself interpolate half way between our manually generated keyframes for frame 19 and frame 20. This is undesirable, since we are intentionally not using Core Animation interpolation in these cases.

Instead, we can set `isHold: true` on the generated keyframes (to indicate that they shouldn't animate) and instead just increase the number of times that we generate a keyframe for to make sure these static keyframes still appear to animate smoothly.

Fixes #2226:

| Before | After |
| ----- | ---- |
|  |  |


Example/Example/ExampleApp.swift
```diff
@@ -5,11 +5,24 @@ import SwiftUI
 
 @main
 struct ExampleApp: App {
+
+  // MARK: Lifecycle
+
+  init() {
+    // Register fonts from the Samples/Fonts directory
+    for fontAssetURL in Bundle.main.urls(forResourcesWithExtension: "ttf", subdirectory: "Samples/Fonts") ?? [] {
```
calda, line 13: Unrelated to this bugfix -- I added a test case for https://github.com/airbnb/lottie-ios/issues/1939, since we didn't have any example animations that used custom fonts.

miguel-jimenez-0529 (approved):
LGTM

PR #2286, opened 2024-01-09
Add time remapping support to Core Animation rendering engine

### LINKED ISSUE #1648: The Core Animation rendering engine partially supports time remapping keyframes
version 3.4.0 master branch

Lottie/AnimationView.swift:1003: Assertion failed: Encountered Core Animation compatibility issues while setting up animation:
[Puke Splash_PRECOMP_02] The Core Animation rendering engine partially supports time remapping keyframes, but this is somewhat experimental and has some known issues. Since it doesn't work in all cases, we have to fall back to using the main thread engine when using `RenderingEngineOption.automatic`.
[Puke Splash_PRECOMP_02] The Core Animation rendering engine partially supports time remapping keyframes, but this is somewhat experimental and has some known issues. Since it doesn't work in all cases, we have to fall back to using the main thread engine when using `RenderingEngineOption.automatic`.

This animation cannot be rendered correctly by the Core Animation engine.
To resolve this issue, you can use `RenderingEngineOption.automatic`, which automatically falls back
to the Main Thread rendering engine when necessary, or just use `RenderingEngineOption.mainThread`.

calda:
This PR adds support for time remapping to the Core Animation rendering engine. Fixes #1648.

Time remapping is a function that changes how time progresses for a layer. It's an arbitrarily complex function from the current "global time" of the overall animation to the "local time" in the layer that should be displayed.

For example, if you have a layer that animates from 0 to 10, you could have that animation play over range 50 to 100 in the overall animation. That would be a function like `local_time = (global_time - 50) / 5`. Time remapping functions can be arbitrarily complex and non-linear, and can even loop / go backwards.

All of the keyframes a Lottie animation are expressed in "local time" for the layer, and we cannot convert these local times to global times. To apply this "global time to local time" time mapping for the layer, we have to manually interpolate the keyframes at each individual global frame, and then apply that time remapping at each frame.

This PR adds a new `Keyframes.manuallyInterpolatedWithTimeRemapping` codepath that is used for any layer that is a child of some `PreCompLayer` with a `timeRemapping` function. 

| Without time remapping | With time remapping |
| --- | --- |
|  |  |

| Without time remapping | With time remapping |
| --- | --- |
|  |  |


Sources/Private/CoreAnimation/Layers/PreCompLayer.swift
```diff
@@ -104,36 +103,3 @@ extension PreCompLayer: CustomLayoutLayer {
...
-extension KeyframeInterpolator where ValueType == AnimationFrameTime {
-  /// A `KeyframeInterpolator` for the given `timeRemapping` keyframes
-  static func timeRemapping(
-    keyframes timeRemappingKeyframes: KeyframeGroup<LottieVector1D>,
-    context: LayerContext)
-    throws -> KeyframeInterpolator<AnimationFrameTime>
-  {
-    try context.logCompatibilityIssue("""
-      The Core Animation rendering engine partially supports time remapping keyframes,
-      but this is somewhat experimental and has some known issues. Since it doesn't work
```
calda, line 117: This previous implementation attempted to invert the time remapping keyframes. This happened to work in some simple cases, but didn't work in general.

Tests/__Snapshots__/AutomaticEngineTests/testAutomaticEngineDetection.Boat_Loader.txt
```diff
@@ -1,2 +1 @@
-Does not support Core Animation engine. Encountered compatibility issues:
-[Chest] The Core Animation rendering engine partially supports time remapping keyframes, but this is somewhat experimental and has some known issues. Since it doesn't work in all cases, we have to fall back to using the main thread engine when using `RenderingEngineOption.automatic`.
\ No newline at end of file
+Supports Core Animation engine
```
calda, line 1: woohoo! 🥳 🎉

Sources/Private/CoreAnimation/Animations/VisibilityAnimation.swift
```diff
@@ -10,32 +10,71 @@ extension CALayer {
...
+      ])
+
+      try addAnimation(
+        for: .isHidden,
+        keyframes: isHiddenKeyframes.map { Hold(value: $0) },
+        value: { $0.value },
+        context: context)
+    }
+
+    /// Otherwise continue using the previous codepath that doesn't support complex time remapping.
```
miguel-jimenez-0529, line 31: I think the wording `previous` here makes sense in a PR but could be confusing after we merge the code. Someone may not know the `else` block was the previous codepath.

Sources/Private/CoreAnimation/Animations/VisibilityAnimation.swift
```diff
@@ -10,32 +10,71 @@ extension CALayer {
...
+    }
+  }
+}
+
+// MARK: - Hold
+
+/// An `Interpolatable` container that animates using "hold" keyframes.
+/// The keyframes do not animate, and instead always display the value from the most recent keyframe.
+/// This is necessary when passing non-interpolatable values to a method that requires an `Interpolatable` conformance.
+struct Hold<T>: Interpolatable {
```
miguel-jimenez-0529, line 70: Should we move this to its own file?

miguel-jimenez-0529 (approved):
LGTM

Sources/Private/CoreAnimation/Animations/VisibilityAnimation.swift
calda: good idea

Sources/Private/CoreAnimation/Animations/VisibilityAnimation.swift
calda: I'll move it to the file with the `Interpolatable` protocol and main interpolation implementations

PR #2288, opened 2024-01-12
Use resource_bundles instead of resource in podspec for privacy manifest

dcacenabes:
In #2252 the privacy manifest was added. For cocoapods it is being added using `resource`, which can end up with duplicate files if other library linked on the application provides a file with the same name and you are linking them statically.

By using resource_bundles, we can ensure that there will not be overlaps.

More details in this comment in a Google repository on this same topic:
https://github.com/google/promises/pull/226#discussion_r1447871477


calda (approved):
Thanks!

maurovc (approved):

Issue #2267, opened 2023-12-21
Size is not propagated from LottieView to placeholder in SwiftUI


denandreychuk:
Body of LottieView is wrapper into ZStack, which cause placeholder not to inherit frame size from parent. 

Example:
```
LottieView {
    try await DotLottieFile.named("animation")
} placeholder: {
    Image("placeholder")
        .resizable()
}
.resizable()
.frame(height: 100)
```

annie-gupta:
Hi @denandreychuk , Please assign this issue to me. Thank you.

denandreychuk:
@annie-gupta unfortunately only someone from team members can do the assignments, but I think that you can make a fork and push the PR to fix this

Issue #2257, opened 2023-12-13
Async loading of animation removes SwiftUI animation


denandreychuk:
Suppose we have a simple animation view which loads DotLottie file:
```
@State private var flag = false

var body: some View {
    LottieView {
        try await DotLottieFile.named(animation)
    } placeholder: {
        Image(placeholder)
    }
    .looping()
    .resizable()
    .offset(y: flag ? 0 : 300)
    .onAppear {
        withAnimation {
            flag = true
        }
    }
}
```

The code above moves to offset 0 immediately, removing SwiftUI offset animation

SoylentGraham:
Can I ask where you found the documentation for this style of loading?
I've spent the last hour trying to load data in 10 different ways, and then stumbled upon this example, which worked! :)

PR #2289, opened 2024-01-20
Fix LottieView bugs

miguel-jimenez-0529:
This PR fixes https://github.com/airbnb/lottie-ios/issues/2257 
This issue was caused by the animation view inside the `if` statement. By the time it was loaded and inserted in the view hierarchy it didn't include the animation transaction therefore it was not animated.

This PR also fixes https://github.com/airbnb/lottie-ios/issues/2267
This was caused by not having no relation between the animation view  and it's placeholder. Now we use a overlay for the placeholder which provides the same size from the animation view to the placeholder

Finally, I added a new Demo screen where we can add simple tests for compatibility with other SwiftUI APIs

https://github.com/airbnb/lottie-ios/assets/20710815/1af6e692-eb68-4575-ba1d-6ee590b233fe

### DIFF
--- a/Sources/Public/Animation/LottieView.swift
+++ b/Sources/Public/Animation/LottieView.swift
@@ -129,6 +129,7 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
       applyCurrentAnimationConfiguration(to: context.view)
     }
     .configurations(configurations)
+    .opacity(animationSource == nil ? 0 : 1)
     .overlay {
       placeholder?()
         .opacity(animationSource == nil ? 1 : 0)
@@ -565,12 +566,13 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
 
 @available(iOS 13.0, tvOS 13.0, macOS 10.15, *)
 extension View {
-  
+
   /// The `.overlay` modifier that uses a `ViewBuilder` is available in iOS 15+, this helper function helps us to use the same API in older OSs
-  fileprivate func overlay<V>(
-    @ViewBuilder content: () -> V
-  ) -> some View where V : View {
-    self.overlay(content(), alignment: .center)
+  fileprivate func overlay(
+    @ViewBuilder content: () -> some View)
+    -> some View
+  {
+    overlay(content(), alignment: .center)
   }
 }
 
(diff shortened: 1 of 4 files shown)


calda (approved):
Somebody called the exterminator! 🐞 🐛

Sources/Public/Animation/LottieView.swift
```diff
@@ -115,26 +115,23 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+        logger: logger)
+    }
+    .sizing(sizing)
+    .configure { context in
+      applyCurrentAnimationConfiguration(to: context.view)
+    }
+    .configurations(configurations)
+    .overlay {
+      placeholder?()
+        .opacity(animationSource == nil ? 1 : 0)
```
calda, line 134: Should we also set `.opacity(animationSource == nil ? 0 : 1)` on the `LottieAnimationView`? That may best match the existing behavior of an `if` where we only every displayed one or the other

calda: Yes, I think this will be necessary -- if you change `.reloadAnimationTrigger(currentURLIndex, showPlaceholder: false)` to `showPlaceholder: true` in `AnimationPreviewView` and test the "Multiple Animations" demo, you'll see the behavior difference:

| Before | With this current change |
| ---- | ---- |
| Screenshot 2024-01-19 at 4 29 11 PM | Screenshot 2024-01-19 at 4 30 03 PM |

Sources/Public/Animation/LottieView.swift
```diff
@@ -115,26 +115,23 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+        fontProvider: fontProvider,
+        configuration: configuration,
+        logger: logger)
+    }
+    .sizing(sizing)
+    .configure { context in
+      applyCurrentAnimationConfiguration(to: context.view)
+    }
+    .configurations(configurations)
+    .overlay {
```
calda, line 132: This is a behavior change that will probably affect existing consumers (see comparison below) but I think this is ok because the updated behavior is seems more expected and predictable

```swift
LottieView {
  try await loadAnimation()
} placeholder: {
  LoadingIndicator()
    .frame(width: 50, height: 50)
}
```

| Before | After |
| ---- | ---- |
| Screenshot 2024-01-19 at 4 29 11 PM | Screenshot 2024-01-19 at 4 31 28 PM |

PR #2290, opened 2024-01-21
CSCache instead of LRUCache for LottieView animation would restart from beginning after backgrounding app

DevVenusK:
This PR is problem where the SwiftUI LottieView would consistently restart playback from the beginning after the app went into the background. It introduces the use of NSCache in this context.

In Swift, NSCache includes the NSDiscardableContent protocol, providing a way to control the timing of object removal from the cache. Using NSCache seems preferable to relying on external libraries. Even when the app is in the background, NSCache won't automatically remove objects that are still in use.

[https://developer.apple.com/documentation/foundation/nscache#//apple_ref/occ/cl/NSCache](NSCache)

A common data type stored in NSCache objects is an object that implements the NSDiscardableContent protocol. Storing this type of object in a cache has benefits, because its content can be discarded when it is not needed anymore, thus saving memory. By default, NSDiscardableContent objects in the cache are automatically removed from the cache if their content is discarded, although this automatic removal policy can be changed. If an NSDiscardableContent object is put into the cache, the cache calls discardContentIfPossible on it upon its removal.


DevVenusK (commented):
LintError
Trailing Newline Violation: Files should have a single trailing newline (trailing_newline)

calda:
Thank you for the suggestion. I prefer LRUCache because it is more controllable and predictable than NSCache. I would prefer to continue using LRUCache.

DevVenusK:
> Thank you for the suggestion. I prefer LRUCache because it is more controllable and predictable than NSCache. I would prefer to continue using LRUCache.

@calda Yes, I respect your opinion, but it seems like there is some code in the current LRUCache logic that may cause some bugs. I'll request a separate PR for that. thank you !

calda:
If there are bugs or issues in LRUCache please open a pull request on https://github.com/nicklockwood/LRUCache itself, then we can pull in updates once they're merged to master there.

PR #2293, opened 2024-01-23
fix: add visionOS support in the Lottie podspec

TheRogue76:
Was updating to 4.4.0 and i noticed that even though the Package.swift declares vision os support, the podspec doesn't. Not sure if i am missing something or not, but i figured a PR wouldn't hurt


NachoSoto:
FYI this is still broken on CocoaPods 1.15.0:
> [!] Unrecognized platform `visionOS`. Valid platforms: iOS, macOS, watchOS, visionOS, tvOS

PR #2296, opened 2024-01-24
Simple: Change accesscontrol to methods that are not used in other files to private.

DevVenusK:
Currently, the updateRasterizationState method is not being used in other files, but the accesscontrol is set to internal. So I changed it to private.

If you need to use it in another file in the future, it would be a good idea to change it to internal.

### DIFF
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -1007,7 +1007,6 @@ public class LottieAnimationLayer: CALayer {
       }
     }
   }
-  
   public func updateAnimationForForegroundState(wasWaitingToPlayAnimation: Bool) {
     if let currentContext = animationContext {
       if wasWaitingToPlayAnimation {
@@ -1467,7 +1466,6 @@ public class LottieAnimationLayer: CALayer {
       return animation?.reducedMotionMarker
     }
   }
-  
   private func loadAnimation(_ dotLottieAnimation: DotLottieFile.Animation) {
     loopMode = dotLottieAnimation.configuration.loopMode
     animationSpeed = CGFloat(dotLottieAnimation.configuration.speed)
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -1089,7 +1089,7 @@ public class LottieAnimationLayer: CALayer {
 
     addNewAnimationForContext(newContext)
   }
-  
+
   func updateRasterizationState() {
     if isAnimationPlaying {
       animationLayer?.shouldRasterize = false
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -1041,14 +1041,6 @@ public class LottieAnimationLayer: CALayer {
     }
   }
 
-  func updateRasterizationState() {
-    if isAnimationPlaying {
-      animationLayer?.shouldRasterize = false
-    } else {
-      animationLayer?.shouldRasterize = shouldRasterizeWhenIdle
-    }
-  }
-
   /// Updates the animation frame. Does not affect any current animations
   func updateAnimationFrame(_ newFrame: CGFloat) {
     // In performance tests, we have to wrap the animation layer setup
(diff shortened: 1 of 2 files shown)


calda:
The implementation of `LottieAnimationView.updateRasterizationState` is:

```swift
  func updateRasterizationState() {
    if lottieAnimationLayer.isAnimationPlaying {
      lottieAnimationLayer.animationLayer?.shouldRasterize = false
    } else {
      lottieAnimationLayer.animationLayer?.shouldRasterize = lottieAnimationLayer.shouldRasterizeWhenIdle
    }
  }
```

which is identical to the implementation of `LottieAnimationLayer.updateRasterizationState`.

It seems like a better change would be to update `LottieAnimationView.updateRasterizationState` to just call through to the underlying layer like this:

```swift
  func updateRasterizationState() {
    lottieAnimationLayer.updateRasterizationState()
  }
```

DevVenusK:
> The implementation of `LottieAnimationView.updateRasterizationState` is:
> 
> ```swift
>   func updateRasterizationState() {
>     if lottieAnimationLayer.isAnimationPlaying {
>       lottieAnimationLayer.animationLayer?.shouldRasterize = false
>     } else {
>       lottieAnimationLayer.animationLayer?.shouldRasterize = lottieAnimationLayer.shouldRasterizeWhenIdle
>     }
>   }
> ```
> 
> which is identical to the implementation of `LottieAnimationLayer.updateRasterizationState`.
> 
> It seems like a better change would be to update `LottieAnimationView.updateRasterizationState` to just call through to the underlying layer like this:
> 
> ```swift
>   func updateRasterizationState() {
>     lottieAnimationLayer.updateRasterizationState()
>   }
> ```

@calda That sounds great! I fixed it and requested it again!

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1008,6 +1008,14 @@ public class LottieAnimationLayer: CALayer {
     }
   }
 
+  public func updateRasterizationState() {
```
calda, line 1011: Let's keep this internal

```suggestion
  func updateRasterizationState() {
```

calda (approved):

Sources/Public/Animation/LottieAnimationLayer.swift
DevVenusK: Oh, no!

Sorry, I misunderstood that LottieAnimationView was a different module. I fixed it and requested it again!

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1007,7 +1007,7 @@ public class LottieAnimationLayer: CALayer {
       }
     }
   }
-
+
```
calda, line 1010:
```suggestion

```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1467,7 +1467,7 @@ public class LottieAnimationLayer: CALayer {
       return animation?.reducedMotionMarker
     }
   }
-
+
```
calda, line 1470:
```suggestion

```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1467,7 +1467,6 @@ public class LottieAnimationLayer: CALayer {
       return animation?.reducedMotionMarker
     }
   }
-
   private func loadAnimation(_ dotLottieAnimation: DotLottieFile.Animation) {
```
calda, line 1470:
```suggestion

  private func loadAnimation(_ dotLottieAnimation: DotLottieFile.Animation) {
```

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1097,6 +1089,14 @@ public class LottieAnimationLayer: CALayer {
...
     addNewAnimationForContext(newContext)
   }
+  
+  func updateRasterizationState() {
+    if isAnimationPlaying {
+      animationLayer?.shouldRasterize = false
+    } else {
+      animationLayer?.shouldRasterize = shouldRasterizeWhenIdle
+    }
+  }
```
calda, line 1099:
```suggestion

  func updateRasterizationState() {
    if isAnimationPlaying {
      animationLayer?.shouldRasterize = false
    } else {
      animationLayer?.shouldRasterize = shouldRasterizeWhenIdle
    }
  }
```

calda: this is what I get for making code changes via the GitHub PR UI

Issue #2297, opened 2024-01-25
no rule to process file 'PrivacyInfo.xcprivacy' of type 'text.xml' with CocoaPods


rradenkov-paysafe:
When integrating Lottie 4.4.0 with CocoaPods a build warning is displayed that PrivacyInfo.xcprivacy can not be processed.
Reason for this is that 'PrivacyInfo.xcprivacy' is included in source files and Xcode treats it as such.
```
s.source_files = 'Sources/**/*'
```

## Which Version of Lottie are you using?

Lottie 4.4.0

## Expected Behavior 
No Xcode warnings are shown

## Actual Behavior
Following Xcode build warning is displayed
`no rule to process file '.../lottie-ios/Sources/PrivacyInfo.xcprivacy' of type 'text.xml' for architecture 'x86_64'`

## Animation JSON
N/A

VaslD:
Are you using Xcode 14? Privacy manifests (`*.xcprivacy`) are introduced in Xcode 15 so I don't know how Xcode 14 handles them.

I just upgraded to Lottie 4.4, with the latest Xcode (15.2) I did not encounter this error.

rradenkov-paysafe:
I am using Xcode 15.2 as well.
![lottie-cocoapods-warning](https://github.com/airbnb/lottie-ios/assets/72293299/f206d117-b470-4977-9878-c2bee56ef83a)

Other libraries that added Privacy manifests don't include it in sources and I don't see build warnings: https://github.com/onevcat/Kingfisher/blob/ac3268e519bd81c333ce7a8d69d56ab66081fa60/Kingfisher.podspec#L38

picselladmin:
Facing the same issue here

calda:
This was fixed by https://github.com/airbnb/lottie-ios/pull/2307 right?

OgabekDev:
I am using LottieView via Cocoapods. Could you tell me what the solution is? I am updating the pods but I haven't gotten a result.

kronenthaler:
I see the fixed merged into the master branch, but can you release a new version with this change?

calda:
Just released 4.4.1 https://github.com/airbnb/lottie-ios/releases/tag/4.4.1

OgabekDev:
Thank you. I updated the pod

OgabekDev:
[!] Error installing lottie-ios
[...]

PR #2307, opened 2024-02-02
Omit privacy manifest from source files in podspec

RhyG:
Raised in [this issue](https://github.com/airbnb/lottie-ios/issues/2297), the privacy manifest was being added to the source files due to the wildcard in the podspec which resulted in this warning:

image

I've just updated the `s.source_files` to only include swift files.

### DIFF
--- a/lottie-ios.podspec
+++ b/lottie-ios.podspec
@@ -27,7 +27,7 @@ Lottie enables designers to create and ship beautiful animations without an engi
   s.osx.deployment_target = '10.12'
   s.tvos.deployment_target = '11.0'
 
-  s.source_files = 'Sources/**/*'
+  s.source_files = 'Sources/**/*.swift'
   s.exclude_files = 'Sources/**/*.md'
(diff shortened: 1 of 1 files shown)


lottie-ios.podspec
```diff
@@ -27,7 +27,7 @@ Lottie enables designers to create and ship beautiful animations without an engi
   s.osx.deployment_target = '10.12'
   s.tvos.deployment_target = '11.0'
 
-  s.source_files = 'Sources/**/*'
+  s.source_files = 'Sources/**/*.swift'
   s.exclude_files = 'Sources/**/*.md'
```
calda, line 31: This is unnecessary now that we only include Swift files, right?

```suggestion
```

calda (approved):
Thanks!

lottie-ios.podspec
Brett-Best: Yep, that's right.

PR #2322, opened 2024-02-21
Update minimum deployment target to iOS 13.0 / macOS 10.15

calda:
This PR updates Lottie's minimum deployment target to iOS 13.0 / macOS 10.15.

Our current deployment target is iOS 11.0. Xcode emits this warning when we build the project:

>  The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 11.0, but the range of supported deployment target versions is 12.0 to 17.2.99.

It seems like a good idea to fix that by at least updating to iOS 12.0+.

iOS 13 seems like a good minimum for us because:
 - it's five years old and has high adoption according to https://iosref.com/ios-usage
 - it's the minimum version for many important APIs we use so simplifies the codebase
   - iOS 13 is the minimum version requirement for SwiftUI and Swift `async` / `await`, which we use heavily, and currently results in a lot of `@available` annotations

I'll release a new version with all of the existing changes on master before merging this change. 

Any consumer with a deployment target lower than iOS 13.0 will still be able to use Lottie 4.4.1 and earlier.

### DIFF
--- a/Lottie.xcodeproj/project.pbxproj
+++ b/Lottie.xcodeproj/project.pbxproj
@@ -4087,7 +4087,7 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_KEY_NSHumanReadableCopyright = "";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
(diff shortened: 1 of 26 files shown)


brynbodayle (approved):

vanniktech:
It feels super weird having this bump to iOS 13 in a patch release. Honestly, I would recommend a major upgrade for this since you're essentially breaking compatibility. I still support iOS 12 and now I can't get any bug fixes from 4.4.3 or later.

PR #2348, opened 2024-03-23
Update lottie-ios.podspec

Dukandar:
Support for vision OS


lottie-ios.podspec
```diff
@@ -26,6 +26,8 @@ Lottie enables designers to create and ship beautiful animations without an engi
   s.ios.deployment_target = '13.0'
   s.osx.deployment_target = '10.15'
   s.tvos.deployment_target = '13.0'
+  # === visionOS
```
calda, line 29: We don't need this comment

```suggestion
```

calda (approved):
Thanks!

Dukandar:
> Thanks!

Thanks for approved. If possible could you please merge

PR #2357, opened 2024-04-02
Fix Incorect InitialValue Usage in SwiftUI Wrapper

brynbodayle:
- Fixes an issue where someone is unable to update a LottieView's animation in SwiftUI due to internal usage of the `State(initialValue:)` initializer which only is run once
- Fixes for new Xcode version concurrency warnings

### DIFF
--- a/Sources/Public/Animation/LottieView.swift
+++ b/Sources/Public/Animation/LottieView.swift
@@ -13,7 +13,7 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
 
   /// Creates a `LottieView` that displays the given animation
   public init(animation: LottieAnimation?) where Placeholder == EmptyView {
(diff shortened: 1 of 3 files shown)


calda (approved):
This makes sense to me, thanks!

Sources/Public/Animation/LottieView.swift
```diff
@@ -106,9 +106,10 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     _ loadAnimation: @escaping () async throws -> LottieAnimationSource?,
     @ViewBuilder placeholder: @escaping () -> Placeholder)
   {
+
     self.loadAnimation = loadAnimation
+    self.localAnimation = nil
```
calda, line 111: applying the change the linter is complaining about

```suggestion
    localAnimation = nil
    self.loadAnimation = loadAnimation
```

PR #2376, opened 2024-04-14
Added: PathValueProvider

kalugny:
Hi,
Thank you for the great work on Lottie.

This PR adds the ability to replace a Path in the lottie, similarly to PointValueProvider.

In order to accomplish this I had to make `BezierPath` public.

Thanks,
  Yuval

### DIFF
--- a/Sources/Private/CoreAnimation/Animations/LayerProperty.swift
+++ b/Sources/Private/CoreAnimation/Animations/LayerProperty.swift
@@ -82,7 +82,6 @@ enum PropertyName: String, CaseIterable {
   case strokeWidth = "Stroke Width"
   case gradientColors = "Colors"
   case path = "Path"
-  case maskPath = "Mask Path"
 }
 
 // MARK: CALayer properties
@@ -394,7 +393,7 @@ extension CustomizableProperty {
 
   static var path: CustomizableProperty<CGPath> {
     .init(
-      name: [.path, .maskPath],
+      name: [.path],
       conversion: { typeErasedValue, _ in
         guard let path = typeErasedValue as? BezierPath else { return nil }
         return path.cgPath()
--- a/Sources/Private/CoreAnimation/Animations/LayerProperty.swift
+++ b/Sources/Private/CoreAnimation/Animations/LayerProperty.swift
@@ -219,7 +219,7 @@ extension LayerProperty {
     .init(
       caLayerKeypath: #keyPath(CAShapeLayer.path),
       defaultValue: nil,
-      customizableProperty: .path )
+      customizableProperty: .path)
   }
 
   static var fillColor: LayerProperty<CGColor> {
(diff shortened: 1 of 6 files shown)


Sources/Private/Utility/Primitives/BezierPath.swift
```diff
@@ -10,7 +10,7 @@ import CoreGraphics
 // MARK: - BezierPath
 
 /// A container that holds instructions for creating a single, unbroken Bezier Path.
-struct BezierPath {
+public struct BezierPath {
```
calda, line 13: Instead of making `BezierPath` public, can we add an `Interpolatable` conformance to `CGPath`?

kalugny: From what I gathered, I need the `valueType` of the `ValueProvider` to return `BezierPath`, because that's the underlying type of the property to which it compares itself.

Similarly, I see that `LottieVector3D` from `PointValueProvider` is public.

Sources/Private/CoreAnimation/Animations/LayerProperty.swift
```diff
@@ -390,6 +391,15 @@ extension CustomizableProperty {
       })
   }
 
+  static var path: CustomizableProperty<CGPath> {
```
calda, line 394: How have you been testing this? Can you add a snapshot test case for this functionality? 

You'll need to add a case for this to `SnapshotConfiguration.swift`, you can follow the examples of the existing value provider test cases. If you add a new sample animation file, you can add it at `Tests/Samples/Issues/pr_2376.json`.

kalugny: I encountered an issue with the tests, but I think it may be a problem with how the tests work.

`recordHierarchyKeypath`, which is used to log all the keypaths in use by the Core Animation engine, logs whatever is used in a ValueProvider. This now includes a `Path` property.

The test fails because the `Path` property seems to be missing in the Main Thread engine.

But the PathValueProvider works for both, so I think the test is at fault.

I will add Snapshot tests as you suggested.

Sources/Private/CoreAnimation/ValueProviderStore.swift
```diff
@@ -48,7 +48,10 @@ final class ValueProviderStore {
     context: LayerAnimationContext)
     throws -> KeyframeGroup<Value>?
   {
-    context.recordHierarchyKeypath?(keypath.fullPath)
+    // NOTE: Not recording Path because it messes up tests
```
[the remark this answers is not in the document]
calda: We shouldn't change the code to work around a test failure. Instead we should adjust the test cases as necessary.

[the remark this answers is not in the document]
calda: > The test fails because the Path property seems to be missing in the Main Thread engine.

If this was the issue, then we need to fix how the Path property is set up. Although I didn't see that issue when I pulled your code and ran the tests (I only saw a failure in `testCoreAnimationEngineKeypathLogging`).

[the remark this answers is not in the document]
kalugny: OK, then it seems I misunderstood how the test works.
I'll record the tests and make sure they pass as well as add tests as necessary.

Ismatilloxon (approved):

calda:
Hi @kalugny, since there hasn't been activity on this PR recently I'm going to go ahead and close it. If you're still interested in working on this please feel free to! We can reopen it if/when you find chance to

PR #2379, opened 2024-04-15
Fix issue where animation size could be incorrect after loading async animation

calda:
This PR fixes an issue where the view's size could be incorrect after loading an async animation.

I believe this issue was introduced by #2289, when we changed the lifecycle of the `LottieAnimationView`. Previously it was only added to the view hierarchy after the animation finished loading. Now it's always present, so can have an intrinsic size of zero if measured before the async animation finishes loading.

Here's the sample code I added to test this case. The async case was broken before. Now both work as expected.

```swift
VStack {
  LottieView(animation: .named("Samples/LottieLogo1"))
    .intrinsicSize()
    .looping()

  Text("intrinsic size")
}

VStack {
  LottieView {
    try await Task.sleep(for: .seconds(1))
    return LottieAnimation.named("Samples/LottieLogo1")
  }
  .intrinsicSize()
  .looping()

  Text("intrinsic size, async")
}
```

#### Before

Screenshot 2024-04-15 at 12 09 25 PM

### After

Screenshot 2024-04-15 at 12 09 50 PM

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,6 +33,7 @@ jobs:
       matrix:
         xcode:
         - '15.2' # Swift 5.9
+        - '15.3' # Swift 5.10
     steps:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/setup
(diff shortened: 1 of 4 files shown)


brynbodayle (approved):
Makes sense to me

PR #2400, opened 2024-05-06
Require Xcode 15 / Swift 5.9

calda:
Our policy is that Lottie supports Swift / Xcode versions back to the minimum version that is permitted by Apple for submissions to the App Store. As of April 2024, this is Xcode 15 / Swift 5.9: https://developer.apple.com/news/?id=fxu2qp7b

This PR:
 - updates the minimum Xcode version to Xcode 15 with Swift 5.9
 - removes all CI jobs using Xcode 14
 - re-runs `bundle exec rake format:swift` with the updated minimum Swift version

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,13 +9,12 @@ on:
 jobs:
   build-package-no-visionOS:
     name: "Build Package"
-    runs-on: macos-13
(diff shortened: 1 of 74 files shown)


miguel-jimenez-0529 (approved):
Nice!!

PR #2412, opened 2024-05-21
Fix for watchOS preview build issue

sergeirr:
For some reason Xcode watchOS SwiftUI preview attempts to build this package against watchOS SDK.

Fixed by adding `canImport(QuartzCore)` check to all files, it also might be replaced with `!os(watchOS)`.

Tested with:
- iOS Preview
- watchOS Preview
- iOS build
- watchOS build

Additional info: https://developer.apple.com/forums/thread/731732

[Here is](https://github.com/sergeirr/WatchOSLottie) an example project to reproduce the issue.
You can try to build this package to "Any watchOS" target as well.
CI check step also can be added to build this package to watchOS target.

### DIFF
--- a/Sources/Private/CoreAnimation/Animations/CAAnimation+TimingConfiguration.swift
+++ b/Sources/Private/CoreAnimation/Animations/CAAnimation+TimingConfiguration.swift
@@ -1,6 +1,7 @@
 // Created by Cal Stephens on 1/6/22.
 // Copyright © 2022 Airbnb Inc. All rights reserved.
 
+#if canImport(QuartzCore)
 import QuartzCore
 
 extension CAAnimation {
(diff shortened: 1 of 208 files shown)


calda:
Hi @sergeirr, thank you for this detailed bug report and change. Since this is a workaround for a specific watchOS build bug, Lottie explicitly doesn't support watchOS, and this change introduces non-trivial overhead for future changes, I'd prefer we not merge this.

Does this issue only occur when integrating Lottie via SPM? Have you considered integrating Lottie another way, like by directly embedding the Lottie.xframework archive from the latest release? If another integration option doesn't have this issue I'd recommend using that option instead. Also, maintaining a forked repo with this change could be reasonable.

PR #2415, opened 2024-05-28
Allow configuring specific LottieView properties

freak4pc:
This is just a tiny helper method that makes configuring specific properties on the internal LottieAnimationView a bit nicer.

It means you can directly do:
```swift
LottieView(...)
    .configure(\.contentMode, .scaleAspectFit)
```

Without opening up an entire closure, which feels a bit more ergonomic and clean in most use-cases.

Let me know your thoughts :)

### DIFF
--- a/Example/Example/LottieViewLayoutDemoView.swift
+++ b/Example/Example/LottieViewLayoutDemoView.swift
@@ -9,10 +9,11 @@ struct LottieViewLayoutDemoView: View {
     HStack {
       VStack {
         LottieView(animation: .named("Samples/LottieLogo1"))
+          .configure(\.contentMode, value: .scaleAspectFit)
           .looping()
           .frame(maxWidth: 100)
 
-        Text("maxWidth: 100")
+        Text("maxWidth: 100, contentMode: .scaleAspectFit")
       }
 
       VStack {
--- a/Example/Example/LottieViewLayoutDemoView.swift
+++ b/Example/Example/LottieViewLayoutDemoView.swift
@@ -9,7 +9,7 @@ struct LottieViewLayoutDemoView: View {
     HStack {
       VStack {
         LottieView(animation: .named("Samples/LottieLogo1"))
-          .configure(\.contentMode, value: .scaleAspectFit)
+          .configure(\.contentMode, to: .scaleAspectFit)
           .looping()
           .frame(maxWidth: 100)
 
--- a/Sources/Public/Animation/LottieView.swift
+++ b/Sources/Public/Animation/LottieView.swift
@@ -141,6 +141,16 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     }
   }
 
+  /// Returns a copy of this `LottieView` updated to have the specific configuration property
+  /// applied to its represented `LottieAnimationView` whenever it is updated via the `updateUIView(...)`
+  /// or `updateNSView(...)` methods.
+  public func configure<Property>(
+    _ property: ReferenceWritableKeyPath<LottieAnimationView, Property>,
+    value: Property
+  ) -> Self {
+    configure { $0[keyPath: property] = value }
+  }
+
   /// Returns a copy of this `LottieView` updated to have the given closure applied to its
   /// represented `LottieAnimationView` whenever it is updated via the `updateUIView(…)`
   /// or `updateNSView(…)` method.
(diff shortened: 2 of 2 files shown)


Example/Example/LottieViewLayoutDemoView.swift
```diff
@@ -9,10 +9,11 @@ struct LottieViewLayoutDemoView: View {
     HStack {
       VStack {
         LottieView(animation: .named("Samples/LottieLogo1"))
+          .configure(\.contentMode, value: .scaleAspectFit)
```
freak4pc, line 12: I was also wondering if it would be nicer to omit the labels entirely:

```suggestion
          .configure(\.contentMode, .scaleAspectFit)
```

If you think that's nicer I'm happy to adjust.

Sources/Public/Animation/LottieView.swift
```diff
@@ -141,6 +141,16 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
   }
 
+  /// Returns a copy of this `LottieView` updated to have the specific configuration property
+  /// applied to its represented `LottieAnimationView` whenever it is updated via the `updateUIView(...)`
+  /// or `updateNSView(...)` methods.
+  public func configure<Property>(
+    _ property: ReferenceWritableKeyPath<LottieAnimationView, Property>,
+    value: Property
+  ) -> Self {
+    configure { $0[keyPath: property] = value }
```
calda, line 151: Could you add an overload with a `Property: Equatable` constraint where the value on the `LottieAnimationView` is only updated when the value actually changes?

Sources/Public/Animation/LottieView.swift
```diff
@@ -141,6 +141,16 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
     }
   }
 
+  /// Returns a copy of this `LottieView` updated to have the specific configuration property
+  /// applied to its represented `LottieAnimationView` whenever it is updated via the `updateUIView(...)`
+  /// or `updateNSView(...)` methods.
```
calda, line 146: Let's also mention that this value is applied on every SwiftUI render pass, which may be expensive for any properties that trigger side-effects when updated.

Example/Example/LottieViewLayoutDemoView.swift
freak4pc: Yup, I like it!

Sources/Public/Animation/LottieView.swift
freak4pc: Sounds good 💪

Sources/Public/Animation/LottieView.swift
freak4pc: Isn't this the case for the regular trailing-based closure `configure` ? I copied the description from there. Should we add that note there as well?

calda: That would be great! Although for that case you'd just say something like "the given closure is called on every SwiftUI render pass, which can be expensive for any closures that trigger side effects"

freak4pc:
> Also remember to run `bundle exec rake format:swift` to apply the necessary code changes to satisfy the linter

done, just wanted to confirm this is how you usually format your code? 
image

calda:
Yes, that's the style used by Airbnb's Swift Style Guide (https://github.com/airbnb/swift)

Sources/Public/Animation/LottieView.swift
```diff
@@ -141,9 +141,45 @@ public struct LottieView<Placeholder: View>: UIViewConfiguringSwiftUIView {
...
+    configure { $0[keyPath: property] = value }
+  }
+
+  /// Returns a copy of this `LottieView` updated to have the specific configuration property
+  /// applied to its represented `LottieAnimationView` whenever it is updated via the `updateUIView(...)`
+  /// or `updateNSView(...)` methods.
+  ///
+  /// - note: This configuration will be applied on every SwiftUI render pass.
+  ///         Be wary of applying heavy side-effects as configuration values.
+  ///
```
calda, line 164: This note isn't necessary here because the value us only updated when the value changes

```suggestion
```

calda (approved):

Sources/Public/Animation/LottieView.swift
freak4pc: Good point ;-)

PR #2439, opened 2024-06-19
[Lottie] Support font provider in composition layer

1an-xu:
This PR fixes a bug that font provider doesn't work for text inside the composition layer.

@calda


calda (approved):
Thanks!

calda:
@1an-xu, the linter is failing -- you can run `bundle exec rake format:swift` locally to auto-apply the changes

PR #2444, opened 2024-07-15
Fix existential-any for Swift6

kitwtnb:
Added `any` to each protocols to resolve `existential-any` build errors on Swift 6.  
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0335-existential-any.md


calda:
Thank you for this -- however, the existential `any` syntax is [not going to be required](https://forums.swift.org/t/progress-toward-the-swift-6-language-mode/68315) in Swift 6:

> The Language Steering Group has decided that one previously-accepted upcoming feature, ExistentialAny, will not be enabled by default in Swift 6

I would prefer not to proactively make changes like this if it won't actually be required.

PR #2455, opened 2024-07-29
LottieAnimationView: update `getValue(for:atFrame:)` docs.

GalBerezansky:
The docs of this method weren't updated, this method will not work as expected when the Core Animation rendering engine is used. You can read more about it at the associated discussion: https://github.com/airbnb/lottie-ios/discussions/2398 .

### DIFF
--- a/Sources/Public/Animation/LottieAnimationView.swift
+++ b/Sources/Public/Animation/LottieAnimationView.swift
@@ -677,8 +677,7 @@ open class LottieAnimationView: LottieAnimationViewBase {
   /// Returns nil if no property is found.
   ///
   /// Note: This method isn't supported by the Core Animation rendering engine and will always return `nil` if used.
-  /// It is still supported by the Main Thread rendering engine. You can read more about it at ``currentRenderingEngine``,
-  /// and it can be set through the view's ``configuration``.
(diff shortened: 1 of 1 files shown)


GalBerezansky:
@calda I would appreciate your review on this, Thank you!

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -676,6 +676,10 @@ open class LottieAnimationView: LottieAnimationViewBase {
   /// Reads the value of a property specified by the Keypath.
   /// Returns nil if no property is found.
   ///
+  /// Note: This method isn't supported by the Core Animation rendering engine and will always return `nil` if used.
+  /// It is still supported by the Main Thread rendering engine. You can read more about it at ``currentRenderingEngine``,
+  /// and it can be set through the view's ``configuration``.
+  ///
```
calda, line 682: I think we can leave this as:

```suggestion
  /// Note: This method isn't supported by the Core Animation rendering engine and will always return `nil` if used.
  /// It is still supported by the Main Thread rendering engine.
  ///
```

calda (approved):

PR #2470, opened 2024-08-29
Refactor: Replace '??' operator with custom Optional extensions

bahattinkoc:
### Description:

This pull request refactors the codebase to use the newly introduced Optional extensions, replacing instances of the `??` operator with more readable and expressive methods. The added extensions provide convenient methods such as `orEmpty`, `orTrue`, `orFalse`, and others, to handle common use cases where the optional needs to be unwrapped or given a default value.

### Changes:

Replaced occurrences of the `??` operator across the project with the corresponding custom Optional extensions.
Improved code readability and consistency by utilizing these extensions.
Added additional _safeguards_, like `isNilOrBlank` for strings and `isNilOrEmpty` for collections, to handle _nil_ and _empty_ checks more elegantly.

### DIFF
--- a/Example/Example/AnimationListView.swift
+++ b/Example/Example/AnimationListView.swift
@@ -130,7 +130,7 @@ extension AnimationListView {
     return ((try? fileManager.contentsOfDirectory(
       at: Bundle.main.resourceURL!.appendingPathComponent(directory),
       includingPropertiesForKeys: [.isDirectoryKey],
-      options: [])) ?? [])
+      options: [])).orEmpty)
       .filter { url in
         var isDirectory = ObjCBool(false)
         fileManager.fileExists(atPath: url.path, isDirectory: &isDirectory)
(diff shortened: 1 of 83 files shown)


calda:
`??` is idiomatic Swift code and I would recommend using it over creating one-off helpers. Creating project-specific helpers that replace this standard language pattern is not more readable (it adds a new separate topic that readers have to learn) and is not more expressive (`??` is objectively more expressive since it handles all of these use cases, vs dedicated helpers only handle a single use case).

bahattinkoc:
> `??` is idiomatic Swift code and I would recommend using it over creating one-off helpers. Creating project-specific helpers that replace this standard language pattern is not more readable (it adds a new separate topic that readers have to learn) and is not more expressive (`??` is objectively more expressive since it handles all of these use cases, vs dedicated helpers only handle a single use case).

Thank you for your feedback! I completely understand the value of using the idiomatic `??` operator, as it is indeed a well-known and concise way to handle optionals in Swift. The intention behind introducing these extensions was to address some specific use cases where readability and code consistency could be improved across the project, especially when dealing with common patterns like handling empty collections or strings.

That said, I recognize that introducing custom helpers can add an additional learning curve for those unfamiliar with the codebase. My goal was to enhance expressiveness and reduce repetitive boilerplate code, but I see your point about maintaining familiarity with standard language patterns.

I'm open to discussing further if you believe sticking to ?? would be more beneficial for the team and the project's long-term maintainability.

PR #2471, opened 2024-08-31
Simplify contains(where:) with Key Paths

bahattinkoc:
Swapped out a few `contains(where:)` calls with key path syntax for cleaner code.


calda:
Thanks for the suggestion. Lottie follows [Airbnb's Swift Style Guide](https://github.com/airbnb/swift) which doesn't state a preference between key paths or closures here. I'd prefer to not merge small stylistic tweaks like this.

Issue #920, opened 2019-05-24
is there a way to remove ValueProvider


harryzjm:
I had tried to AnimationView.reloadImages, but it doesn't work.

harryzjm:
I have done this, let it all reload
animateV.animation = animateV.animation
so,  I close this issue.

endy-s:
That doesn't work for me.
Any other tips?
This is my code to set the valueProvider:
```swift
        let keypath = AnimationKeypath(keypath: "**.Color")
        let colorValueProvider = ColorValueProvider(.red.lottieColorValue)
        animationView.setValueProvider(colorValueProvider, keypath: keypath)
```

PR #2474, opened 2024-09-06
Add possibility to remove ValueProvider

batanus:
### Summary
Previously, when the `animation` property was set, all ValueProviders for AnimatioView were reset (see https://github.com/airbnb/lottie-ios/issues/920). Now it's not so (see https://github.com/airbnb/lottie-ios/discussions/2390)

### DIFF
--- a/Sources/Private/CoreAnimation/ValueProviderStore.swift
+++ b/Sources/Private/CoreAnimation/ValueProviderStore.swift
@@ -41,7 +41,7 @@ final class ValueProviderStore {
   }
 
   /// Removes all ValueProviders for the given `AnimationKeypath`
-  func removeValueProvider(keypath: AnimationKeypath) {
+  func removeValueProvider(for keypath: AnimationKeypath) {
       valueProviders.removeAll(where: { $0.keypath == keypath })
   }
 
--- a/Sources/Private/CoreAnimation/ValueProviderStore.swift
+++ b/Sources/Private/CoreAnimation/ValueProviderStore.swift
@@ -40,6 +40,11 @@ final class ValueProviderStore {
     valueProviders.append((keypath: keypath, valueProvider: valueProvider))
   }
 
+  /// Removes all ValueProviders for the given `AnimationKeypath`
+  func removeValueProvider(keypath: AnimationKeypath) {
+      valueProviders.removeAll(where: { $0.keypath == keypath })
+  }
+
   /// Retrieves the custom value keyframes for the given property,
   /// if an `AnyValueProvider` was registered for the given keypath.
   func customKeyframes<Value>(
--- a/Sources/Private/CoreAnimation/ValueProviderStore.swift
+++ b/Sources/Private/CoreAnimation/ValueProviderStore.swift
@@ -42,7 +42,7 @@ final class ValueProviderStore {
 
   /// Removes all ValueProviders for the given `AnimationKeypath`
   func removeValueProvider(for keypath: AnimationKeypath) {
-    valueProviders.removeAll(where: { $0.keypath == keypath })
+    valueProviders.removeAll(where: { $0.keypath.matches(keypath) })
   }
 
   /// Retrieves the custom value keyframes for the given property,
--- a/Sources/Private/CoreAnimation/ValueProviderStore.swift
+++ b/Sources/Private/CoreAnimation/ValueProviderStore.swift
@@ -42,7 +42,7 @@ final class ValueProviderStore {
 
   /// Removes all ValueProviders for the given `AnimationKeypath`
   func removeValueProvider(for keypath: AnimationKeypath) {
-      valueProviders.removeAll(where: { $0.keypath == keypath })
+    valueProviders.removeAll(where: { $0.keypath == keypath })
   }
 
   /// Retrieves the custom value keyframes for the given property,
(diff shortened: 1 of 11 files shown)


Naveen-C-Ramachandrappa (approved):

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -673,6 +673,12 @@ open class LottieAnimationView: LottieAnimationViewBase {
     lottieAnimationLayer.setValueProvider(valueProvider, keypath: keypath)
   }
 
+  /// Sets a ValueProvider for the specified keypath. The value provider will be removed
+  /// on all properties that match the keypath.
+  public func removeValueProvider(keypath: AnimationKeypath) {
```
calda, line 678: A better name for this would probably be:

```suggestion
  public func removeValueProvider(for keypath: AnimationKeypath) {
```

Sources/Private/CoreAnimation/ValueProviderStore.swift
```diff
@@ -40,6 +40,11 @@ final class ValueProviderStore {
     valueProviders.append((keypath: keypath, valueProvider: valueProvider))
   }
 
+  /// Removes all ValueProviders for the given `AnimationKeypath`
+  func removeValueProvider(keypath: AnimationKeypath) {
+      valueProviders.removeAll(where: { $0.keypath == keypath })
```
calda, line 45: What about if the given `keypath` is something like `Layer.*`? Should it remove key paths like `Layer.Color`? It seems like that probably wouldn't happen with the current implementation using `==`.

calda: Could you please add a test case for this to `ValueProvidersTests.swift`, with some cases like that where the the key path being removed isn't directly present in the list of providers and instead matches using wildcards?

calda:
You'll also need to run `bundle exec rake format:swift`, which automatically reformats the code to follow Airbnb's [Swift Style Guide](https://github.com/airbnb/swift).

Sources/Private/CoreAnimation/ValueProviderStore.swift
batanus: Is that behavior really expected? The method is named `removeValueProvider`, which suggests that it should only remove the exact value provider for the specified keyPath, and not multiple `ValueProviders`.
Another point - in `setValueProvider` the value providers are removed the same way as is implemented in this method (see here - https://github.com/airbnb/lottie-ios/blob/master/Sources/Private/CoreAnimation/ValueProviderStore.swift#L39). If you anyway would like to have functionality for wildcard case (like `Layer.*`) - I'm not sure how exactly it should work, so help would be wanted.
I added test for the current `removeValueProvider` implementation. 

Thank you

calda: > Is that behavior really expected?

Good question, I can see it making sense for the behavior to stay as-is in the Core Animation rendering engine.

However, I'm pretty sure that this behavior that I describe is the current behavior implemented in this PR for the Main Thread rendering engine. We should make sure the two implementations are consistent.

Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift
```diff
@@ -241,6 +241,17 @@ final class MainThreadAnimationLayer: CALayer, RootAnimationLayer {
     }
   }
 
+  func removeValueProvider(for keypath: AnimationKeypath) {
+    for layer in animationLayers {
+      if let foundProperties = layer.nodeProperties(for: keypath) {
```
calda, line 246: Like I mentioned in https://github.com/airbnb/lottie-ios/pull/2474#discussion_r1750299310, I think this behavior is different than the behavior implemented for the Core Animation rendering engine in `ValueProviderStore`.

`layer.nodeProperties(for: keypath)` returns all of the properties that match the given key path, so for example if you add a provider with `Layer.Color` and then call `removeValueProvider(for: "**.Color")`, this implementation will remove the provider for `Layer.Color` (but the implementation in the Core Animation rendering engine will not).

calda: I don't have a strong preference on which approach is better (removing all matches, or just removing exact matches) but I think the behavior of the Core Animation engine and Main Thread engine needs to be the same.

calda: I have a feeling updating the behavior of `ValueProviderStore` to check for matches (using `keypath.matches(...)` instead of `==`) would be easier than trying to change the implementation here to check for exact matches on the original input key path (which I don't think we have access to anymore).

batanus: Thanks, @calda, for your assistance!

calda (approved):
Thanks for the update and test! Just one last comment before we merge this: https://github.com/airbnb/lottie-ios/pull/2474/files#r1752184422

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -854,6 +854,13 @@ public class LottieAnimationLayer: CALayer {
     animationLayer.setValueProvider(valueProvider, keypath: keypath)
   }
 
+  public func removeValueProvider(for keypath: AnimationKeypath) {
+    guard let animationLayer = rootAnimationLayer else { return }
+
+    valueProviders[keypath] = nil
```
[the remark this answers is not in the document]
batanus: The `removeAll(where:)` method isn't available for Dictionaries, so I used `.forEach` with a necessary check. Thank you 👍

calda (approved):
Thank you!

PR #2478, opened 2024-09-12
Trim shape default type value

alexallah:
I created a .lottie animation using a third party tool and I tried to render it using lottie-ios. Unfortunately, it logged an error and didn't display the file.

However, my animation was rendering just fine in the browser. After some digging, it turns out that the the trim shape dictionary decoder requires it to have a type specified. The JavaScript implementation (and probably others) doesn't require it and sets default value when not found:
https://github.com/LottieFiles/lottie-js/blob/a1bc6c5668323161a08159a2e040ff0cf14ab46c/src/shapes/trim-shape.ts#L40

In this PR, I set default trim type to `individual`. Similar to the JavaScript implementation.


Sources/Private/Model/ShapeItems/Trim.swift
```diff
@@ -34,7 +34,7 @@ final class Trim: ShapeItem {
     end = try KeyframeGroup<LottieVector1D>(dictionary: endDictionary)
     let offsetDictionary: [String: Any] = try dictionary.value(for: CodingKeys.offset)
     offset = try KeyframeGroup<LottieVector1D>(dictionary: offsetDictionary)
-    let trimTypeRawValue: Int = try dictionary.value(for: CodingKeys.trimType)
+    let trimTypeRawValue: Int = try dictionary.value(for: CodingKeys.trimType) ?? TrimType.individually.rawValue
```
calda, line 37: Please also make this change in the `Decodable` init above

calda:
Could you add a sample json animation using this functionality as `Tests/Samples/Issues/pr_2478.json`, and then generate snapshot images by re-running the tests with `bundle exec rake test:package`

calda:
You have to record snapshots for the new sample animation, that's why the tests are failing :)

calda:
I pushed a change that adds the snapshots.

calda (approved):
Thanks!

PR #2481, opened 2024-09-18
- iOS 18 and latest xCode crash fix with

CricHeroesiOS:
Replacing withCheckedThrowingContinuation Calls With withUnsafeThrowingContinuation


calda:
It sounds like this was an issue affecting iOS 18 betas, which has since been fixed: https://developer.apple.com/forums/thread/761580

Since this is an SDK-level issue, I'd prefer we not try to work around it in Lottie library code.

PR #2507, opened 2024-12-05
Add additional dynamic library product to Package.swift

JensK611:
Hi Lottie-Team,

I would like to link the Lottie framework dynamically to use it inside another framework. 
To avoid duplicate symbols the library should not embedded in the framework target and only embedded in the app target.

To solve this issue (since SPM Dependencies are automatically built statically) another product has been added to the `Package.swift` file allowing to build a dynamic framework. This seems to be a common practice as it is used by the SnapKit and RxSwift framework.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,10 @@ let package = Package(
   name: "Lottie",
   // Minimum platform versions should be kept in sync with the per-platform targets in Lottie.xcodeproj, lottie-ios.podspec, and lottie-spm's Package.swift
   platforms: [.iOS("13.0"), .macOS("10.15"), .tvOS("13.0"), .custom("visionOS", versionString: "1.0")],
-  products: [.library(name: "Lottie", targets: ["Lottie"])],
+  products: [
+    .library(name: "Lottie", targets: ["Lottie"]),
+    .library(name: "Lottie-Dynamic", type: .dynamic, targets: ["Lottie"]),
+  ],
   dependencies: [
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -5,7 +5,10 @@ let package = Package(
   name: "Lottie",
   // Minimum platform versions should be kept in sync with the per-platform targets in Lottie.xcodeproj, lottie-ios.podspec, and lottie-spm's Package.swift
   platforms: [.iOS("13.0"), .macOS("10.15"), .tvOS("13.0"), .custom("visionOS", versionString: "1.0")],
-  products: [.library(name: "Lottie", targets: ["Lottie"])],
+  products: [
+    .library(name: "Lottie", targets: ["Lottie"]),
+    .library(name: "Lottie-Dynamic", type: .dynamic, targets: ["Lottie"]),
```
calda, line 10: Can you add a comment here explaining the difference between Lottie and Lottie-Dynamic, and a quick explanation of when you would prefer to use Lottie-Dynamic instead of Lottie?

calda (commented):
Seems reasonable to me

calda (approved):
Thanks!

roman-suvorov-onseo:
Hey! This is precisely what we need.
Would you kindly tell us when we can expect a public release with this change to happen?

P.S.
Maybe it makes sense to make a cherry-pick to make a small patch-release with only this change, based on the current release version?

calda:
Sure, I just released 4.5.1: https://github.com/airbnb/lottie-ios/releases/tag/4.5.1

PR #2509, opened 2024-12-06
Fix Haiku URL in README

hmarr:
The current URL in the README points to a domain for sale. This updates it to the project's current website. That said, it does look like the project has been inactive for some time, so removing it entirely would also be an option.


calda (approved):
Thanks!

PR #2559, opened 2025-04-28
Support customizing `contentsScale` of text layer

calda:
This PR adds support for customizing the `contentsScale` of text layers, to control text rendering sharpness.

Text in the animation below is aliased poorly because of how the text layer is being rasterized at a high resolution and then downscaled. As a workaround, we can reduce the `contentsScale` of the layer in our feature code.

### Before

Screenshot 2025-04-25 at 2 07 20 PM

### After, with customized `contentsScale = 0.25`

Screenshot 2025-04-25 at 2 06 54 PM (1)


Tests/SnapshotConfiguration.swift
```diff
@@ -11,10 +11,9 @@ import UIKit
 struct SnapshotConfiguration {
   /// The precision that should be used when comparing the
   /// captured snapshot with the reference image in `Tests/__Snapshots`
-  ///  - Defaults to 1.0 (the snapshot must match exactly).
   ///  - This can be lowered for snapshots that render somewhat nondeterministically,
   ///    but should be kept as high as possible (while still permitting the diff to succeed)
-  var precision: Float = 1
+  var precision: Float = 0.985
```
miguel-jimenez-0529, line 16: Should we add a comment about why this is not 1?

miguel-jimenez-0529 (approved):
Great addition!

Tests/SnapshotConfiguration.swift
calda: Good idea

Tests/SnapshotConfiguration.swift
```diff
@@ -11,10 +11,9 @@ import UIKit
 struct SnapshotConfiguration {
   /// The precision that should be used when comparing the
   /// captured snapshot with the reference image in `Tests/__Snapshots`
-  ///  - Defaults to 1.0 (the snapshot must match exactly).
   ///  - This can be lowered for snapshots that render somewhat nondeterministically,
   ///    but should be kept as high as possible (while still permitting the diff to succeed)
```
calda, line 15:
```suggestion
  ///    but should be kept as high as possible (while still permitting the diff to succeed)
  /// - Defaults to 0.985 to reduce the number of flaky snapshots due to small differences
  ///    in rendering based on OS version.
```

Issue #2546, opened 2025-03-18
Lottie Animation Not Rendering on iOS 18


ViktoriaSaklakova:
Description

After updating to iOS 18, Lottie animations fail to render properly in my app. The same animations work correctly on earlier iOS versions (e.g., iOS 17 and below). I have already updated to the latest Lottie version (4.5.1), but the issue persists.
Expected Behavior

Lottie animations should render consistently across iOS versions, including iOS 18.

[job_done.json](https://github.com/user-attachments/files/19315783/job_done.json)

Additional Notes

    I tested the animation on iOS 17 and below, and it works fine.
    The animation plays correctly on LottieFiles Preview.

Actual result: checkmark is waving

![Image](https://github.com/user-attachments/assets/2912c52e-8649-4348-8b5e-20d783b08cc4)

Issue #2548, opened 2025-03-24
Lottie Animation Not Rendering


sapphirezzz:
[diagnosis_background.json](https://github.com/user-attachments/files/19426283/diagnosis_background.json)

PR #2560, opened 2025-04-28
Fix CoreAnimation issues in iOS 18 Patch

tinder-jasonpicallos:
This PR addresses any iOS 18 CoreAnimation regression issues which result in slow/dropped frames and buggy animations. 
Fixes:  #2546  ,  #2548 

**Context:**
   iOS 18 CAKeyFrameAnimation no longer advances beyond certain frames due to an apple compositor bug. Setting to the .mainThread render engine will do a manual frame by frame pass via CADisplayLink and avoids the broken GPU path. 
   
 **Changes**
  A specific version check for iOS 18 will serve as a temporary patch to resolve all animation latencies, this is done by setting the render engine explicitly to .mainThread
  
 **Impact**
 On iOS 18 animations use CPU driven fallback and improved animation renders, on versions before **no change.**
 
 
 **MEDIA**
 
 BUG VIDEO (iOS 18): 
 

https://github.com/user-attachments/assets/f632f9f0-270b-4085-8082-04a16b56ba40

FIX VIDEO (iOS 18):

https://github.com/user-attachments/assets/08fe9312-d9cc-427c-93cf-9bf5959c672c

### DIFF
--- a/Sources/Public/Configuration/LottieConfiguration.swift
+++ b/Sources/Public/Configuration/LottieConfiguration.swift
@@ -24,7 +24,15 @@ public struct LottieConfiguration: Hashable {
 
   /// The global configuration of Lottie,
   /// which applies to all `LottieAnimationView`s by default.
-  public static var shared = LottieConfiguration()
+    ///
+    public static var shared: LottieConfiguration = {
+        var configuration: LottieConfiguration = .init()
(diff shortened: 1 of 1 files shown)


calda:
I don't think we should globally change the default like this, since that will have potential compatibility impacts for all users.

I would suggest configuring this within your app instead.

Issue #2485, opened 2024-09-20
[FR] Add an static xcframework to tags.


jesus-mg-ios:
## Which Version of Lottie are you using?

Lottie 4.5.0

## Expected Behavior 
Have a way to download precompiled static xcframework. 

## Actual Behavior
The xcframework right now is dynamic, you can check it with otool.

PR #2561, opened 2025-05-09
Adds static framework generation on CI

Loupehope:
This pull-request adds static framework generation on CI.

Fixes: https://github.com/airbnb/lottie-ios/issues/2485 and https://github.com/airbnb/lottie-ios/discussions/2486

Additionally later, can be added to https://github.com/airbnb/lottie-spm

I would be grateful if you could double-check the code, I don't work with ruby that often.

Example output of a type of framework for dynamic linking:
Screenshot 2025-05-09 at 18 51 49

Example of framework type output for static linking:
Screenshot 2025-05-09 at 18 52 17

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -111,18 +111,13 @@ jobs:
           xcode: ${{ matrix.xcode }}
       - name: Build Dynamic XCFramework
         run: bundle exec rake build:xcframework\['dynamic'\]
-      - name: Upload Dynamic XCFramework
-        uses: actions/upload-artifact@v4
-        with:
-          name: BuildProducts
-          path: .build/archives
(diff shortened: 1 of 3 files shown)


calda:
What use case do you have in mind for a static XCFramework?

Loupehope:
We use Tuist for development, which suggests using a static framework for SPM dependencies by default - https://docs.tuist.dev/en/guides/develop/projects/dependencies#tuists-xcodeprojbased-integration.

In addition, the release application is built in static to avoid wasting time loading dynamic libraries on startup.

It seemed to me that it would be convenient for such a set-up to have a static framework in addition to the pre-built dynamic framework.

calda:
Sounds good, thanks for the PR

PR #2563, opened 2025-05-13
Fix: Prevent animationSpeed = 0 from blocking playback (#2549)

### LINKED ISSUE #2549: Last frame of animation is missing
I’m encountering an issue where the last frame of a Lottie animation does not render on iOS. The same .json works correctly on both Web and Android platforms.

**Expected Behavior:**
The animation should render completely, including the final frame (op) defined in the Lottie JSON.

**Actual Behavior:**
On iOS, the animation stops one frame early, missing the final frame.

**Tested On:**
	✅ Web (working)
	✅ Android (working)
	❌ iOS (not rendering last frame)

**Environment:**
     https://github.com/airbnb/lottie-spm.git
	•	lottie iOS-version: [e.g., 4.5.0]
	•	iOS Version: [e.g., iOS 17.3, 18...]
	•	Device: [e.g., iPhone 14, iPhone 13 Pro...]

https://github.com/user-attachments/assets/89ff4dac-1ef5-4abb-af18-fa1d4420614e

https://github.com/user-attachments/assets/26337cd2-5039-4f3a-9681-545aa4679191

[339.json](https://github.com/user-attachments/files/19461363/339.json)

guavacadodev:
## Summary

Fixes a bug where setting `animationSpeed = 0` before calling `play()` would prevent the animation from starting, even if the speed is later changed to a non-zero value.

## Problem

When `animationSpeed == 0`, Lottie tries to start a `CABasicAnimation` with zero speed, which causes the animation not to play and prevents progress.

## Solution

- Prevent scheduling the animation if `animationSpeed == 0`.
- Store the playback context and resume when the speed is later set to a valid value.
- Updated `animationSpeed`'s `didSet` to resume any pending animation.

## Testing

Tested by:
- Manually setting `animationSpeed = 0`, calling `play()`, then restoring the speed to `1.0` after a short delay. Animation now plays as expected.

### DIFF
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -728,11 +728,15 @@ public class LottieAnimationLayer: CALayer {
   }
 
   /// Sets the speed of the animation playback. Defaults to 1
-  public var animationSpeed: CGFloat = 1 {
-    didSet {
-      updateInFlightAnimation()
+    public var animationSpeed: CGFloat = 1 {
+        didSet {
+            // If a previous animation was queued but not played due to zero speed, trigger it now
+            if animationContext != nil, animationSpeed != 0, !isAnimationPlaying {
+                updateInFlightAnimation()
+            }
+        }
     }
-  }
+
 
   /// When `true` the animation will play back at the framerate encoded in the
   /// `LottieAnimation` model. When `false` the animation will play at the framerate
@@ -1439,7 +1443,10 @@ public class LottieAnimationLayer: CALayer {
     let layerAnimation = CABasicAnimation(keyPath: "currentFrame")
     layerAnimation.fromValue = playFrom
     layerAnimation.toValue = playTo
-    layerAnimation.speed = Float(animationSpeed)
+      if animationSpeed == 0 {
+          self.animationContext = animationContext
+          return
+      }
     layerAnimation.duration = TimeInterval(duration)
     layerAnimation.fillMode = CAMediaTimingFillMode.both
     layerAnimation.repeatCount = loopMode.caAnimationConfiguration.repeatCount
--- a/Tests/AnimationViewTests.swift
+++ b/Tests/AnimationViewTests.swift
@@ -70,6 +70,28 @@ final class AnimationViewTests: XCTestCase {
 
     wait(for: [expectation], timeout: 1.0)
   }
+    
+    func testPlayWithSpeedZeroThenResume() {
+        let animation = LottieAnimation.named("LottieLogo1")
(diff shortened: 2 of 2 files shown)


Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1439,7 +1443,10 @@ public class LottieAnimationLayer: CALayer {
     let layerAnimation = CABasicAnimation(keyPath: "currentFrame")
     layerAnimation.fromValue = playFrom
     layerAnimation.toValue = playTo
-    layerAnimation.speed = Float(animationSpeed)
```
calda, line 1442: Why would we remove `layerAnimation.speed = Float(animationSpeed)`? That sounds like it will break something.

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1439,7 +1443,10 @@ public class LottieAnimationLayer: CALayer {
...
+      if animationSpeed == 0 {
+          self.animationContext = animationContext
+          return
```
calda, line 1448: Why is this here?

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -728,11 +728,15 @@ public class LottieAnimationLayer: CALayer {
   }
 
   /// Sets the speed of the animation playback. Defaults to 1
-  public var animationSpeed: CGFloat = 1 {
-    didSet {
-      updateInFlightAnimation()
+    public var animationSpeed: CGFloat = 1 {
+        didSet {
+            // If a previous animation was queued but not played due to zero speed, trigger it now
+            if animationContext != nil, animationSpeed != 0, !isAnimationPlaying {
```
calda, line 734: Why do we need the additional `if animationContext != nil, animationSpeed != 0, !isAnimationPlaying {` condition? The comment seems to talk about when we need to call `updateInFlightAnimation()`, but previously `updateInFlightAnimation()` was always called.

Is there as case where we _shouldn't_ be calling `updateInFlightAnimation()` that is now fixed by this change?

calda (commented):
Please include a video recording showing a demo that previously did not work correctly, but now does work correctly.

guavacadodev:
Hey calda, sorry about that. I am a bit new to Github and still learning how it works, trying to figure out how to fork, clone, and make contributions. 

I'd love to help out with the issue if I can. I will continue practicing in some other repositories.

calda:
Sounds good, thanks. Maybe check out another project I work on: github.com/nicklockwood/SwiftFormat. Contributions are always welcome 👌🏻

guavacadodev:
Great thanks !

PR #2572, opened 2025-06-08
Fix: Remove unused variable always set to false

ohsg0226:
This PR removes the `waitingToPlayAnimation` variable.  
The variable's default value was `false`, and there were no places in the code where it was ever set to `true`.

### DIFF
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -1023,9 +1023,11 @@ public class LottieAnimationLayer: CALayer {
     }
   }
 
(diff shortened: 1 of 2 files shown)


Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -1023,11 +1023,9 @@ public class LottieAnimationLayer: CALayer {
     }
   }
 
-  public func updateAnimationForForegroundState(wasWaitingToPlayAnimation: Bool) {
```
calda, line 1026: This is a public API so changing it would be a breaking change

PR #2598, opened 2025-08-24
Fix: isAnimationPlaying is false when backgroundBehavior is continuePlaying

### LINKED ISSUE #2597: `isAnimationPlaying` becomes false when animation continues in background
# 1. Environment
- Lottie iOS version: 4.5.2
- iOS version: 16.4
- Device: iPhone 8 / Simulator

# 2. Steps to Reproduce
- Create LottieAnimationView and set `backgroundBehavior = .continuePlaying`
- Start a Lottie animation in LottieAnimationView.
- Put the app in background.
- Return to foreground.
- Observe LottieAnimationView.isAnimationPlaying is false even though animation is visually still running.

### 📹 Video demo
https://github.com/user-attachments/assets/a79ddbc6-e48b-42be-8230-08ff34511ab4

### Example project
https://github.com/sunnysideupdowntown/LottieBugExample

# 3. Expected Behavior
isAnimationPlaying should remain true as long as the animation is visually playing.

# 4. Actual Behavior
isAnimationPlaying is set to false when the app returns from background.

# 5. Notes / Workaround
The issue occurs because CoreAnimationLayer's `timedProgressAnimation.isRemovedOnCompletion = true` so when the app go to background, animation is removed. But another layers inside CoreAnimationLayer are set to false, so they are not removed and still playing.

sunnysideupdowntown:
# Description
This PR fixes an issue where isAnimationPlaying becomes false even though the animation is still running.
- Root cause: The issue occurs because CoreAnimationLayer's timedProgressAnimation.isRemovedOnCompletion = true so when the app go to background, animation is removed. But another layers inside CoreAnimationLayer are set to false, so they are not removed and still playing.
- Solution: ensure isAnimationPlaying is properly set when playback continues.

# Changes
Fixed logic to correctly maintain isAnimationPlaying value
Added tracking using animationStartedAt to verify real-time playback

# Related Issue
Fixes #2597 

# Video / Demo
https://github.com/user-attachments/assets/81d17607-d8d7-4ce1-843f-89e552a4aebb

### DIFF
--- a/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
+++ b/Sources/Private/CoreAnimation/CoreAnimationLayer.swift
@@ -90,7 +90,8 @@ final class CoreAnimationLayer: BaseAnimationLayer {
       let durationMultiplier = timingConfiguration.autoreverses ? 2.0 : 1.0
       let speed = Double(timingConfiguration.speed)
       let repeatCount = Double(timingConfiguration.repeatCount)
-      return animationDuration * durationMultiplier / speed * repeatCount
+      let timeOffset = timingConfiguration.timeOffset
+      return animationDuration * durationMultiplier / speed * repeatCount - timeOffset
     }
 
     static func ==(_ lhs: AnimationConfiguration, _ rhs: AnimationConfiguration) -> Bool {
--- a/Tests/AnimationConfigurationTests.swift
+++ b/Tests/AnimationConfigurationTests.swift
@@ -64,4 +64,15 @@ final class AnimationConfigurationTests: XCTestCase {
 
     XCTAssertEqual(config.expectedAnimationDuration, 6.0, accuracy: 0.0001)
   }
+
+  func testExpectedDuration_ComplexContext() {
+    let context = AnimationContext(playFrom: 30, playTo: 300, framerate: 30, closure: nil)
+    let timing = CoreAnimationLayer.CAMediaTimingConfiguration(autoreverses: true, repeatCount: 3, speed: 2, timeOffset: 2)
+    let config = CoreAnimationLayer.AnimationConfiguration(
+      animationContext: context,
+      timingConfiguration: timing,
+      recordHierarchyKeypath: nil)
+
+    XCTAssertEqual(config.expectedAnimationDuration, 25.0, accuracy: 0.0001)
+  }
 }


Sources/Private/CoreAnimation/CoreAnimationLayer.swift
```diff
@@ -343,15 +348,29 @@ extension CoreAnimationLayer: RootAnimationLayer {
...
+      return true
+
     case .paused:
-      false
+      return false
+
     case nil:
       switch playbackState {
       case .playing:
-        animation(forKey: #keyPath(animationProgress)) != nil
```
calda, line 352: This is clever, but changing it to no longer use the layer itself as the source of truth sounds like it will cause unexpected behavior changes. This is also the same approach used by the Main Thread rendering engine.

Is there another way to fix this issue without modifying this check? Perhaps we could update the specific call site of this check that is related to the bug you are investigating?

calda: I'm open to trying it though. I see how this is probably the only solution to the issue.

Sources/Private/CoreAnimation/CoreAnimationLayer.swift
```diff
@@ -343,15 +348,29 @@ extension CoreAnimationLayer: RootAnimationLayer {
...
+        guard let animationStartedAt, let currentAnimationConfiguration else { return false }
+        if currentAnimationConfiguration.timingConfiguration.repeatCount == .greatestFiniteMagnitude {
+          return true
+        }
+
+        let animationDuration = Double(
+          currentAnimationConfiguration.animationContext.playTo -
+            currentAnimationConfiguration.animationContext.playFrom) / animation.framerate
+        let speed = Double(currentAnimationConfiguration.timingConfiguration.speed)
+        let repeatCount = Double(currentAnimationConfiguration.timingConfiguration.repeatCount)
```
calda, line 368: Please factor this logic into a method like `var expectedAnimationDuration: TimeInterval`. It would be great if you could include unit tests for the logic.

calda, line 368: What about `autoreverses = true`? I think that would double the playback duration

sunnysideupdowntown: I missed it. Applied the fix here -> [8dceb0f](https://github.com/airbnb/lottie-ios/pull/2598/commits/8dceb0f125fd93feee927473197eb0170243005a)

sunnysideupdowntown: After writing the unit test, I realized that I also need to consider `timeOffset`.
I think I’ll need some time to dig into it, so maybe I can handle it next weekend.

sunnysideupdowntown: To handle this, I added `framerate` in AnimationContext. [f9577c0](https://github.com/airbnb/lottie-ios/pull/2598/commits/f9577c068296d56cd3d5e43b21773bd64d0835f2)
Then added `expectedAnimationDuration` in AnimationConfiguration. [ce97167](https://github.com/airbnb/lottie-ios/pull/2598/commits/ce971679d7ada989b3a9300e1dc01345b6884126)
And used it. [c699253](https://github.com/airbnb/lottie-ios/pull/2598/commits/c699253ba26046176b450068c2bc8b47ba32eafd)

I've written some unit tests, but since I'm new to them, I'm not entirely sure if they're correct. [4d5b08e](https://github.com/airbnb/lottie-ios/pull/2598/commits/4d5b08e9c2aa5e81b73ceb513d4464fe575d2cb0)

sunnysideupdowntown: @calda 
I’ve updated the expectedAnimationDuration calculation to include timeOffset and added related unit tests. 
[b5e4467](https://github.com/airbnb/lottie-ios/pull/2598/commits/b5e4467bf9a307af0730a37e66bb2f092da3c6fd)

PR #2603, opened 2025-09-25
Invoke `animationDidLoad` on next run loop

miguel-jimenez-0529:
I notice an issue where a consumer of `LottieView` will update a SwiftUI `@State` variable in the `animationDidLoad` closure. 

```swift
@State var animationLoaded = false

LottieView {
 ...
}
.animationDidLoad { _ in
 // triggers a "Modifying state during view update, this will cause undefined behavior"
 animationLoaded = true
}
```
Updating state is a common pattern so we should fix it in the view


calda (approved):
Thanks!

PR #2639, opened 2026-01-09
Add padding around text layer to compensate for buggy fonts

### LINKED ISSUE #2638: Text is clipped when using certain fonts that report wrong geometry
Texts that use certain fonts (in this example, Venezuela) get clipped, due to the bounds being too small.
Example - words Captions, and This. Red borders are around CoreText layers.

Image
Image

In the current PR:

Image
Image

Attached example animation (needs external fonts)
[captions-style-18.json](https://github.com/user-attachments/files/24530875/captions-style-18.json)

eXCoreX:
Fixes #2638

### DIFF
--- a/Sources/Private/MainThread/LayerContainers/Utility/CoreTextRenderLayer.swift
+++ b/Sources/Private/MainThread/LayerContainers/Utility/CoreTextRenderLayer.swift
@@ -196,7 +196,7 @@ final class CoreTextRenderLayer: CALayer {
       }
 
     // This fixes a vertical padding issue that arises when drawing some fonts.
-    // For some reason some fonts, such as Helvetica draw with and ascender that is greater than the one reported by CTFontGetAscender.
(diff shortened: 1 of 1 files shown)


rostislav-litvinov-ek:
@calda There's a small difference in the only unpassing test - due to the width being different, the core text decided to shift some letters around, I don't think it's significant

Sources/Private/MainThread/LayerContainers/Utility/CoreTextRenderLayer.swift
```diff
@@ -346,21 +362,22 @@ final class CoreTextRenderLayer: CALayer {
         CGSize(width: CGFloat.greatestFiniteMagnitude, height: CGFloat.greatestFiniteMagnitude),
         nil
       )
+      let adjustedSize = CGSize(width: size.width + compensationPadding * 2, height: size.height)
```
calda: A comment here explaining what `adjustedSize` is for would be useful

calda (approved):
Thanks!

PR #2646, opened 2026-02-02
Fix animation state lost when view removed from window in SwiftUI

kirillsh:
## Summary

  - Fix animation state being lost when a SwiftUI view is removed from the window hierarchy (e.g. `NavigationStack`, `TabView`, sheets, cell reuse)
  - Add `windowBackgroundBehavior` property to configure animation behavior for window attachment changes independently from app foreground/background transitions
  - Expose `.windowBackgroundBehavior(_:)` modifier on `LottieView` for SwiftUI usage

## Motivation

Previously, `backgroundBehavior` controlled both app background transitions and window detach/attach events. In SwiftUI, views are frequently removed from the window during navigation, which triggered the same stop/pause behavior as backgrounding the app. This made it impossible to keep animations running through navigation without also keeping them running when the app is backgrounded.

Related: #1332

### DIFF
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -487,23 +487,6 @@ public class LottieAnimationLayer: CALayer {
   /// Will inform the receiver the type of rendering engine that is used for the layer.
   public var animationLayerDidLoad: ((_ animationLayer: LottieAnimationLayer, _ renderingEngine: RenderingEngineOption) -> Void)?
 
-  /// Behavior when the view is added to or removed from the window hierarchy.
-  ///
-  /// Views may be removed from the window hierarchy in scenarios such as:
-  /// - SwiftUI navigation (`NavigationStack`, `TabView`, `LazyVStack`, sheets)
-  /// - `UICollectionView` / `UITableView` cell reuse
-  /// - `UIPageViewController` page transitions
-  /// - Custom container view controllers
-  ///
-  /// By default, this uses the same value as `backgroundBehavior`. Set this property
-  /// when you need different behavior for window attachment changes versus app
-  /// foreground/background transitions.
-  ///
-  /// For example, use `.continuePlaying` here with the Core Animation rendering engine
-  /// if you want animations to seamlessly resume after navigation, while still using
-  /// a different `backgroundBehavior` for app background transitions.
-  public var windowBackgroundBehavior: LottieBackgroundBehavior?
-
   /// The configuration that this `LottieAnimationView` uses when playing its animation
   public var configuration: LottieConfiguration {
     didSet {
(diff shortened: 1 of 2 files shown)


Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -996,43 +996,57 @@ public class LottieAnimationLayer: CALayer {
...
+    } else if backgroundBehavior == .pauseAndRestore {
+      restorePreservedAnimation(startIfWaiting: false)
     }
   }
 
+  /// Called when view is removed from window hierarchy (NOT app background).
+  /// Always preserves animation state for restore, regardless of `backgroundBehavior`.
+  /// This fixes SwiftUI compatibility where views are frequently removed/re-added
+  /// during transitions, navigation, or conditional rendering.
+  public func updateAnimationForWindowDisappear() {
```
calda: Do these need to be public?

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -996,43 +996,57 @@ public class LottieAnimationLayer: CALayer {
...
+    guard animationContext != nil else { return }
+    pauseAndPreserveAnimation()
```
calda: Shouldn't we respect the `backgroundBehavior`?

Sources/Public/Animation/LottieAnimationView.swift
```diff
@@ -1034,9 +1034,9 @@ open class LottieAnimationView: LottieAnimationViewBase {
     guard superview != nil else { return }
 
     if window != nil {
-      updateAnimationForForegroundState()
+      updateAnimationForWindowAppear()
     } else {
-      updateAnimationForBackgroundState()
+      updateAnimationForWindowDisappear()
     }
```
calda: It seems like previously this would respect the `backgroundBehavior`, and now no longer respects the `backgroundBehavior`. I think we should continue respecting the background behavior, realistically. Do you need the ability to configure the background behavior for `animationMovedToWindow` differently from the app backgrounding behavior?

Sources/Public/Animation/LottieAnimationLayer.swift
```diff
@@ -996,43 +996,57 @@ public class LottieAnimationLayer: CALayer {
...
-      } else if backgroundBehavior == .pauseAndRestore {
-        /// Restore animation from saved state
-        updateInFlightAnimation()
-      }
+    guard animationContext != nil else { return }
+
+    if wasWaitingToPlayAnimation {
+      restorePreservedAnimation(startIfWaiting: true)
+    } else if backgroundBehavior == .pauseAndRestore {
+      restorePreservedAnimation(startIfWaiting: false)
```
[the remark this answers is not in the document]
kirillsh: I'll rename param to `startFromBeginning` to make the intent more clear. The actual meaning for this property is either 'continue playing from previous position' (for false) or 'start from the beginning' (for true)

PR #2676, opened 2026-04-15
Fix incorrect mask animation timing on precomp layers

RohitMaru:
- Fix incorrect mask animation timing on precomp layers when using the CoreAnimation rendering engine
  - Mask keyframes are defined in the parent's global timeline, but were receiving the precomp's time-remapped child context, shifting them by `startTime` frames

  **Changes:**
  - `BaseAnimationLayer.setupAnimations`: iterate `sublayers ?? []` instead of `managedSublayers`, so the mask layer is no longer included in the time-remapped child context pass
  - `BaseCompositionLayer.setupLayerAnimations`: explicitly invoke `maskLayer.setupAnimations(context:)` with the parent's own context after drop shadow setup, matching the Android behavior where masks receive global progress and children receive local/remapped progress

  **Android reference:**
  Android's `CompositionLayer.setProgress` already does this correctly — masks get the global progress via `super.setProgress`, while child layers get the local
  time-remapped progress.


calda:
Can you please add a sample animation that renders incorrectly without this change to the Samples directory, and then generate the regression testing snapshots for it?

RohitMaru:
Lottie file to repro it: https://drive.google.com/file/d/1EeXSJtg7AE4sUneq9GH9rMOS88ckaxSI/view?usp=sharing is the lottie file. 

**_Observe the timer and the animation delay_**

Before fix:
https://github.com/user-attachments/assets/35308982-c051-4e11-9f1c-8476fa2aaef0


After fix:
https://github.com/user-attachments/assets/eac4171a-5340-442b-bfef-b7fb296df487

RohitMaru:
Hi @calda , can you share details on how to run these snapshot tests? I ran them locally to verify though. Thanks

calda:
You can run the Snapshot Tests tests in Xcode, it will generate any new files as needed

RohitMaru:
@calda  Done. Please review, thanks.

calda (approved):

RohitMaru:
Thanks a ton!

PR #2677, opened 2026-04-23
Value Providers Not Working Correctly When Setting Gradient Colors

### LINKED ISSUE #2616: Value Providers Not Working Correctly with SwiftUI LottieView
## Bug Description
When using value providers (specifically `GradientValueProvider`) with the native SwiftUI `LottieView`, gradients are not rendered correctly. The same code works perfectly fine with UIKit's `LottieAnimationView`.

## Environment
- Lottie Version: 4.5.2
- iOS Version: 26.0 & 18.2
- Xcode Version: 26.1.1

## Expected Behavior
Gradient colors should be applied to the animation using `GradientValueProvider`, matching the behavior in UIKit.

## Actual Behavior
1. When gradient has fewer colors than the original animation (e.g., 2 colors when animation expects 3), the gradient renders as **solid black**
2. When gradient color count matches the animation (e.g., 3 colors), only the **first color renders as a solid color** across the entire gradient
3. Both issues only occur in SwiftUI, not in UIKit

### SwiftUI (NOT working):
```swift
import SwiftUI
import Lottie

struct SecondView: View {
    var body: some View {
        VStack(spacing: 20) {
            LottieView(animation: .named("19633-voice-video-switch"))
                .configure { view in
                    applyAnimationChanges(to: view)
                    view.logHierarchyKeypaths()
                }
                .playing(loopMode: .loop)
                .frame(maxWidth: .infinity, maxHeight: 400)
        }
    }
    
    private func applyAnimationChanges(to animationView: LottieAnimationView) {
        let color1 = LottieColor(r: 0/255, g: 80/255, b: 115/255, a: 1)
        let color2 = LottieColor(r: 113/255, g: 199/255, b: 236/255, a: 1)
        
        // This doesn't work in SwiftUI but works in UIKit
        let gradientValueProvider = GradientValueProvider([color1, color2])
        let gradientKeyPath = AnimationKeypath(keypath: "**.**.**.Colors")
        animationView.setValueProvider(gradientValueProvider, keypath: gradientKeyPath)
        
        let colorValueProvider = ColorValueProvider(color1)
        let footKeyPath = AnimationKeypath(keypath: "Foot.Rectangle 1.Stroke 1.Color")
        animationView.setValueProvider(colorValueProvider, keypath: footKeyPath)
    }
}
```

### UIKit (Working correctly):
```swift
import UIKit
import Lottie

class SecondViewController: UIViewController {
    @IBOutlet weak var animationView: LottieAnimationView!
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        let animation = LottieAnimation.named("19633-voice-video-switch")
        animationView.animation = animation
        animationView.contentMode = .scaleAspectFit
        animationView.loopMode = .loop
        
        applyAnimationChange()
        animationView.logHierarchyKeypaths()
        animationView.play()
    }
    
    private func applyAnimationChange() {
        let color1 = LottieColor(r: 0/255, g: 80/255, b: 115/255, a: 1)
        let color2 = LottieColor(r: 113/255, g: 199/255, b: 236/255, a: 1)
        
        // This works perfectly in UIKit
        let gradientValueProvider = GradientValueProvider([color1, color2])
        let gradientKeyPath = AnimationKeypath(keypath: "**.**.**.Colors")
        animationView.setValueProvider(gradientValueProvider, keypath: gradientKeyPath)
        
        let colorValueProvider = ColorValueProvider(color1)
        let footKeyPath = AnimationKeypath(keypath: "Foot.Rectangle 1.Stroke 1.Color")
        animationView.setValueProvider(colorValueProvider, keypath: footKeyPath)
    }
}
```

## Additional Notes
The `.configure` modifier seems to execute at the right time (the closure is called), but value providers don't apply correctly. This appears to be a timing or initialization issue specific to the SwiftUI implementation.

Is there a recommended way to use value providers with the native SwiftUI `LottieView`?

dinozaurus0:
This PR fixes an issue where the gradient color value provider does not work correctly when it supplies fewer colors than those defined in the original `Lottie` file. Fixes [issue](https://github.com/airbnb/lottie-ios/issues/2616).

To render gradients, `CARenderer` uses a `CAGradientLayer` under the hood. This layer requires its `colors` property to be set, prior to adding any colors related animation, it order for the animation to work.

The issue arises when the number of colors provided by the value provider does not match the number defined in the original `Lottie` file. If the value provider supplies fewer colors than those initially set on the colors property, `CoreAnimation` ignores the gradient animation altogether.

To address this, we should derive the color count from the value provider (when one is set), and fall back to the `Lottie` file only if no value provider is provided.

| Before | After |
| ---- | ---- |
| before |  After |


Tests/__Snapshots__/AnimationKeypathTests/testCoreAnimationEngineKeypathLogging.Issues-issue_1854.txt
```diff
@@ -13,4 +13,5 @@
 .accent.Transform.Position
```
dinozaurus0, line 14: Hey @calda  👋 

The MR isn’t fully ready yet, so I’ve marked it as draft until I fix the failing tests. I’m opening it now to clarify a couple of things related to testing.

After my change, some tests related to value providers and colors started failing. The root cause seems to be that an additional keypath is now used to fetch the number of colors passed through the value provider. Because of this, issue_1854 fails.

From what I understand, the immediate fix is to update the corresponding .txt file by adding the missing color keypath. I may need to dig a bit deeper into ValueProviderStore, but regardless, this update seems necessary.

While looking into this, I noticed there are many “issue tests” in the app that rely on .txt files matching keypaths generated from Lottie files. This leads to my first question:
- Do I need to add/update a .txt file for this issue as well, even if the behavior is covered by an existing test?

My second question is about snapshot tests. The repo mentions that changes should be accompanied by snapshot tests:
- Should I add a new snapshot test for this change? 

Thank you 🙏

calda (commented):
Nice work! Just need to get the tests passing

Tests/__Snapshots__/AnimationKeypathTests/testCoreAnimationEngineKeypathLogging.Issues-issue_1854.txt
calda: Yes! Please add a new snapshot test for the animation you were using to test this, and configure the gradient provider in `SnapshotConfiguration.swift`.

New snapshot files are automatically created as things are added. You'll need to re-record any failing snapshots, either by setting `isRecording = true` or deleting the failing ones. Just made sure you don't accidentally batch re-generate a large number of unrelated image snapshots.

Tests/__Snapshots__/AnimationKeypathTests/testCoreAnimationEngineKeypathLogging.Issues-issue_1854.txt
```diff
@@ -13,4 +13,5 @@
...
 .accent.Transform.Rotation
 .accent.Transform.Scale
-.accent.Transform.Scale
\ No newline at end of file
+.accent.Transform.Scale
+Colors
```
calda: This key path doesn't seem right, seems like it's missing the layer name, should be something like `.accent.Gradient Fill 1.Colors` instead.

dinozaurus0: Update this to include the full path 🙏

Tests/__Snapshots__/AnimationKeypathTests/testCoreAnimationEngineKeypathLogging.Issues-issue_1854.txt
dinozaurus0: I made all test works so now it should be fine. During the process I had some issues with the following test:

`testCoreAnimationEngineKeypathCompatibility`. 

What happened is that for a certain Lottie file, the main thread did not include the entire Core Animation key paths. I am not sure why this happened. The file was a file I modified by hand and I might have missed some things. My question here is why is this a requirements for the Core Animation keys paths to be a subset of the main thread keys?

ps: As a side note, pretty neat way of generating tests ^^

calda (approved):
Thank you!

Issue #2624, opened 2025-12-02
CALayerInvalidGeometry with iOS 26 and Pro Max Devices


kiancheong:
Reopening this issue: https://github.com/airbnb/lottie-ios/issues/2617
We managed to reproduce the issue with **iPhone 16 Pro Max** and **iPhone 17 Pro Max** on both **device** and **simulator**. 
However, the issue **did not occur** on iPhone 15 Pro Max. 

Crash message

> Fatal Exception: CALayerInvalidGeometry
> CALayer position contains NaN: [nan 89.8333]. Layer: <CALayer:0x1290123a0; position = CGPoint (inf 89.8333); bounds = CGRect (0 0; inf 179.667); delegate = <TtGC6Lottie27SwiftUIMeasurementContainerCS_19LottieAnimationView: 0x1293a8a00; frame = (nan 0; inf 179.667); autoresize = W+H; layer = <CALayer: 0x1290123a0>>; sublayers = (<CALayer: 0x129010780>); opaque = YES; allowsGroupOpacity = YES; >

We are using `Lottie 4.5.2`

Here's the code we used for the LottieView

```
VStack(spacing: .zero) {
      LottieView(animation: .named(lottieFileName, bundle: .module))
        .resizable()
        .scaledToFill()
        .frame(
          width: Constants.lottieAnimationWidth,
          height: Constants.lottieAnimationHeight
        )
        .padding(.horizontal, 16)
        .padding(.bottom, 40)

      Text(someText)
        .foregroundStyle(Color.blue))
        .multilineTextAlignment(.center)
    }
    .frame(maxWidth: .infinity, maxHeight: .infinity)
    .padding(.horizontal, 16)
    .padding(.vertical, 16)
}
```

kiancheong:
Have provided sample code that reproduces this issue above.

calda:
Thanks, I can reproduce it now.

sutheesh:
Hi @calda  are you looking on this? do you need an extra hand?

calda:
PRs welcome!

sutheesh:
@calda  created a pr to to consume latest epoxy 
https://github.com/airbnb/lottie-ios/pull/2681

PR #2678, opened 2026-04-26
#2624: Fix CALayerInvalidGeometry crash and layout hang 

sutheesh:
## Problem: #2624

On iOS 26 with `fullScreenCover`, using `.resizable().scaledToFill()` on a `LottieView` inside a `VStack` with wrapping text caused two distinct failures.

### 1. CALayerInvalidGeometry crash

`SwiftUIMeasurementContainer` initialised with `bounds = (0, 0)` on iOS 16+. When SwiftUI's `.scaledToFill()` queried the view's ideal size and received `(0, 0)`, it computed a scale factor of `proposedWidth / 0 = ∞` and assigned an infinite frame to the `UIViewRepresentable` host.

On iOS 26, UIKit applies **proportional sublayer repositioning** inside UIView animation blocks. When the container's bounds transitioned from `(0, 0)` to the final size during the `fullScreenCover` presentation animation, UIKit computed:

```
newPosition = oldPosition × (newBoundsWidth / oldBoundsWidth)
            = 70 × (432 / 0) = 70 × ∞ = NaN
```

This NaN value was passed directly to `-[CALayer setPosition:]`, throwing `CALayerInvalidGeometry`.

### 2. 3–4 minute layout hang

iOS 26's `fullScreenCover` uses a significantly longer presentation animation than previous iOS versions. During this animation, UIKit drives the container's bounds through many intermediate values. The existing `layoutSubviews()` check used exact float equality (`bounds.size != latestMeasurementBoundsSize`), which fired `invalidateIntrinsicContentSize()` on every intermediate frame — potentially thousands of times. Each call forced SwiftUI to re-layout the entire view hierarchy, including the expensive multi-pass text measurement needed for wrapping labels, causing the modal to appear after 3–4 minutes.

### Why only with wrapping text?

Short single-line text completes its layout in a single pass, which finishes *before* the `fullScreenCover` animation block starts. Long wrapping text requires a multi-pass VStack layout that runs *inside* the animation block, triggering both the proportional repositioning NaN and the repeated invalidation.

---

## Fix

Five coordinated changes across three files.

### `SwiftUIMeasurementContainer.swift`

- **Non-zero initial bounds** — on iOS 16+, the container now uses the content's `intrinsicContentSize` as its initial frame instead of `(0, 0)`. For a 140×140 animation the transition becomes `70 × (432 / 140) = 216` — always finite, no NaN. Falls back to `(0, 0)` for async-loaded animations where intrinsic size is unavailable at init time.
- **`NonFiniteClampingLayer` + `_SwiftUIMeasurementContainerBase`** — a `CALayer` subclass that clamps any `∞`/`NaN` `position` or `bounds` before forwarding to `super`. Placed in a non-generic concrete base class to ensure reliable ObjC runtime dispatch for `+layerClass` and `-setFrame:`.
- **`measureView()` fallback** — when the strategy is `.proposed` and the fitting size is zero (async animations, or any remaining zero-bounds edge case), the measurement falls back to `content.intrinsicContentSize`, ensuring `.scaledToFill()` always receives a real aspect ratio and never produces an infinite scale factor.
- **1pt threshold in `layoutSubviews()`** — `invalidateIntrinsicContentSize()` is now only called when bounds change by ≥ 1pt in any dimension, skipping the sub-point intermediate values driven by iOS 26's long presentation animation.

### `MeasuringViewRepresentable.swift`

- **`finiteOrNil` on proposals** — `CGFloat.infinity` proposals from SwiftUI modifiers like `.scaledToFill()` are normalised to `noIntrinsicMetric` (unconstrained) instead of being treated as a concrete size that propagates through to the frame.

### `LottieAnimationView.swift`

- **`layoutAnimation()` guard** — skips animation layout when `bounds` are non-finite or zero, preventing `viewAspect = 0/0 = NaN` in the scale calculation.

---

### DIFF
--- a/Sources/Private/EmbeddedLibraries/EpoxyCore/SwiftUI/LayoutUtilities/MeasuringViewRepresentable.swift
+++ b/Sources/Private/EmbeddedLibraries/EpoxyCore/SwiftUI/LayoutUtilities/MeasuringViewRepresentable.swift
@@ -55,10 +55,14 @@ extension MeasuringViewRepresentable {
     // extra work here.
     let children = Mirror(reflecting: proposedSize).children
 
-    // Creates a `CGSize` by replacing `nil`s with `UIView.noIntrinsicMetric`
+    // Creates a `CGSize` by replacing `nil`s and non-finite values (e.g. `CGFloat.infinity`
+    // proposed by `.scaledToFill()`) with `UIView.noIntrinsicMetric`. Non-finite values must be
+    // treated as unconstrained — passing them through causes `SwiftUIMeasurementContainer` to
+    // return an infinite fitting size, which SwiftUI propagates to the host view frame and triggers
+    // a `CALayerInvalidGeometry` crash before layout even reaches the animation layer.
     uiView.proposedSize = .init(
-      width: children.first { $0.label == "width" }?.value as? CGFloat ?? ViewType.noIntrinsicMetric,
-      height: children.first { $0.label == "height" }?.value as? CGFloat ?? ViewType.noIntrinsicMetric
+      width: (children.first { $0.label == "width" }?.value as? CGFloat).finiteOrNil ?? ViewType.noIntrinsicMetric,
+      height: (children.first { $0.label == "height" }?.value as? CGFloat).finiteOrNil ?? ViewType.noIntrinsicMetric
     )
 
     size = uiView.measuredFittingSize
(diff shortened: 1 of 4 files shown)


sutheesh:
https://github.com/user-attachments/assets/2f132ae1-8ef1-44dc-9b42-3a2732340fb7

@calda Please check this and let me know your thoughts

calda (commented):
Libraries under `/EmbeddedLibraries` are not meant to be edited directly, they should be kept in sync with versions of the upstream library.

calda:
Does updating to the latest version of epoxy-ios fix the issue? If not we would need to fix the crash upstream in Epoxy and then update the Epoxy dependency here

sutheesh:
Huhh My bad, I see the readme now, do you think I can open pr to https://github.com/airbnb/epoxy-ios ?

calda:
Yeah, although last time I checked there were also compatibility issues updating the epoxy-ios version used in this repo breaking macOS support: https://github.com/airbnb/epoxy-ios/pull/163

So we would also need to check / resolve that issue before we can update the dependency version

Issue #2679, opened 2026-04-28
Xcode 26 and later versions crash


Enjoyworld-AI:
<img width="1692" height="1160" alt="Image" src="https://github.com/user-attachments/assets/fe3b282e-9d7c-4b5a-b966-84882a31d1fd" />

efstathiosntonas:
same here

guptamohit1907:
Added the fix in this PR 
https://github.com/airbnb/lottie-ios/pull/2682

Enjoyworld-AI:
When will the next tag version be released? @calda

Huaian666:
the docs are actually good?? rare W

jjquesada03:
￼

PR #2680, opened 2026-05-02
Fix re-entrant make​Animation​Layer crash on iOS 26 / Xcode 26 (#2679)

pgovindaraj1:
## Root Cause

The crash is caused by re-entrant calls to make​Animation​Layer(using​Engine:):

1. Lottie​Animation​Layer​.init calls make​Animation​Layer → creates a Core​Animation​Layer (automatic engine)
2. add​Sublayer(animation​Layer) is called inside make​Animation​Layer
3. On iOS 26, Core Animation calls display() on the new sublayer synchronously before add​Sublayer returns
4. display() → setup​Animation throws a compatibility error → did​Set​Up​Animation fires → automatic​Engine​Layer​Did​Set​Up​Animation → make​Animation​Layer(using​Engine: .main​Thread) (re-entrant call)
5. The re-entrant call tears down root​Animation​Layer and replaces it
6. Control returns to the original make​Animation​Layer call, which now accesses the deallocated layer → EXC​_​BAD​_​ACCESS

## Fix

Added a reentrancy guard (is​Making​Animation​Layer) to make​Animation​Layer(using​Engine:). If a re-entrant call arrives while the method is executing, it is stored in pending​Animation​Layer​Engine and dispatched via defer after the original call fully completes. This preserves the automatic→main-thread fallback behavior while ensuring layer state is consistent throughout.

### DIFF
--- a/Sources/Public/Animation/LottieAnimationLayer.swift
+++ b/Sources/Public/Animation/LottieAnimationLayer.swift
@@ -1154,6 +1154,23 @@ public class LottieAnimationLayer: CALayer {
   }
 
   fileprivate func makeAnimationLayer(usingEngine renderingEngine: RenderingEngineOption) {
+    // Guard against re-entrant calls. On iOS 26+, Core Animation can call `display()` on a
+    // newly-created sublayer synchronously during `addSublayer`, which may trigger the automatic
+    // engine fallback path and attempt to call `makeAnimationLayer` again while we're still inside
+    // this method. Defer the re-entrant request and execute it after the current call completes.
+    guard !isMakingAnimationLayer else {
+      pendingAnimationLayerEngine = renderingEngine
+      return
+    }
+    isMakingAnimationLayer = true
+    defer {
+      isMakingAnimationLayer = false
+      if let pendingEngine = pendingAnimationLayerEngine {
+        pendingAnimationLayerEngine = nil
+        makeAnimationLayer(usingEngine: pendingEngine)
+      }
+    }
+
     /// Disable the default implicit crossfade animation Core Animation creates
     /// when adding or removing sublayers.
     actions = ["sublayers": NSNull()]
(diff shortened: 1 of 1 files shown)


calda:
Do you have repro steps for this crash? A sample project demonstrating the issue would be helpful.

pgovindaraj1:
@calda Ive created a sample app and push up the repo. https://github.com/pgovindaraj1/Lottie---Demo
I able to make it crash when I ran the app on my iPhone running iOS 26.3. Here is the screenshot.
Screenshot 2026-05-03 at 10 17 12 AM

calda:
What are the repro steps? When I build and run the app, I don't hit a crash.

This PR adds a fair bit of extra complexity to the animation setup flow, and I'm also somewhat doubtful that Apple changed the behavior of Core Animation in a way that would introduce this sort of reentrancy.

pgovindaraj1:
@calda Not sure if you're running my sample app on a simulator or a real device. I wasn't able to make it crash on a simulator, but an actual device.

Repro steps: 
1. Run my test app, https://github.com/pgovindaraj1/Lottie---Demo on an actual device(iPhone).
2. App will crash when launched.

calda:
When I run your sample project on my device, I get a "cannot find Lottie dylib" crash.

When I change it from lottie-spm to lottie-ios, that crash is resolved, and then the app does not crash on launch. I'm testing on a device with iOS 26.4.2. Are you able to repro your crash on that OS version? Can you share the exact stack trace?

calda:
This content from your screenshot matches the issue I see when using lottie-spm as opposed to lottie-ios.

image

Given this matches your screenshot and is not related to the `makeAnimationLayer` change, I am going to close this PR for now. I will check if I can fix the issue with lottie-spm.

calda:
If you can create a repro case that demonstrates the `makeAnimationLayer` crash, feel free to re-open this PR 👍🏻

calda:
The lottie-spm problem is that the project is not embedding the Lottie framework:

image


Normally there is an "Always Embed" option in this UI that you must ensure is selected, but I can't figure out why it isn't present. Generally I think this is an Apple / Xcode bug. Once the library is configured to be embedded, it will work correctly. I confirmed that the Example project in the lottie-spm repo still embeds the library correcly.

Issue #2537, opened 2025-02-19
Add support for IANA approved mime-type "video/json+lottie" and extension ".lot"


kudanai:
As part of efforts by LAC, IANA has approved the `video/json+lottie` type and `.lot` extension for use with Lottie animations. In light of this we should consider adding the relevant recognition in the library.

---

### Is your feature request related to a problem? Please describe.
Enhancement. lottie-android should recognize `.lot` and the `video/json+lottie` mime type on the network loader

### Describe the solution you'd like

* pick up `.lot` from storage, and inside `.zip|.lottie` files (if applicable)
* recognize `video/json+lottie` in the network pipeline

In both these instances, they should be treated the same as .json would be.
for clarity, this in now way disqualifies the usage of `.json` to transport animations. The `.lot` extension is in addition to the existing mechanism

### Additional context

[official Assignment](https://www.iana.org/assignments/media-types/video/lottie+json)
[standards listings](https://www.iana.org/assignments/media-types/media-types.xhtml)
[organizational listing with LAC recognized](https://www.iana.org/assignments/iesg-recognized-organizations/iesg-recognized-organizations.xhtml)

kudanai:
minor correction, the correct type is: `video/lottie+json`. Edited the original post

PR #2688, opened 2026-06-09
feat: recognize the IANA `.lot` extension as a Lottie JSON alias

tsushanth:
## Why

Implements the file-extension half of #2537.

IANA registered the [\`video/lottie+json\`](https://www.iana.org/assignments/media-types/video/lottie+json) media type and the \`.lot\` short extension as a canonical identifier for Lottie animations. The Android counterpart issue tracks the same change.

## What

Treats \`.lot\` as a synonym for \`.json\` in three small places:

- New \`DotLottieUtils.lotExtension\` constant.
- \`URL.isJsonFile\` returns \`true\` for either \`.json\` or \`.lot\`.
- \`Bundle.getAnimationData\` falls back to a \`.lot\` lookup when the \`.json\` lookup misses; the bundled \`removingJSONSuffix\` helper strips a trailing \`.lot\` too so callers can pass either filename form.

\`.lottie\` (DotLottie archive) means a different file format and is left untouched.

## What's intentionally out of scope

Network-side recognition of the \`video/lottie+json\` content-type header. The SDK doesn't inspect response MIME types on \`LottieAnimation\` network loads today, so adding that would be a separate, larger change (URLSession plumbing + \`AnimationCacheProvider\` keys + tests). Happy to do that as a follow-up if you want this PR limited to file-extension support first.

## Build

\`swift build\` → clean.


calda (commented):
Thanks! Could you please also add a `.lot` file snapshot test case under `Tests/Snapshots` and record a snapshot for it?

tsushanth:
Thanks @calda — pushed [`12495084`](https://github.com/airbnb/lottie-ios/pull/2688/commits/12495084) with `Tests/Samples/DotLotExtension/Switch.lot` (a copy of `Switch.json` — kept it simple since the goal here is to exercise the extension lookup path, not introduce new animation content).

Wired into the existing Samples enumeration:
- `Samples.sampleAnimationURLs` now also globs `.lot`
- `Samples.sampleAnimationNames` strips the `.lot` suffix alongside `.json` / `.lottie`
- `testSnapshotsHaveCorrespondingSampleFile` accepts `.lot` as a valid backreference suffix

I'll record the snapshot PNG and push a follow-up commit shortly. Let me know if you'd prefer the `.lot` sample to live somewhere flatter (e.g. directly under `Tests/Samples/Switch.lot` instead of the subdir) — happy to move it.

tsushanth:
Snapshots recorded in [`e5f5fb26`](https://github.com/airbnb/lottie-ios/pull/2688/commits/e5f5fb26) — 10 PNGs (2 rendering engines × 5 default progress percentages) under `Tests/__Snapshots__/SnapshotTests/`, captured on iPhone SE (3rd gen) simulator, iOS 18.2, 2x scale to match the existing snapshot device class.

Since `Switch.lot` is a byte-for-byte copy of `Switch.json`, the rendered output is identical to the existing `Switch` snapshots — what's being exercised is the `.lot` extension resolution through `URL.isJsonFile` and `Bundle.getAnimationData`, not new animation content. Happy to swap the file for a smaller/different animation if you'd prefer.

CI should be green now. PTAL whenever you get a chance.

calda (approved):

tsushanth:
One more update in [`f8a8f0ca`](https://github.com/airbnb/lottie-ios/pull/2688/commits/f8a8f0ca) — noticed the new snapshots were rendering in the animation's native teal because `DotLotExtension/Switch` wasn't picking up the same `ColorValueProvider` overrides that `Switch` has. Mirrored the `customMapping` entry so they render identically (black/white toggle). That makes the test's story cleaner: same file content + same value providers = byte-equivalent rendering, proving the `.lot` resolution path is a true alias.

Final state of the branch is now ready for review.

calda:
There are a few other tests failing from missing snapshots, could you ensure the full test suite passes?

anandfresh (approved):

PR #2689, opened 2026-06-10
Update embedded ZIPFoundation to 0.9.20 to prevent symlink path traversal during ZIP extraction

rm335:
## Summary
- Fix a symlink path traversal vulnerability in the embedded ZipFoundation library used for
  `.lottie` (ZIP) extraction: extracted file paths were validated via `isContained(in:)`, but
  symlink **targets** were not — a malicious `.lottie` file could create symlinks pointing to
  arbitrary system files (e.g., `/etc/passwd`)
- Fix by updating the embedded ZIPFoundation from 0.9.16 to 0.9.20, which validates symlink
  targets by default (`allowUncontainedSymlinks: false`): both absolute and relative targets are
  resolved and must remain within the symlink's parent directory, otherwise extraction throws
  `ArchiveError.uncontainedSymlink`
  - Note: the first commit on this branch added a custom `allowedDestination` validation; it was
    superseded by upstream's built-in containment check when the library was updated to 0.9.20
- Supporting changes to keep the embedded copy buildable:
  - Added the three files new in 0.9.20 (`Archive+Deprecated.swift`, `Date+ZIP.swift`,
    `FileManager+ZIPDeprecated.swift`) to `Lottie.xcodeproj` for all four framework targets —
    without these, every xcodeproj-based build fails (SPM globs sources; the Xcode project doesn't)
  - Reformatted the imported ZipFoundation sources with the repo's SwiftFormat configuration
  - Changed upstream's `var typeInt` to `let` in `FileManager+ZIP.swift` (never mutated; CI builds
    with `-warnings-as-errors`)

## Test plan
- [x] `swift package format --lint` passes
- [x] `xcodebuild build` for the Lottie iOS/macOS/tvOS schemes with `-warnings-as-errors` succeeds locally
      (visionOS not verified locally — SDK not installed — but it compiles identical sources)
- [ ] Create a test `.lottie` ZIP containing a symlink with an absolute target outside the
      extraction directory (e.g., `/etc/passwd`) and confirm extraction throws
      `ArchiveError.uncontainedSymlink`
- [ ] Create a test `.lottie` ZIP containing a symlink with a relative target escaping via `../../`
      and confirm extraction throws
- [ ] Create a test `.lottie` ZIP containing a symlink with a valid relative target within the
      extraction directory and confirm extraction succeeds
- [ ] Run existing unit/integration tests to verify no regressions (CI `Test Package` job)

### DIFF
--- a/Lottie.xcodeproj/project.pbxproj
+++ b/Lottie.xcodeproj/project.pbxproj
@@ -110,6 +110,9 @@
 		080DEFD32A95711400BE2D96 /* Entry+Serialization.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08C001E92A46150D00AB54BA /* Entry+Serialization.swift */; };
 		080DEFD42A95711400BE2D96 /* Archive+Helpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08C001E02A46150D00AB54BA /* Archive+Helpers.swift */; };
 		080DEFD52A95711800BE2D96 /* Archive+ReadingDeprecated.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08C001EF2A46150D00AB54BA /* Archive+ReadingDeprecated.swift */; };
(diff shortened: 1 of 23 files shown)


calda (approved):
Thanks!

PR #1421, opened 2021-12-01
Resolved keyNotFound error for iOS

aphex3k:
This change more closely aligns the lottie-ios library with Web and Android as this behavior is akin to the other platforms instead of throwing an error. Resolves #1403

PR #2690, opened 2026-06-13
  Don't fail decoding when text font 'ascent' or line height ('lh') is missing

fanmingyi:
## Background

  While exporting an animation as JSON from a web-based Lottie editor
  (LottieFiles Creator), I hit a confusing inconsistency: the exact same
  file rendered correctly in **lottie-web** and **lottie-android**, but
  failed to load **entirely on lottie-ios** — not just the text, the whole
  animation was blank.

  It took a lot of digging to track down, because the file is a perfectly
  valid Lottie and works everywhere else. The root cause turned out to be a
  difference in how lottie-ios parses text layers compared to the other
  official players: lottie-ios treats two text fields as **required** and
  throws when they're missing, while lottie-web and lottie-android treat
  them as optional and fall back to sane defaults. The web editor simply
  doesn't emit those two fields, so only iOS rejects the file.

  ## Root cause
  
  The `ascent` field on a font and the `lh` (line height) field on a text
  document are **optional** in the Lottie format (neither is listed in the
  `required` array of the [lottie-docs](https://github.com/LottieFiles/lottie-docs)
  schema for `font` / `text-document`), but lottie-ios decodes both as
  required `Double`s. When either key is absent, decoding throws and fails
  the **entire** animation rather than just the text layer.

  This diverges from the other official players:
  
  - **lottie-web** (`FontManager.js` / `TextProperty.js`): never requires
    `ascent` (glyph metrics are measured at runtime) and falls back to
    `fontSize * 1.2` when `lh` is missing.
  - **lottie-android** (`FontParser.java` / `DocumentDataParser.java`):
    initializes both `ascent` and `lineHeight` to `0` and only overwrites
    them when the key is present, so a missing field never fails decoding.

  ## Fix
  
  Align lottie-ios with lottie-android by defaulting both fields to `0`
  when absent instead of throwing. The fix is applied to **both** decoding
  paths so they stay consistent (`ParsingTests` cross-checks them):

  - `.dictionaryBased` (default): use the existing
    `(try? dictionary.value(for:)) ?? default` idiom.
  - `.legacyCodable`: add an explicit `init(from:)` that mirrors the
    compiler-synthesized behavior but uses `decodeIfPresent(...) ?? 0` for
    `ascent` and `lineHeight` (same pattern already used by `Mask`,
    `ShapeItem`, etc.).
  
  This follows the same cross-platform-alignment approach previously
  accepted in #1421 ("Resolved keyNotFound error for iOS").

  ## Testing

  Added a `Text_NoAscentOrLineHeight` sample (a text animation with both
  fields stripped) under `Tests/Samples/`, so the existing data-driven
  `ParsingTests` and `SnapshotTests` automatically cover it:

  - `ParsingTests` verifies the sample now decodes identically under both
    `.dictionaryBased` and `.legacyCodable` (previously `.legacyCodable`
    threw `keyNotFound`).
  - `SnapshotTests` confirms it renders correctly; the recorded snapshots
    are checked in.

  Animations exported without these fields (e.g. from LottieFiles Creator)
  now decode and play, matching lottie-web and lottie-android.
  
  ## Example

  Here is a minimal animation exported from the web editor that reproduces
  the issue. The font has no `ascent` and the text document has no `lh`.
  It renders fine on lottie-web and lottie-android, but on lottie-ios
  (before this PR) the whole animation fails to decode. With this PR it
  decodes and renders correctly.
image

[JaName (1).json](https://github.com/user-attachments/files/28914466/JaName.1.json)


calda (approved):
Thank you for the high quality investigation and pull request!

fanmingyi:
@calda Thank you so much for the kind words and the quick review — really appreciate it! 🙏

  Would you have a rough idea of when this could be merged and included in a
  release? We're hitting this exact issue in production (text animations
  exported from the web editor fail to load on iOS only), so a released
  version with the fix would really help unblock our team. No rush on the
  exact timing — just trying to plan on our side. Thanks again!

calda:
Its been a while since the last release so I will make one now

PR #2694, opened 2026-07-07
Fix crash on Xcode 26 by setting `actions` before `anchorPoint` in CompositionLayer

### LINKED ISSUE #2679: Xcode 26 and later versions crash

leticiamazzoportela:
## Problem

Apps using Lottie crash on launch when built with Xcode 26 / iOS 26 SDK:

```
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000039
0 QuartzCore -[CALayer setAnchorPoint:] + 116
1 Lottie CompositionLayer.init(layer:size:) + 32
2 Lottie PreCompositionLayer.init(...) + 34
```

Fixes #2679

## Root cause

In `CompositionLayer.init(layer:size:)`, `anchorPoint = .zero` was assigned **before** the `actions` dictionary was configured, for both `self` and `contentsLayer`. When `setAnchorPoint:` runs, Core Animation checks the `"anchorPoint"` key in `actions` to decide whether to create an implicit animation. Since `actions` was still `nil` at that point, it took the animation-creation code path — which, in the QuartzCore shipped with Xcode 26 / iOS 26, dereferences a pointer that's null because the layer hasn't yet been attached to a layer tree.

## Fix

Reorder the assignments so `actions` is set before `anchorPoint`, for both `self` and `contentsLayer`. No behavior change otherwise — this only closes the window where `actions` was nil during the anchorPoint mutation.

## Testing

Added `Tests/Samples/Issues/issue_2679.json`, a minimal precomp-based sample that exercises the exact `PreCompositionLayer → CompositionLayer.init` path from the crash. Verified it renders correctly (no crash) across both the Main Thread and Core Animation rendering engines; new reference snapshots included.

## Checklist

- [x] Scoped to a single issue
- [x] Includes a snapshot test case for the fix
- [x] No new SwiftLint violations (`bundle exec rake lint:swift` passes)

## Notes

The root cause here was originally diagnosed in #2682 by @guptamohit1907, which bundled this fix together with an unrelated change for #2205. Per @calda's review feedback there, this PR isolates just the #2679 fix and includes the requested snapshot test.


calda (approved):
Thanks for the fix! The change itself looks harmless otherwise and I see all the snapshot tests are passing.

PR #2697, opened 2026-08-28
Add watchOS support

Clancey:
Lottie has never built for watchOS. This adds it.

## Why it was blocked, and why that turns out to be surmountable

The usual explanation is that watchOS has no CoreAnimation. That is not so — I checked:

- `CALayer` answers **469 selectors** on watchOS.
- The watchOS QuartzCore binary exports `CALayer`, `CAShapeLayer`, `CAGradientLayer`, `CAAnimation` and 92 other CA classes.
- Reached through the ObjC runtime, a `CAShapeLayer` on watchOS renders a filled disc into a bitmap context — measured centre pixel `255,38,0,255`, corner transparent.

What is withheld is only the *declarations*: every CA type is marked `API_UNAVAILABLE(watchos)` in the SDK headers.

That distinction is load-bearing, because Lottie's renderer is built from `CALayer` subclasses. No compiler flag gets past it — `-disable-availability-checking` permits *calling* an unavailable class but not *inheriting* from one.

## Approach

A **`CAShim` target**: the QuartzCore headers with the availability macros stripped. Lottie's `QuartzCore` imports route through it on watchOS, and the symbols resolve against the real framework, linked explicitly since watchOS does not link it implicitly.

The rest follows from watchOS having no `UIView`:

- The view layer (`LottieAnimationView`, `AnimatedControl` and friends), the ObjC compatibility types and EpoxyCore's SwiftUI plumbing are fenced off. **`LottieAnimationLayer` remains and is the entry point on watchOS.**
- **UIKit is excluded from the watchOS build entirely.** Importing it *anywhere* re-exports the gated CoreAnimation, and because Swift compiles a module as a whole, one such file beats the shim in every other. So image decoding goes through ImageIO, text attributes and paragraph styles through CoreText, and reduced motion through a declared WatchKit symbol.
- `BundleImageProvider` gets a watchOS counterpart. `NSDataAsset` is UIKit's, so asset-catalogue lookup is unavailable there; file and data loading — what exported animations use — is unaffected.

## A caveat consumers should know

Rendering must use `LottieConfiguration(renderingEngine: .mainThread)`. The Core Animation engine drives frames on the render server, where `render(in:)` sees the model layer and returns nothing.

Worth saying plainly: this links a class Apple does not publish on watchOS. It is not private API in the usual sense — `CALayer` is public on every other platform — but it is not in the watchOS SDK, and I would not want that buried. Happy to gate it behind a trait or an opt-in product if you would prefer.

## Verification

- Rendered an animation on an Apple Watch Ultra simulator: parsed at 375×157 over 60 frames, rasterised to a bitmap, pixels checked.
- Built **every** platform scheme with `-warnings-as-errors`: iOS, macOS, tvOS, visionOS, watchOS — all pass. `rake build:package:all` covers watchOS now.
- `swift build` passes; `swift package format --lint` reports 0 violations; the EmbeddedLibraries lint passes.
- The test suite is unchanged by this. It has 5 failures on my machine — and **the same 5 reproduce on a clean checkout of master**, so they are down to my Xcode/simulator versions, not this change.

## Left out on purpose

The CocoaPods podspec and the xcframework build still omit watchOS. Both want a bit of thought about how the shim headers should be vended, and I did not want to guess at it inside a PR that is already large. Glad to follow up.


calda:
Thank you for the PR. At this time I would not like to ship a Core Animation shim inside Lottie. This is a cool approach though, perhaps it would be best as a separate package that you publish, or a fork.

Issue #2698, opened 2026-09-01
Lottie crash while decoding Base64 image assets


Hashim1999164:
Lottie crash while decoding Base64 image assets

Crash

Fatal Exception: NSMallocException
*** -[_NSPlaceholderData _initWithBase64EncodedObject:options:]: unable to allocate memory for length (30576)

Stack trace

CoreFoundation
__exceptionPreprocess

Foundation
Data.init(base64Encoded:options:)

ImageAsset.swift:105
Data.init(dataString:options:) + 105

ImageAsset.swift:81
BundleImageProvider.imageForAsset(asset:) + 81

protocol witness for AnimationImageProvider.imageForAsset(asset:)
in conformance BundleImageProvider

CachedImageProvider.swift:26
CachedImageProvider.imageForAsset(asset:) + 26

protocol witness for AnimationImageProvider.imageForAsset(asset:)
in conformance CachedImageProvider

ImageLayer.swift:42
ImageLayer.setupImage(context:) + 42

ImageLayer.swift:19
LayerModel.makeAnimationLayer(context:) + 19

Description

Lottie crashes with "NSMallocException" while decoding a Base64 image asset.

The crash occurs inside:

Data(base64Encoded:options:)

Expected behavior

Lottie should handle invalid or unexpectedly large Base64 image assets gracefully without crashing the application.

Actual behavior

The application terminates with "NSMallocException" during image asset loading.

Thanks

PR #2700, opened 2026-09-01
Handle Base64 image decode failures gracefully

Hashim1999164:
## Summary

Fixes the crash reported in discussion #2699 (converted from issue #2698).

Embedded Base64 image assets were decoded with `Data(base64Encoded:)`, which can raise `NSMallocException` when allocation fails. This change decodes Base64 in Swift instead, returning nil for invalid or oversized payloads so Lottie skips the image instead of terminating the app.

## Test plan

- [ ] Run `DataURLTests` in Xcode
- [ ] Load an animation with a valid Base64 image asset and confirm it still renders
- [ ] Load an animation with an invalid Base64 image asset and confirm the app no longer crashes


Sources/Private/Model/Assets/ImageAsset.swift
```diff
@@ -120,6 +132,62 @@ extension Data {
 
 }
 
+// MARK: - Base64Decoder
+
+/// A small Swift base64 decoder that avoids Foundation APIs which can raise NSMallocException.
+private enum Base64Decoder {
```
calda: I think it's very likely this will be much less performant than `Data.init(base64Encoded:)`, and I also don't see how this avoids out-of-memory scenarios that trigger `NSMallocException`.

For example, your crash report included:

```
Fatal Exception: NSMallocException
*** -[_NSPlaceholderData _initWithBase64EncodedObject:options:]: unable to allocate memory for length (30576)
```

If we aren't able to allocate 30kb with of memory, I think we have bigger problems! And I don't think this Swift approach will avoid that.

Hashim1999164: Fair point on the 30kb number. This crash came from an iPhone XR, and I agree it is hard to believe that device was genuinely out of memory for a ~30kb decode. That is part of why I do not think this was a normal low memory case.

From what I saw in production, the failure happened while loading a base64 embedded image asset inside a Lottie animation, not because the phone had no RAM left. `NSMallocException` here is raised from Foundation's ObjC base64 path, and that exception is not something Swift can catch. So even a bad or unexpected payload can terminate the app instead of returning nil and skipping the asset.

That is the main reason for the change: avoid an uncatchable crash path and fail gracefully when the payload is invalid or the decode cannot complete. I am not claiming a pure Swift decoder magically fixes real device wide memory pressure.

On performance, this path only runs for embedded `data:` image assets, which is already uncommon compared to bundle or filepath images. If you would prefer a smaller change, I am happy to explore an ObjC `@try/@catch` wrapper around `Data(base64Encoded:)` instead, as long as we still avoid the hard crash.

calda: I don't think it makes sense in principle to attempt to recover from an `NSMallocException`:

> No, you cannot safely recover from an NSMallocException in an iOS or macOS application.
>
> An NSMallocException is thrown by the Foundation framework when a memory allocation request fails—meaning your application has run out of available memory to allocate for a specific object or buffer.
>
> While Objective-C allows you to technically intercept exceptions using @try/@catch blocks, doing so for a memory exception is highly dangerous. If the system lacks the memory to allocate the requested object, it likely lacks the memory to process the recovery code or exception handler safely. If your app encounters this exception, it will almost always be terminated immediately by the operating system's Out-Of-Memory (OOM) killer.

calda: How large is the image being decoded in your use case? I can imagine adding a guard like checking whether the image size is less than the amount of memory currently available to the process.

Hashim1999164: The Lottie animation we show in this flow is about 926kb overall. The crash report itself pointed at a single base64 decode step trying to allocate around 30kb, so I think that is likely one embedded image asset inside the JSON rather than the full file size.

On frequency, this is very rare for us. We are seeing about 1 crash in every 10k sessions on this screen.

Given how uncommon it is, I would also be fine with a lighter guard based approach if that fits better here, like checking estimated decode size against available memory before calling into Foundation. My main goal is just to stop the app from hard crashing on this path when something about the payload or decode fails unexpectedly.

calda: Proactively checking the available memory vs the expected image size sounds like the best approach to me

Hashim1999164: That makes sense. I pushed an update that drops the custom Swift decoder and keeps `Data(base64Encoded:)`.

Before decoding, it now estimates the decoded payload size from the base64 string and compares that against `os_proc_available_memory()`. If the estimate is larger than what the process has available, it returns nil and skips the asset instead of attempting the allocation.

Let me know if you would like any tweaks to the guard logic or logging when it bails out.

Sources/Private/Model/Assets/ImageAsset.swift
```diff
@@ -120,6 +124,38 @@ extension Data {
 
 }
 
+// MARK: - Base64DataURLDecoding
+
+/// Guards Base64 Data URL decoding when the estimated payload exceeds available process memory.
+private enum Base64DataURLDecoding {
```
calda: Thanks! Could you please generate unit tests for this? I would suggest changing it to a struct that you can then inject a custom testable `availableMemoryByteCount` closure into (which uses `os_proc_available_memory` in production)

calda: `Base64DataURLDecoder` would be a good name for the type

Hashim1999164: Done. I renamed it to `Base64DataURLDecoder` and added an injectable `availableMemoryByteCount` closure. Production uses `os_proc_available_memory()` on iOS, tvOS, and visionOS.

I also added unit tests in `DataURLTests` for the estimated size helper, skipping when available memory is too low, and decoding when enough memory is reported. When the closure returns 0, the guard is skipped so macOS can still decode without that API.

Let me know if you would like the tests split into a separate file.

Sources/Private/Model/Assets/ImageAsset.swift
```diff
@@ -120,6 +127,58 @@ extension Data {
...
+
+  // MARK: Lifecycle
+
+  init(availableMemoryByteCount: @escaping () -> Int) {
+    self.availableMemoryByteCount = availableMemoryByteCount
+  }
+
+  // MARK: Internal
+
+  static let shared = Base64DataURLDecoder(
```
calda: It is not necessary to create a singleton for this, let's just create a new struct instance every time.

Sources/Private/Model/Assets/ImageAsset.swift
```diff
@@ -120,6 +127,58 @@ extension Data {
...
 }
 
+// MARK: - Base64DataURLDecoder
+
+/// Guards Base64 Data URL decoding when the estimated payload exceeds available process memory.
+struct Base64DataURLDecoder {
+
+  // MARK: Lifecycle
+
+  init(availableMemoryByteCount: @escaping () -> Int) {
```
calda: Would suggest just doing:

```suggestion
  init(availableMemoryByteCount: @escaping () -> Int = { os_proc_available_memory() }) {
```

Sources/Private/Model/Assets/ImageAsset.swift
```diff
@@ -120,6 +127,58 @@ extension Data {
...
+    }
+  )
+
+  let availableMemoryByteCount: () -> Int
+
+  func decode(_ encodedString: String) -> Data? {
+    let estimatedDecodedByteCount = Self.estimatedDecodedByteCount(for: encodedString)
+    let availableMemory = availableMemoryByteCount()
+
+    if availableMemory > 0, estimatedDecodedByteCount > availableMemory {
```
calda: Let's also log a message to `LottieLogger.shared.warn` when we hit this condition of not having enough memory to decode an image

Hashim1999164: All three are covered in the latest push: no more singleton, default init with the memory closure (iOS/tvOS/visionOS use os_proc_available_memory), and LottieLogger.shared.warn when we skip decode for memory.

macOS still uses the 0 fallback since os_proc_available_memory is not available there, so the guard stays off on Mac like before. The shared instance is gone though, so at least it is not hogging the spotlight anymore 😅

Happy to tweak the log wording if you want it louder 🙏

Sources/Private/Model/Assets/ImageAsset.swift
```diff
@@ -134,29 +134,28 @@ struct Base64DataURLDecoder {
 
   // MARK: Lifecycle
 
-  init(availableMemoryByteCount: @escaping () -> Int) {
+  init(availableMemoryByteCount: @escaping () -> Int = {
+    #if os(iOS) || os(tvOS) || os(visionOS)
+    return Int(os_proc_available_memory())
+    #else
+    return 0
+    #endif
```
calda: What about macOS? I suppose you are special-casing zero below. It would be better to instead have `availableMemoryByteCount` return `nil` on macOS rather than `0`.

Hashim1999164: Good call. Pushed an update that switches `availableMemoryByteCount` to `() -> Int?`.

macOS now returns `nil` instead of `0`, and the guard only runs when we get a real value back from `os_proc_available_memory()` on iOS, tvOS, and visionOS. No more treating zero like a secret handshake for "skip the check" 🙂

Also aligned the default init with your earlier suggestion as closely as macOS allows. Happy to adjust further if you want it even tighter.

Sources/Private/Model/Assets/ImageAsset.swift
Hashim1999164: Updated in the latest push: the default closure now uses `os_proc_available_memory()` on iOS, tvOS, and visionOS, with `Int?` so macOS can return `nil` instead of faking it with zero 👍

Sources/Private/Model/Assets/ImageAsset.swift
```diff
@@ -120,6 +127,58 @@ extension Data {
...
+
+/// Guards Base64 Data URL decoding when the estimated payload exceeds available process memory.
+struct Base64DataURLDecoder {
+
+  // MARK: Lifecycle
+
+  init(availableMemoryByteCount: @escaping () -> Int? = {
+    #if os(iOS) || os(tvOS) || os(visionOS)
+    let availableMemory = os_proc_available_memory()
+    return availableMemory > 0 ? Int(availableMemory) : nil
```
calda: I don't think it's necessary to special case `0` on this branch. `0` from `os_proc_available_memory` means that the app has already exceeded its memory limit, not that there is no memory limit.

Hashim1999164: Fair point: 0 means we are already over the memory budget, not unlimited RAM. Removed the > 0 special case and added a test so decode skips when available memory is 0. Thanks!

calda (approved):

Sources/Private/Model/Assets/ImageAsset.swift
Hashim1999164: Fair point! I dropped the singleton in a follow up push, so we just do Base64DataURLDecoder().decode(...) each time now. Much cleaner. 👍

Hashim1999164:
Hey @calda 👋

Thanks for the approval and the review guidance throughout this!

A couple of CI checks are still red on the latest push:
- **Lint Swift**: `organizeDeclarations` on `Base64DataURLDecoder` in `ImageAsset.swift`
- **Test Package**: `testValidDataURL` is failing, and the `Nonanimating/base64Test` snapshots are off too. Looks like when `os_proc_available_memory()` returns `0` on the simulator, we skip every base64 decode (including valid ones). Happy to adjust that if you have a preferred approach.

Is there anything else you need from me before this can merge? Also, which Lottie version should this land in once it ships?

Happy to push a follow up for the CI bits as soon as you say the word 🙂

calda:
Yes, please fix the CI failures, then we can merge. For the simulator issue I would suggest adding a `#if targetEnvironment(simulator) return  nil` branch to that closure to represent that the check isn't supported in the simulator.

Hashim1999164:
Pushed a follow up for the red CI.

os_proc_available_memory() can return 0 on the simulator, which was treated as no memory and skipped every base64 decode (including valid ones). Decode now proceeds unless a positive budget is actually too small. Also reordered Base64DataURLDecoder for organizeDeclarations.

Lint Swift and Test Package should be the ones to watch on this push.

Hashim1999164:
Followed the simulator suggestion.

`availableMemoryByteCount` now returns nil under `targetEnvironment(simulator)` (and on macOS), so snapshot tests decode normally. On device, a reported 0 still means the process is already over its memory limit, so decode is skipped.

Lint Swift was still failing because the default closure sat inline on `init`; that memory helper is now a static method so `organizeDeclarations` can see a normal type layout.

Hashim1999164:
@calda Done. Added `#if targetEnvironment(simulator) return nil` so the memory check is skipped on simulator, and reordered `Base64DataURLDecoder` for the lint organizeDeclarations failure.

Latest push should clear Lint Swift and Test Package. Thanks!

Hashim1999164:
Hi @calda, I believe that all the checks have passed. Let me know if there's anything required from my side. Any CLA or anything. Thanks :)

PR #2703, opened 2026-09-15
Fix: Core Animation engine ignores Repeater's startOpacity/endOpacity

carlosefonseca:
## Summary

A shape `Repeater` (`"ty": "rp"`) with a per-copy opacity falloff (`tr.so`/`tr.eo`) renders every copy at full 100% opacity under the Core Animation rendering engine. Position and rotation transforms (`tr.p`, `tr.r`) are applied correctly per copy — only the start/end opacity interpolation is silently dropped.

## Root cause

`RepeaterTransform` (used by `RepeaterLayer` to build each repeated copy's transform) computes `position`/`rotationX/Y/Z`/`scale` per copy index, but has no concept of opacity at all. `RepeaterLayer.setupAnimations` only calls `addTransformAnimations(for:context:)`, which explicitly does *not* apply `transform.opacity` (see its doc comment) — so `Repeater.startOpacity`/`endOpacity` (which the model already decodes correctly, see `Repeater.swift`) are decoded but never consumed anywhere in the Core Animation engine.

## Fix

- `RepeaterTransform` now takes the total `copyCount` and linearly interpolates a new `opacity: KeyframeGroup<LottieVector1D>` between `startOpacity` (copy 0) and `endOpacity` (last copy), matching After Effects' repeater behavior. This reuses the existing `Keyframes.combined` helper (for correctly merging two keyframe groups, even if `startOpacity`/`endOpacity` themselves animate over time) and the existing `OpacityAnimationModel`/`addOpacityAnimation(for:context:)` infrastructure already used by `Transform`, `Fill`, `Stroke`, etc.
- `RepeaterLayer.setupAnimations` now also calls `addOpacityAnimation(for: repeaterTransform, context:)`.
- `ShapeLayer.setUpRepeater` passes the already-in-scope `copyCount` through to `RepeaterLayer.init`.

## Testing

- Added `Tests/RepeaterOpacityTests.swift`: builds a minimal 4-copy repeater animation (`startOpacity: 100`, `endOpacity: 0`), renders it via `CoreAnimationLayer`, and asserts each copy's resulting `CALayer.opacity` matches the expected linear interpolation (`1.0, 0.667, 0.333, 0.0`). Verified this test fails without the fix and passes with it.
- Manually verified against a real production animation (an 8-copy leaf-ring loading spinner with `so: 100`/`eo: 15`) rendered via `LottieView` with `.configuration(LottieConfiguration(renderingEngine: .coreAnimation))` — before the fix all 8 copies rendered as solid, identical color; after the fix the ring shows the expected fading tail.


Tests/RepeaterOpacityTests.swift
```diff
@@ -0,0 +1,98 @@
+// Created for a fix verifying that Repeater `startOpacity`/`endOpacity` are
+// applied by the Core Animation rendering engine.
+
+import QuartzCore
+import XCTest
+
+@testable import Lottie
+
+@MainActor
+final class RepeaterOpacityTests: XCTestCase {
```
calda: Please add a snapshot test case that demonstrates this behavior, rather than a unit test.

PR #2704, opened 2026-09-18
Resolve track mattes referenced by `tp`

### LINKED ISSUE #2702: Track matte referenced by `tp` is ignored: the matte source layer is drawn, and mattes are paired with the wrong layer
## What happened

A layer can point at its track matte explicitly: the masked layer carries `tt` plus `tp` (the `ind` of the matte layer), and the matte layer itself is flagged with `td: 1`. Lottie doesn't decode either key — `LayerModel` only reads `tt` — and both rendering engines assume the matte source is the immediately neighbouring layer.

When an animation uses `tp` and the matte source is not adjacent to the layer that references it, two things go wrong:

1. the matte source is drawn as a normal, fully visible layer;
2. layers carrying `tt` are matted by whatever layer happens to sit next to them.

The first one is the painful one in practice: matte sources are usually filled with a bright helper colour that was never meant to be seen.

## Expected behaviour

The matte source is resolved through `tp` and is never rendered on its own, matching lottie-web, ThorVG (the dotLottie player) and the spec.

From the [Lottie spec](https://lottie.github.io/lottie-spec/latest/specs/layers/):

> `tp` | Matte Parent | Index of the layer used as matte, if omitted assume the layer above the current one
> `td` | 0-1 integer | Matte Target | If set to 1, it means a layer is using this layer as a track matte

## Reproduction

Attached `track-matte-tp.json` is built from `Tests/Samples/setValueTest.json` and has three layers, in this order:

| layer | `ind` | keys |
| --- | --- | --- |
| masked layer (blue) | 1 | `tt: 1`, `tp: 2` |
| unrelated layer (green) | 3 | — |
| matte source (red) | 2 | `td: 1` |

```swift
let animation = try LottieAnimation(dictionary: json)
let animationView = LottieAnimationView(animation: animation)
```

| actual | expected |
| --- | --- |
| Image | Image |

The red layer should never be visible, and the blue layer should be clipped to the red silhouette. The "expected" image is the same animation after rewriting `tp` into the adjacency form Lottie understands.

## Where it goes wrong

`tp` and `td` are not part of the layer coding keys — only `tt` is decoded:

https://github.com/airbnb/lottie-ios/blob/8384dfbc429a5b3748e51d2744cbaad9ecb7a6fe/Sources/Private/Model/Layers/LayerModel.swift#L227

The Core Animation engine pairs a matted layer with the next layer in z-order:

https://github.com/airbnb/lottie-ios/blob/8384dfbc429a5b3748e51d2744cbaad9ecb7a6fe/Sources/Private/CoreAnimation/Layers/CALayer%2BsetupLayerHierarchy.swift#L146-L167

The main thread engine does the same:

https://github.com/airbnb/lottie-ios/blob/8384dfbc429a5b3748e51d2744cbaad9ecb7a6fe/Sources/Private/MainThread/LayerContainers/MainThreadAnimationLayer.swift#L61-L74
https://github.com/airbnb/lottie-ios/blob/8384dfbc429a5b3748e51d2744cbaad9ecb7a6fe/Sources/Private/MainThread/LayerContainers/CompLayers/PreCompositionLayer.swift#L63-L76

Since nothing consumes it, the `td: 1` layer falls into the "no mask" branch and is rendered like any other layer.

## Suggested fix

Decode `tp` and `td` in `LayerModel`, resolve the matte through `tp` when it is present (falling back to the neighbouring layer when it isn't), and skip layers flagged `td: 1` that no layer consumes. ThorVG's Lottie loader implements exactly this and is a compact reference:

https://github.com/thorvg/thorvg/blob/main/src/loaders/lottie/tvgLottieBuilder.cpp

```cpp
if (child->mix == -1) {          // no "tp"
    child->matteTarget = static_cast<LottieLayer*>(*(p - 1));
} else child->matteTarget = parent->layerByIdx(child->mix);
...
child->matteTarget->matteSrc = true;   // never drawn on its own
```

## Environment

- Lottie 4.6.1, also reproduces on `master` (8384dfb)
- Both `.coreAnimation` and `.mainThread` rendering engines (images above were rendered with the main thread engine)
- iOS and macOS
- Reproduces with plain `.json` animations and with `.lottie` (dotLottie) files alike

[track-matte-tp.json](https://github.com/user-attachments/files/32006926/track-matte-tp.json)

Maher-Reven:
Fixes #2702.

## Problem

A layer can point at its track matte explicitly: the masked layer carries `tt` plus `tp` (the `ind` of the matte layer), and the matte layer itself is flagged with `td: 1`. Lottie decoded neither key — `LayerModel` only read `tt` — and both rendering engines assumed the matte source was the immediately neighbouring layer.

When an animation uses `tp` and the matte source is not adjacent to the layer that references it:

1. the matte source is drawn as a normal, fully visible layer — usually a bright helper colour that was never meant to be seen;
2. layers carrying `tt` are matted by whatever layer happens to sit next to them.

From the [Lottie spec](https://lottie.github.io/lottie-spec/latest/specs/layers/):

> `tp` | Matte Parent | Index of the layer used as matte, if omitted assume the layer above the current one
> `td` | 0-1 integer | Matte Target | If set to 1, it means a layer is using this layer as a track matte

## Changes

- Decode `tp` as `LayerModel.matteParent` and `td` as `LayerModel.isMatteTarget`. `td` is specified as a 0-1 integer, but some authoring tools emit a boolean, so both are accepted.
- Add `Collection<LayerModel>.resolveTrackMattes()`, shared by both rendering engines. It resolves the matte through `tp`, and otherwise falls back to the existing adjacency behavior. A layer used as a track matte is never rendered on its own.
- The Core Animation engine's `pairedLayersAndMasks()` is replaced by the shared implementation.
- The main thread engine now applies its mattes after every layer has its `bounds`, since a matte can be provided by a layer anywhere in the layer list and `InvertedMatteLayer` reads the bounds of its input matte upfront.

An unresolvable `tp` (referencing a layer that isn't in the comp) falls back to adjacency as well, so malformed files behave exactly as they do today.

## Compatibility

No existing snapshot changes. Every sample animation that uses `tp` today points at the adjacent layer, and every `td: 1` layer is already consumed as a matte, so the new resolution is a no-op for all of them.

I verified this rather than assuming it: a temporary test compared `resolveTrackMattes()` against verbatim copies of **both** previous implementations (the Core Animation engine's `pairedLayersAndMasks()` and the main thread engine's inline pairing) across all 366 comps in the 210 sample animations. The rendered layer order, the matte source for each layer, and the matte type were identical in every case.

## Testing

Adds `Tests/Samples/Issues/issue_2702.json` — the sample attached to the issue, where the matte source is referenced by `tp` and is neither adjacent to, nor above, the layer that references it. It's picked up by the existing snapshot tests, which recorded 10 images (5 progress values x 2 rendering engines) plus the `AutomaticEngineTests` entry.

I verified the snapshots actually capture the bug rather than assuming it: with `tp`/`td` decoding disabled to restore the previous behavior, all 10 fail, and **no other sample's snapshot changes** — which is what the compatibility audit above predicts.

Ran locally: 48/48 tests pass including the full snapshot suite, `Lottie (iOS)` and `Lottie (macOS)` build with `-warnings-as-errors`, and `swift package format --lint` is clean.

I recorded the snapshots on an iPad simulator rather than the iPhone SE CI uses, because `SnapshotTests.enabled` only requires a 2x-scale device and no iPhone SE runtime is installable on macOS 26 / Xcode 26. Before recording anything I checked that every existing CI-recorded snapshot passes unmodified in that setup, so the new images should be consistent with the rest of the suite.

## Note

Removing `pairedLayersAndMasks()` left `maskLayer(for:type:context:)` as the only `fileprivate` member of `CALayer+setupLayerHierarchy.swift`, so the format lint now requires it to be `private`. That one-line change is included.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Tests/TrackMatteTests.swift
```diff
@@ -0,0 +1,298 @@
...
+import XCTest
+
+@testable import Lottie
+
+// MARK: - TrackMatteTests
+
+/// Tests for track mattes referenced explicitly through `tp` (`LayerModel.matteParent`)
+/// and flagged through `td` (`LayerModel.isMatteTarget`).
+@MainActor
+final class TrackMatteTests: XCTestCase {
```
calda: Rather than adding unit tests, please add a snapshot test file that uses this functionality.

Maher-Reven: Done — `Tests/TrackMatteTests.swift` is removed and replaced with `Tests/Samples/Issues/issue_2702.json`, the sample attached to the issue. The existing snapshot tests pick it up and recorded 10 images (5 progress values x 2 rendering engines) plus the `AutomaticEngineTests` entry.

I checked that the snapshots actually capture the regression: with `tp`/`td` decoding disabled to restore the previous behavior, all 10 fail, and no other sample's snapshot changes.

One thing worth flagging: I recorded them on an iPad simulator rather than the iPhone SE that CI uses. `SnapshotTests.enabled` only requires a 2x-scale device, and no iPhone SE runtime is installable on macOS 26 / Xcode 26. Before recording anything I confirmed every existing CI-recorded snapshot passes unmodified in that setup, so the new images should be consistent with the rest of the suite — but CI is the real check, and I'm happy to re-record if anything comes out off.

calda (approved):
Thanks!

